Issue #114, opened 2023-09-27
[`UndefinedObject`] Introduce support for variable dependency annotation


karreiro:
**Are you seeking a solution to a specific problem? Let's delve into it.**

Some Shopify Liquid assets, like snippets, depend on variables that are not defined in the global scope. Instead, these variables find their definition within the parent template.

This is about supporting a type annotation to declare those variables, so the `UndefinedObject` may rely on those annotations to understand the variables available in runtime for a given template. By doing so, we hope to create an intuitive interface definition for Liquid assets.

To illustrate a potential approach, consider the following code snippet. Please note that this is not the final syntax, but merely an example to help visualize the concept:

```liquid
{% comment %}
  This snippet is designed to render product buy-buttons.
  It accepts several parameters:
  - product: {Object} Represents the product object.
  - block: {Object} Contains information about the block.
  - product_form_id: {String} Identifies the product form.
  - section_id: {String} Specifies the section to which this snippet belongs.
  - show_pickup_availability: {Boolean} Controls whether pickup availability is rendered. If set to true, pickup availability is displayed; if false, it's not (this is optional).

  Here's an example of how you might use it:
  {% render 'buy-buttons', block: block, product: product, product_form_id: product_form_id, section_id: section.id, show_pickup_availability: true %}
{% endcomment %}
```

By exploring this approach, we hope to enhance the flexibility and usability of Liquid assets, making them even more powerful tools in your development toolkit.

charlespwd:
I think we can close this. We'll reopen separate issues that break down the tasks of [Request for suggestions: LiquidDoc Shopify/liquid#1852](https://github.com/Shopify/liquid/issues/1852)

Issue #113, opened 2023-09-27
Introduce global suggestions for supporting `{% # theme-check-disable <type> %}`


karreiro:
**Is your feature request related to a problem? Please describe.**

Currently, our language server doesn't offer suggestions to ignore theme-check offenses. I believe it would be a great to have a single implementation to ignore across all checks (something like: `{% # theme-check-disable <check_name> %}`).

I don't foresee this feature promoting excessive use of the ignore function. Given that comments in the code already create a certain level of noise, so this feature seems to be more beneficial than not providing those suggestions.

charlespwd:
https://github.com/Shopify/theme-tools/blob/main/packages/theme-check-common/src/disabled-checks/index.ts#L15

charlespwd:
Ahhhhh my bad I totally missed that. à la ESLint, I love this. Great suggestion! :D

PR #115, opened 2023-09-27
Introducing `UndefinedObject` check

karreiro:
Fixes https://github.com/Shopify/theme-check-js/issues/66

https://github.com/Shopify/theme-tools/assets/1079279/7ea13af5-11c6-44d6-a585-6ce6cb965c97

The `UndefinedObject` check is being introduced to identify references to undefined Liquid objects.

This implementation largely mirrors the Ruby version, and intentionally skips snippet assets. This is because these assets depend on variables defined in parent templates, and Theme Check is designed to focus on single asset checks.

Looking ahead, the following issue is planned to be addressed next: https://github.com/Shopify/theme-tools/issues/114. Once resolved, developers will be able to annotate available variables, thereby establishing an interface/contract between Liquid templates.

Finally, this check doesn't currently provide a fix or suggestion, as the main approaches would be about (1) removing the variable or (2) ignoring the check. To enhance this, I've opened a new issue: https://github.com/Shopify/theme-tools/issues/113. This proposes a global 'ignore check' implementation, so all checks will be able to suggest an ignore dialog. I don't foresee this leading to overuse of the ignore function, as I the ignore-code introduces some noise.

## Before you deploy

- [x] This PR fixes a bug

### DIFF
--- a/packages/theme-check-common/src/checks/undefined-object/index.ts
+++ b/packages/theme-check-common/src/checks/undefined-object/index.ts
@@ -1,0 +1,156 @@
+import {
+  LiquidHtmlNode,
+  LiquidTag,
+  LiquidTagAssign,
+  LiquidTagCapture,
+  LiquidTagFor,
+  LiquidTagTablerow,
+  LiquidVariableLookup,
+  NamedTags,
+  NodeTypes,
+  Position,
+} from '@shopify/liquid-html-parser';
+import { LiquidCheckDefinition, Severity, SourceCodeType } from '../../types';
+import { last } from '../../utils';
+
+type Scope = { start?: number; end?: number };
+
+export const UndefinedObject: LiquidCheckDefinition = {
+  meta: {
+    code: 'UndefinedObject',
+    name: 'Undefined Object',
+    docs: {
+      description: 'This check exists to identify references to undefined Liquid objects.',
+      recommended: true,
+      url: 'https://shopify.dev/docs/themes/tools/theme-check/checks/undefined-object',
+    },
+    type: SourceCodeType.LiquidHtml,
+    severity: Severity.WARNING,
+    schema: {},
+    targets: [],
+  },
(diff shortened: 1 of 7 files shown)


packages/theme-check-common/src/checks/undefined-object/index.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    const variableScopes: Map<string, Scope[]> = new Map();
+    const variables: LiquidVariableLookup[] = [];
+
+    function indexVariableScope(variableName: string | null, scope: Scope) {
+      if (!variableName) return;
+
+      const indexedScope = variableScopes.get(variableName) ?? [];
+      const currentScope = [...indexedScope, scope];
```
charlespwd, line 49: Might need a sort instead for this. Since the async nature of the check doesn't _guarantee_ that they are traversed in order.

packages/theme-check-common/src/checks/undefined-object/index.ts
```diff
@@ -0,0 +1,156 @@
...
+
+        if (isLiquidTag(parent) && isLiquidTagCapture(parent)) return;
+
+        variables.push(node);
+      },
+
+      async onCodePathEnd() {
+        if (context.themeDocset) {
+          const objects = await context.themeDocset.objects();
+          objects.forEach((obj) => variableScopes.set(obj.name, []));
```
charlespwd, line 88: You're missing a bit of `access` logic in here. See TypeSystem's in the language server. (e.g. `image` is only a type and never a "globally accessible" variable)

karreiro: Excellent call :) thank you! I've applied the `access` logic here :)

karreiro:
Thanks a lot for the review, @charlespwd! Changes applied :)

PR #116, opened 2023-09-27
Added `RemoteAsset` check.

haribshahbaz:
Resolves https://github.com/Shopify/theme-check-js/issues/62

This PR introduces the `RemoteAsset` check into our existing `AssetUrlFilters` check. The `RemoteAsset` check ensures that assets are served by the Shopify CDN for optimal performance, and this is checked for in our newly added functions: `valueIsDefinitelyNotShopifyHosted` and `isUrlHostedbyShopify`.

We chose to integrate this check into the `AssetUrlFilters` check because they share similar logic and use the same traversing method. This approach allows us to leverage the existing code and maintain a clean codebase. If we had decided to make `RemoteAsset` into its own separate check, it would have been challenging to ensure that the two checks don't overlap. Integrating them into the same check allows us to handle potential overlaps more efficiently, as we can choose one approach or another based on the specific situation.

To differentiate between the two checks, we introduced different message reporting. For `RemoteAsset`, the message is: `Asset should be served by the Shopify CDN for better performance.` This message will help developers understand the performance implications of not using the Shopify CDN.

# Tophat screenshots
Screenshot 2023-10-04 at 1 55 18 PM
Screenshot 2023-10-04 at 1 57 28 PM

## Before you deploy

- [x] This PR includes a new checks


packages/theme-check-common/src/checks/asset-url-filters/index.ts
```diff
@@ -1,11 +1,26 @@
-import { HtmlRawNode, HtmlVoidElement, LiquidFilter, NodeTypes } from '@shopify/liquid-html-parser';
-import { LiquidCheckDefinition, LiquidHtmlNode, Severity, SourceCodeType } from '../../types';
+import {
```
charlespwd, line 1: We might want to rename the file from `asset-url-filters` to `remote-asset`, and also rename the check from AssetUrl to RemoteAsset.

charlespwd (changes requested):
Minor changes requested. Should also add `font_url` as an accepted filter. 

Otherwise code looks good!

charlespwd (approved):
LGTM 🙏

PR #126, opened 2023-10-02
Improve resilience of theme check pipeline by preventing entire failure if an individual check fails

karreiro:
Fixes https://github.com/Shopify/theme-check-js/issues/151

This PR addresses an issue where a crash in any check would lead to the failure of the entire pipeline, thereby impacting the execution of the remaining valid checks.

To mitigate this, the PR introduces the `onError` callback in the `Config` interface. This allows each channel (node, browser, etc.) to optionally define a unique callback to handle or report errors, enhancing error management capabilities.

By default, errors are still ignored, but they no longer impact the functioning of other checks.

**Before:**

https://github.com/Shopify/theme-tools/assets/1079279/1fdd7bd6-19a0-463f-892f-54bde634705e

**After:**

https://github.com/Shopify/theme-tools/assets/1079279/b48bf1bb-7f10-43fc-b22b-153ab365a601

## Before you deploy

- [x] This PR fixes a bug

### DIFF
--- a/.changeset/silent-baboons-confess.md
+++ b/.changeset/silent-baboons-confess.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-check-common': patch
+---
+
+Improve resilience of theme check pipeline by preventing entire failure if an individual check fails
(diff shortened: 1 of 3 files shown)


albchu (approved):
This is great. I especially like that we'll be able to capture errors in a bespoke way on OSW

PR #131, opened 2023-10-03
Add Shopify Reference links at the bottom of Liquid completion/hover items

charlespwd:
We do that for HTML, we should do it for our own stuff too!

- Add `Shopify Reference` links at the bottom of hover and completion items for Liquid Docset Entries.

https://github.com/Shopify/theme-tools/assets/4990691/cea437a3-65c0-4e16-8617-b711d017af4b

## Before you deploy


albchu (approved):
Rad this looks great! Brings us inline to the "MDN Reference" when hovering on html elements.

PR #163, opened 2023-10-10
Fix idempotence of script tags with backticks and Liquid in 'em

### LINKED ISSUE #162: New lines keep being added within script block
**Describe the bug**
`<script>` tag in liquid is adding additional new lines after opening tag and before closing tag on each run of the formatter. In other words, if the formatter runs on the file 5 times, there will be 5 blank lines before and after the contents of the script tag.

**Unformatted source**
```liquid
<script>
  if (condition) {
    ...
  }
</script>
```

**Expected output**
```liquid
<script>
  if (condition) {
    ...
  }
</script>
```

**Actual output**
```liquid
<script>

  if (condition) {
    ...
  }

</script>
# then on the next format run
<script>


  if (condition) {
    ...
  }


</script>
# and on and on
```

**Debugging information**
 - MacOS 13.4.1
 - "@shopify/prettier-plugin-liquid": "^1.0.6"

**Additional context**
I've only seen this on this script tag. Perhaps the condition block being the only JS within it is the issue?

charlespwd:

- Fixes the idempotence of pretty printing script tags with Liquid _and_ backticks in them.
  - We were wrapping the contents in new lines and that wasn't the desired behaviour.

Fixes #162

## before 
https://github.com/Shopify/theme-tools/assets/4990691/2adcca35-4c6f-4836-ba15-a58125b04bd9 
## after
https://github.com/Shopify/theme-tools/assets/4990691/475430d6-5062-4bef-977c-16fa3121b26c

## Before you deploy

- [x] This PR fixes a bug

### DIFF
--- a/packages/prettier-plugin-liquid/src/printer/printer-liquid-html.ts
+++ b/packages/prettier-plugin-liquid/src/printer/printer-liquid-html.ts
@@ -49,7 +49,7 @@ import { printClosingTagSuffix, printOpeningTagPrefix } from '~/printer/print/ta
 import { bodyLines, hasLineBreakInRange, isEmpty, isTextLikeNode, reindent } from '~/printer/utils';
 
 const { builders, utils } = doc;
-const { fill, group, hardline, indent, join, line, softline } = builders;
+const { fill, group, hardline, dedentToRoot, indent, join, line, softline } = builders;
 
 const oppositeQuotes = {
   '"': "'",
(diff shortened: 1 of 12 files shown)


packages/prettier-plugin-liquid/src/printer/printer-liquid-html.ts
```diff
@@ -228,17 +228,30 @@ function printNode(
...
-        : softline;
+      const body =
+        lines.length > 0 && lines.find((line) => line.trim() !== '')
+          ? join(hardline, reindent(lines, shouldSkipFirstLine))
+          : softline;
+
+      if (shouldIndentBody) {
+        return [indent([hardline, body]), hardline];
+      } else {
+        return [dedentToRoot([hardline, body]), hardline];
```
charlespwd, line 253: Had to duplicate this logic slightly because it's complicated to explain 😓. 

Basically the parent node would not know if we were able to embed or not (i.e. switch rendering from this plugin's to JS, or TypeScript, or whatever). But we need to know **if it failed** in order to remove the hardlines and print as is without the surrounding hardlines (which we do on line 231).

packages/prettier-plugin-liquid/src/printer/printer-liquid-html.ts
```diff
@@ -228,17 +228,30 @@ function printNode(
       };
 
       if (isRawMarkupIdentationSensitive()) {
-        return node.value;
+        return [node.value.trimEnd(), hardline];
```
charlespwd, line 231: This small difference makes the endtag indent properly.

PR #170, opened 2023-10-17
Verify region fold extra new lines problem

### LINKED ISSUE #169: Disable auto-formatting and auto-indenting
Since last update (11 oct. 2023), it seems the Shopify liquid extension is auto-formatting and auto-indenting the code without reason.

I use a region folding extension that wraps the code into regions, and it seems to trigger the file formatting which it never did before the Shopify liquid extension last update. I have `formatOnSave` settings set to false, it's only when I create a region folding through a shortcut that Shopify liquid extension formats the file. I don't see how this issue can be related to the folding extension, beside this extension was not updated recently, nor my settings.

Auto-formatting as shown in this screen recording: https://i.vgy.me/CnQc4e.gif

charlespwd:
Fixes #169


charlespwd:
Don't need to commit.

PR #177, opened 2023-10-18
Add LSP Completion support to codemirror-language-client

charlespwd:

- Add completion support to the CodeMirror language client lib
- Add `{ shouldLint, shouldComplete }` optional params to CodeMirror's `.extension` method.
  - This would let us disable linting without disabling completion and vice versa
- Also exposed the quick info renderer so that we may try rendering them in admin

Fixes Shopify/code-mirror-language-client#10

## Before you deploy

- [x] I included a `changeset` to this PR


albchu (approved):
This change looks great! 

How restricted are we to direct DOM manipulation in the infoRenderer logic?

Im wondering whether we can inject react components for different parts of the generated markdown or whether we'd need to style the raw elements ourselves.

charlespwd:
While I was there, I also added support for `textEdit` completion. Which makes cool stuff like the translation complete:


https://github.com/Shopify/theme-tools/assets/4990691/822e3762-8c53-4c90-b622-355409d5fb86

PR #209, opened 2023-10-26
Fix tsconfig project references

charlespwd:

- Fix the tsconfigs so that the code loaded by webpack is the code in src and not the one in dist
- Fix the tsconfigs so that the code changed in one project causes a refresh (codemirror playground)

## What did I learn?

- The `paths` compiler option is for telling TS where to find the code for some imports. We have to put the `/src` at the end of those or else we'd use the one in package.json (which points to dist).
- The `references` compiler options is a list of paths to folders with tsconfig files in them. It can _also_ be the path to a tsconfig file (say you wanted build to reference _build_ tsconfigs!)

## Before you deploy

- [x] This does not require a `changeset`

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,15 +1,16 @@
 {
-	"version": "2.0.0",
-	"tasks": [
+  "version": "2.0.0",
+  "tasks": [
     {
-      "type": "npm",
-      "script": "build:extension:dev",
-      "path": "packages/vscode-extension",
+      "type": "shell",
+      "command": "yarn --cwd packages/vscode-extension build:extension:dev",
       "group": "build",
-      "label": "npm: build:extension:dev - packages/vscode-extension",
-      "detail": "webpack --mode development",
+      "label": "vscode dev watch",
       "dependsOrder": "parallel",
-      "dependsOn": ["parser pre build", "theme check build"],
+      "dependsOn": [
+        "parser pre build",
+        "theme check build"
+      ],
     },
     {
       "label": "parser pre build",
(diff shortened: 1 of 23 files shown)


.vscode/tasks.json
```diff
@@ -1,15 +1,16 @@
...
     {
-      "type": "npm",
-      "script": "build:extension:dev",
-      "path": "packages/vscode-extension",
+      "type": "shell",
+      "command": "yarn --cwd packages/vscode-extension build:extension:dev",
       "group": "build",
-      "label": "npm: build:extension:dev - packages/vscode-extension",
-      "detail": "webpack --mode development",
+      "label": "vscode dev watch",
```
charlespwd, line 8: Ugh so I tried watch but couldn't get it to work... _yet_.

packages/theme-language-server-node/src/dependencies.ts
```diff
@@ -4,8 +4,7 @@ import { basename } from 'node:path';
 import * as fs from 'node:fs/promises';
 import { promisify } from 'node:util';
 import { glob as callbackGlob } from 'glob';
-import { Config, Translations } from '@shopify/theme-check-common';
-import { loadConfig as loadConfigFromPath } from '@shopify/theme-check-node';
+import { Config, Translations, loadConfig as loadConfigFromPath } from '@shopify/theme-check-node';
```
charlespwd, line 7: Because `@shopify/theme-check-common` wasn't a project reference of theme-language-server-node, we were loading the `dist` files here instead of the build ones. Meaning a `--watch` build would be using stale code if we were modifying any of those interfaces.

PR #235, opened 2023-11-02
We were making a bad assumption in there about the request IDs

charlespwd:

Fixup request IDs in LanguageClient. This was breaking hot reload in admin.

## Before you deploy


charlespwd:
Actually that doesn't quite solve our problem. Happens with the lifecycle and one worker. The initialize of the new client happens concurrently with the destroy of the other one. The new client gets the response for the shutdown request (!)

PR #247, opened 2023-11-15
Add `theme-liquid-docs [...] shopify_system_translations.json` to `theme-tools`

karreiro:
(wip)


.changeset/kind-birds-push.md
```diff
@@ -0,0 +1,7 @@
+---
+'@shopify/theme-language-server-common': patch
+'@shopify/theme-check-docs-updater': patch
+'@shopify/theme-check-common': patch
```
charlespwd, line 4: I believe this is minor given it kind of changes the public API.

packages/theme-check-docs-updater/src/themeLiquidDocsManager.spec.ts
```diff
@@ -96,6 +98,15 @@ describe('Module: ThemeLiquidDocsManager', async () => {
     });
   });
 
+  describe('Unit: systemTranslations', () => {
+    it('should return an array', async () => {
```
charlespwd, line 102:
```suggestion
    it('should return the parsed JSON content of the system translations', async () => {
```

packages/theme-check-docs-updater/src/themeLiquidDocsManager.ts
```diff
@@ -30,6 +31,10 @@ export class ThemeLiquidDocsManager implements ThemeDocset, JsonSchemaValidators
     return this.loadResource('tags', []);
   });
 
+  systemTranslations = memo(async (): Promise<Translations> => {
+    return this.loadResource('shopify_system_translations', []);
```
charlespwd, line 35: I think we want the default value to be `{}` here?

packages/theme-language-server-common/src/server/startServer.ts
```diff
@@ -88,7 +88,17 @@ export function startServer(
   const getTranslationsForURI: GetTranslationsForURI = async (uri) => {
     const rootURI = await findThemeRootURI(uri);
     const theme = documentManager.theme(rootURI);
-    return useBufferOrInjectedTranslations(getDefaultTranslationsFactory, theme, rootURI);
+
+    const getTranslationsFactory = (rootUri: string) => {
+      return async () => {
+        const defaultTranslations = await getDefaultTranslationsFactory(rootUri)();
+        const shopifyTranslations = await themeDocset.systemTranslations();
```
charlespwd, line 95: Minor nit since this is sequential instead of "parallel while waiting for IO." [`Promise.all`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) would make this run in parallel :D 

```suggestion
        const [defaultTranslations, shopifyTranslations] = await Promise.all([
          getDefaultTranslationsFactory(rootUri)(),
          themeDocset.systemTranslations(),
        ]);
```

packages/theme-check-docs-updater/src/themeLiquidDocsManager.ts
karreiro: Great catch! Fixed :)

PR #248, opened 2023-11-16
Add support for `themeCheck.checkOn{Open,Change,Save}` configurations

charlespwd:

- Add support for `themeCheck.checkOn{Open,Save,Change}` Language Server configurations

It's a parity feature. We don't really need it but in case someone doesn't want to have squiggly lines all the time, they can toggle that stuff off.

## Before you deploy


packages/theme-language-server-common/src/ClientCapabilities.ts
```diff
@@ -1,13 +1,47 @@
...
+    // { 'themeCheck.checkOnSave': true }
+    const direct = this.initializationOptions?.[key];
+    if (direct !== undefined) return direct;
+
+    // { themeCheck: { checkOnSave: true } }
+    const nested = pathGet<T>(this.initializationOptions, key);
+    if (nested !== undefined) return nested;
+
+    // fallback
+    return defaultValue;
```
charlespwd, line 40: This is how neovim does it. It's a port of the code I had in theme-check-ruby

PR #253, opened 2023-11-22
Improve support for unclosed HTML nodes inside branching code

### LINKED ISSUE #196: LiquidHTMLSyntaxError might be too strict
Hi,

I am unsure this can be solved, but maybe you have an idea.

In our themes, for some sections, I am doing some complex Liquid processing where I am opening and closing a div conditionally based on a lot of tricky logic:

```
{%- if block.type == 'button' and is_inside_button_group_context == false -%}
  {%- assign is_inside_button_group_context = true -%}
  <div class="button-group justify-{{ section.settings.text_position }}">
{%- elsif block.type != 'button' and is_inside_button_group_context -%}
  {%- assign is_inside_button_group_context = false -%}
  </div>
{%- endif -%}
```

With this code, I end up having this error:

image

charlespwd:

This PR is about adding prettier/LiquidHTMLSyntaxError support for _unclosed HTML element nodes inside branching code_.

It turns out that the following use case is rather common and wasn't supported before this PR:

```liquid
{% if some_condition %}
  <details>
    <summary>
{% else %}
  <h2>
{% endif %}

...summary/header body...

{% if some_condition %}
  {% render 'some icon' %}
  </summary>
{% else %}
  </h2>
{% endif %}

...body of maybe accordion..

{% if some_condition %}
  </details>
{% endif %}
```

This PR adds prettier support (example above is the logic we'd use to pretty print)

What happens in the PR:
- Add full support for unclosed HTML elements inside branching code
- Allow 'HtmlDanglingMarkerClose' nodes to have non-dangling marker siblings
- Move the `prettier-plugin-liquid/test` folder insider `src` so that vitest correctly runs the code on the latest code and not on the `dist` code.

Fixes #196 
Fixes https://github.com/Shopify/prettier-plugin-liquid/issues/213

## screnshot of a diff in dawn

22-57-rxpuq-k8jf8

- See if we can add an `UnclosedHtmlElement` check that verifies that the stacks are correctly closed.

## What did you learn?

- Wasn't as _hard_ as I thought it'd be? Definitely hard though.

## Before you deploy

- [x] I included a `changeset`s 
  - a major for `@shopify/liquid-html-parser` since it removes `HtmlDanglingMarkerOpen`
  - minor for prettier plugin (expands what we can pretty-print)


packages/liquid-html-parser/src/stage-2-ast.ts
```diff
@@ -850,12 +824,15 @@ export function toLiquidHtmlAST(
...
 
-type LiquidCSTNode = LiquidCST[number];
+type ConcreteCloseNode = ConcreteLiquidTagClose | ConcreteHtmlTagClose;
 
 class ASTBuilder {
   /** the AST is what we're building incrementally */
   ast: LiquidHtmlNode[];
 
+  /** because we might be doing fun stuff */
+  options: ASTBuildOptions;
```
charlespwd, line 834: Didn't use this after all. Should remove.

packages/liquid-html-parser/src/stage-2-ast.ts
```diff
@@ -1915,15 +1933,15 @@ function isAcceptableDanglingMarkerClose(
...
+  return isAcceptableDanglingMarker(builder);
 }
 
 // This function checks that the builder.current node accepts dangling nodes.
 //
 // The current logic is:
 //  - Grandparent node must be an if-like statement
 //  - Parent node must be a LiquidBranch
-//  - All sibling nodes must be of the same type (all close or all open)
+//  - All sibling nodes must be flat
```
charlespwd, line 1944: Not true, it's even more relaxed than this. It's the immediate parent of the dangling close must be a branch. That covers it and allows for 
```liquid
{% if cond %}
  <svg>...</svg>
  {% # works because the dangling is an immediate child of the branch %}
  </summary>
{% endif %}
```

packages/prettier-plugin-liquid/src/printer/print/tag.ts
```diff
@@ -386,6 +310,96 @@ export function printOpeningTagStartMarker(node: LiquidHtmlNode | undefined) {
...
+  }
+
+  switch (node.type) {
+    // case 'ieConditionalComment':
+    // case 'ieConditionalEndComment':
+    //   return '[endif]-->';
+    // case 'ieConditionalStartComment':
+    //   return ']><!-->';
+    // case 'interpolation':
+    //   return '}}';
```
albchu, line 388: cleanup?

PR #254, opened 2023-11-24
Add `UnclosedHTMLElement` check

charlespwd:

- Add the `UnclosedHTMLElement` check

  This check warns users of unclosed HTML elements in branching code (or wrongly closed) by making sure that they are opened/closed in the same parent element, condition and stack order.

  Again but slower :sweat_smile:

  - Dangling open/close must be in conditional branches of the same parent
    ```liquid
    no good because the details is closed in the `<p>` and not the `<div>`
    <div>
      {% if cond %}
        <details>
      {% endif %}
      <p>
        {% if cond %}
          </details>
        {% endif %}
      </p>
    </div>
    ```
  - They must open/close in the same conditional branch identifier (same if/unless condition)

    ```liquid
    no good! different conditions open/close the tags!
    <div>
      {% if something %}
        <details>
      {% endif %}

      {% if something_else %}
        </details>
      {% endif %}
    </div>
    ```
    
    - `else` and `unless` evaluate to the negated condition identifier

      ```liquid
      Works! Same condition
      <div>
        {% if cond %}
        {% else %}
          <details>
        {% endif %}

        {% unless cond %}
          </details>
        {% endunless %}
      </div>
      ```

  - They must open/close in the correct order (<details><summary></summary></details>)

    ```liquid
    not good (should close in reverse order)
    <div>
      {% if cond %}
        <details
          <summary>
      {% endif %}

      {% if cond %}
        </details>
        </summary>
      {% endif %}
    </div>
    ```

- [ ] Add docs for UnclosedHTMLElement

## What did you learn?

Surprised! That wasn't as hard as I thought! It's clever af, but it works!

## Before you deploy

  - [x] I ran `yarn update-configs` and committed the updated configuration files

### DIFF
--- a/packages/theme-check-common/src/checks/unclosed-html-element/index.spec.ts
+++ b/packages/theme-check-common/src/checks/unclosed-html-element/index.spec.ts
@@ -1,0 +1,122 @@
+import { expect, describe, it } from 'vitest';
+import { UnclosedHTMLElement } from '.';
+import { runLiquidCheck, highlightedOffenses } from '../../test';
+
+describe('Module: UnclosedHTMLElement', () => {
+  it('should report an offense for a branch open tag without its close tag', async () => {
+    const file = `
+      <div>
+        {% if cond %}
+          <main>
+        {% endif %}
+      </div>
+    `;
+
+    const offenses = await runLiquidCheck(UnclosedHTMLElement, file);
+
+    expect(offenses).to.have.length(1);
+    expect(offenses).to.containOffense({
+      message: `Opening tag does not have a matching closing tag for condition \`cond\` in HtmlElement 'div'`,
+    });
+    expect(highlightedOffenses(file, offenses)).to.include('<main>');
+  });
+
+  it('should report an offense for a branch close tag without its open tag', async () => {
+    const file = `
+      <div>
+        {% if cond %}
+          </main>
+        {% endif %}
+      </div>
+    `;
+
+    const offenses = await runLiquidCheck(UnclosedHTMLElement, file);
+
+    expect(offenses).to.have.length(1);
+    expect(offenses).to.containOffense({
+      message: `Closing tag does not have a matching opening tag for condition \`cond\` in HtmlElement 'div'`,
+    });
+    expect(highlightedOffenses(file, offenses)).to.include('</main>');
+  });
+
+  it('should report offenses when branches are closed in the wrong order', async () => {
+    const file = `
+      <div>
+        {% if cond %}
+          <details>
+            <summary>
+        {% endif %}
+
+        {% if cond %}
+          </details>
+          </summary>
+        {% endif %}
+      </div>
+    `;
+
+    const offenses = await runLiquidCheck(UnclosedHTMLElement, file);
+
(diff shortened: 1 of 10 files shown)


packages/theme-check-common/src/checks/unclosed-html-element/index.ts
```diff
@@ -0,0 +1,251 @@
...
+    case 'else':
+      switch (parent.name) {
+        case NamedTags.if:
+          return negateIdentifier(getConditionIdentifierForMarkup(parent.markup));
+        case NamedTags.unless:
+          return getConditionIdentifierForMarkup(parent.markup);
+        case NamedTags.case:
+          return `case ${getConditionIdentifierForMarkup(parent.markup)}`;
+        default:
+          '??';
```
charlespwd, line 187: Should prob add tests / functionality for weird edge case?

packages/theme-check-common/src/checks/unclosed-html-element/index.spec.ts
```diff
@@ -0,0 +1,122 @@
+import { expect, describe, it } from 'vitest';
+import { UnclosedHTMLElement } from '.';
+import { runLiquidCheck, highlightedOffenses } from '../../test';
+
+describe('Module: UnclosedHTMLElement', () => {
+  it('should report an offense for a branch open tag without its close tag', async () => {
```
charlespwd, line 6: Should add tests for case statements

karreiro (approved):
Thank you, @charlespwd! Very cool stuff! 🚀 I've noticed two minor issues, but I believe we may handle them in a separate future PR:

1. I believe we could handle dependent branches:

https://github.com/Shopify/theme-tools/assets/1079279/74aa82f0-516b-4131-aa6b-ee15564094ae

2. I don't think we're reporting unclosed tags when the number of unclosed elements is even:

https://github.com/Shopify/theme-tools/assets/1079279/9d2c0bf1-fbaa-4b47-95ed-f67a2569f9a8

Both scenarios are very specific corner cases tho. Thanks again for this PR!

haribshahbaz:
> 3\. I don't think we're reporting unclosed tags when the number of unclosed elements is even:

thats interesting. i believe what is happening here is if there's an even number of unbalanced tags, and they are in pairs of matching opening and closing tags, they will all be popped off the stack and none will be reported as unbalanced. to fix this we might need to modify the code to also consider the branch or condition each tag is in.

charlespwd:
@karreiro 

> I believe we could handle dependent branches:

Oh man you're making me work hard lol. I'm inclined to leave it as a warning for those use cases to be ignored. Did you have a solution in mind?

> I don't think we're reporting unclosed tags when the number of unclosed elements is even:

That's f'ed up. Need to find out why.

karreiro:
> Oh man you're making me work hard lol. I'm inclined to leave it as a warning for those use cases to be ignored

🤣 100% -- I feel both corner cases can be solved in the future, and we may leave a warning until there

> Did you have a solution in mind?

I'm not so sure if it's the best approach, but I believe we could set the dependency tree based on the branch `blockStartPosition`

PR #294, opened 2024-02-16
fix: prettier plugin liquid number starting index variable lookup

BJvdA:

When prettifying liquid the following happens currently:
`{{ x['20back'] }}` turns into `{{ x.20back }}`

This causes the following error: _Expected id but found number_
This PR adds logic to make sure index access strings starting with numbers are not converted to dot notation


BJvdA:
I have signed the CLA!

PR #297, opened 2024-02-19
Improve YAML parsing error message for unresolved aliases

### LINKED ISSUE #276: VS Code Error: Request textDocument/documentLink failed
**Describe the bug**
With the latest version of Shopify Liquid (VS Code Extension) I now get this error message, which is a Theme Check Language Server issue. I'm running Windows 11. 

The error appears in the Output tab:
```
Request textDocument/documentLink failed.
  Message: Request textDocument/documentLink failed with message: Unresolved alias (the anchor must be set before the alias): .code-workspace
  Code: -32603 
```

On the right hand side a popup appears each time I open any file, the error reads `Request textDocument/documentLink failed.` and the Source: `Shopify Liquid`

When I open VS Code if there are files open, then each file open causes this error message to appear. So there's a cascade of errors that appear in quick succession.

I tried adding this to my VS Code settings, but it caused the server to crash permanently:
```
  "shopifyLiquid.legacyMode": true,
  "themeCheck.onlySingleFileChecks": true,
  "shopifyLiquid.languageServerPath": "windows-path-removed-for-security/theme-check-language-server.bat",
```

**Debugging information**
 - OS Windows 11 (latest)
 - Shopify Liquid v2.0.0
 - Shopify CLI v3.55.3

charlespwd:
Fixes #276

- Upgrades vitest to v1.3+
- Deduplicates dependencies
- Fixes debugging issues that happened because the vitest VS Code extension was expecting vitest v1.0+.

## Before you deploy

### DIFF
--- a/.changeset/silver-terms-happen.md
+++ b/.changeset/silver-terms-happen.md
@@ -1,0 +1,23 @@
+---
+'@shopify/theme-check-node': patch
+---
+
+Improve YAML parsing error messages caused by YAML aliases
+
+```yaml
+ignore:
+  # this causes an error because * at the start of a statement in YAML
+  # is used for aliases to anchors such as `&.code-workspaces`
+  - *.code-workspaces
+
+  # what you want is this
+  - '*.code-worksapces'
+```
+
+It’s a pretty obscure error for the non-initiated, so now we’ll
+instead throw an error like this:
+
+```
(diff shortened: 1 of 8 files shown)


charlespwd:
Ugh vitest 1.3 is _much_ slower than 0.3

Issue #296, opened 2024-02-16
checkout.liquid objects


charlespwd:
**Describe the bug**
Undefined object, hover, etc does not work for checkout.liquid stuff

We should make it work, but only for `layout/checkout.liquid`.

https://shopify.dev/docs/themes/architecture/layouts/checkout-liquid

jagmeetc7:
One more thing merchant added:

 {{ brand_principles_styles }} . If you look into this, that will be great. I believe this is also required, but somehow Shopify CLI showing undefined object for this.

jagmeetc7:
Hey @charlespwd are you asking if the merchant is following any dev docs for this issue they are dealing with?

jagmeetc7:
@charlespwd merchant responded confirming - Yes, {{ brand_principles_styles }} is user defined variable and values does coming inside it. For now, they’re not using the values coming inside this object. So, they will remove it from their code base.

PR #301, opened 2024-02-21
Add Language Server and Theme Check support for checkout.liquid objects

jamesmengo:
- Fixes: [legacy checkout objects being reported as undefined](https://github.com/Shopify/theme-tools/issues/296)
- Adds: Hover tooltip for legacy checkout objects (all strings at the moment)

- Add support for `checkout.liquid` [objects](https://shopify.dev/docs/themes/architecture/layouts/checkout-liquid) in theme-check and language server
- These objects are only valid in `checkout.liquid`

## What did you learn?
- This entire area is new to me - I found it interesting how create a dictionary of top-level objects so that we can have nested types such as `checkout.applied_discounts.x.y.z`, where all additional types must be defined at the top level Docset
- Pretty cool how we auto configure types by referencing the Docset

## Before you deploy
  - [ ] I ran `yarn update-configs` and committed the updated configuration files
    
    
    

packages/theme-language-server-common/src/TypeSystem.spec.ts
```diff
@@ -307,6 +312,7 @@ describe('Module: TypeSystem', () => {
       ['recommendations', 'sections/recommendations.liquid'],
       ['app', 'blocks/recommendations.liquid'],
       ['app', 'snippets/recommendations.liquid'],
+      ['locale', 'layout/checkout.liquid'],
```
jamesmengo, line 315: just checking a single value for the checkout.liquid path

lukeh-shopify (approved):
👏🏻 🥳

PR #314, opened 2024-03-01
Fix weird bug in AssetSizeAppBlock{CSS,JavaScript}

charlespwd:
We were getting "Can't find file size" for assets/app.js errors being
logged. That was because the function wasn't being called with an
absolute path but with a relative path.

For fuck's sake. I refactored our root finding algo for a long time
thinking it was the cause. It wasn't.

It was this fucking dumb implementation bug.

Fixes Shopify/cli#3497

## Before you deploy


charlespwd:
Fixes Shopify/cli#3344

PR #322, opened 2024-03-10
Fixed - Theme Check: Issue of 0 files being inspected.

justshrey:

Fixes shopify/cli#3497

The main issue was that on windows, no liquid or json files were being selected despite `'**/*.{liquid,json}'` as the glob function. 

this happens because in windows, the path separator output by path.join is '\' which conflicts with the glob escape character. 

To solve it, have normalized the path and just replaced '\' with  '/' . This should solve this since it wont have any impact on existing linux paths. 

N/A - 

## What did you learn?
image

### DIFF
--- a/packages/theme-check-node/src/index.ts
+++ b/packages/theme-check-node/src/index.ts
@@ -102,7 +102,15 @@ async function getThemeAndConfig(
 }
 
 export async function getTheme(config: Config): Promise<Theme> {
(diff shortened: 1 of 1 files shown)


justshrey:
I have signed the CLA!

charlespwd:
You are amazing. This was on my TODO list but couldn't get around doing it with everything else on there. Thank you!

I'll add a changeset for a patch bump and see that this gets into the CLI ASAP. 

🙏

charlespwd (approved):

PR #331, opened 2024-03-26
Get rid of AJV, long live vscode-json-languageservice.

charlespwd:

This is a big refactor.

- Removes `ajv` dependency (replaced by `vscode-json-languageservice`)
- Breaking: Removes `theme-docs compile-schema`
- Breaking: Redesigns the `jsonValidationSet` public API
- Adds "bring-your-own-json-schema" support
- Adds "$ref" support to schemas for code reuse :D 
- Adds `ValidJSON` check

Fixes Shopify/develop-advanced-edits#145

## Before you deploy

  - [x] I ran `yarn update-configs` and committed the updated configuration files

- [ ] My feature is backward compatible: Yes and no (but it's documented!) It isn't for the browser-dependencies.


packages/theme-check-common/src/checks/valid-schema/index.ts
```diff
@@ -21,88 +16,25 @@ export const ValidSchema: LiquidCheckDefinition = {
...
+        for (const problem of problems) {
           context.report({
-            message,
-            ...errorIndicies,
+            message: problem.message,
+            startIndex: node.blockStartPosition.end + problem.startIndex,
+            endIndex: node.blockStartPosition.end + problem.endIndex,
           });
-        });
+        }
```
charlespwd, line 37: I'm very proud of this cleanup. So much simpler.

packages/theme-check-common/src/checks/valid-schema/parse-json-body.ts
```diff
@@ -1,77 +0,0 @@
-import { LiquidRawTag, Position } from '@shopify/liquid-html-parser';
```
charlespwd, line 1: bye

packages/theme-check-common/src/types/theme-liquid-docs.ts
```diff
@@ -20,32 +20,46 @@ export interface ThemeDocset {
...
+   * How would I do that in admin? Just plug it in? I mean that shit
+   * should be optional, no? How do I switch theme/app blocks filematch btn contexts?
+   * That _could_ be a .theme-check.yml thing? Which would mean I'd need a
+   * schemas per rootURI... yuk.
+   *
+   * If I make it an array, then I make it the implementer's job to know
+   * what URLs/filePatterns to match. Which is also YUK.
+   *
+   * But if I do it this way, then I can make it resolve "chains" rather
+   * neatly.
```
charlespwd, line 60: Oops should get rid of rubber duck comment 🙃

packages/theme-check-docs-updater/package.json
```diff
@@ -31,7 +31,6 @@
   },
   "dependencies": {
     "@shopify/theme-check-common": "^2.3.0",
-    "ajv": "^8.12.0",
```
charlespwd, line 34: bye

packages/theme-check-common/src/JSONValidator.ts
```diff
@@ -0,0 +1,52 @@
+import { LanguageService, TextDocument, getLanguageService } from 'vscode-json-languageservice';
+import { JsonValidationSet, SourceCodeType, ValidateJSON } from './types';
+
+/** This simply memoizes the result to prevent over fetching */
```
charlespwd, line 4: Stale comment

karreiro (approved):
Thank you, @charlespwd! Great stuff 🔥  I've left only one minor question about the `JSONDocument` object and the document manager :)

packages/theme-check-common/src/JSONValidator.ts
```diff
@@ -0,0 +1,52 @@
...
+   * (presumably by doing some offset logic)
+   */
+  public validate: ValidateJSON<SourceCodeType> = async (sourceCode, jsonString) => {
+    const jsonTextDocument = TextDocument.create(
+      'file:' + sourceCode.absolutePath,
+      'json',
+      0,
+      jsonString,
+    );
+    const jsonDocument = this.service.parseJSONDocument(jsonTextDocument);
```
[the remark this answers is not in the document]
charlespwd: This here is kind of subtle. For one, there is no document manager in theme-check. We only have an array of SourceCode to deal with. 

We _could_ add those to the interface of SourceCode, but I think it would be rather weird. For one, in Liquid files, this only gets called with the body of `{% schema %}` tags. In JSON files, we use the contents as is. Even then, there's quite a bit of subtlety about offsetting the startIndex with the endIndex of the {% schema %} blockStart. 

I'm not 100% sure how we would put that on the SourceCode interface tbh. Right now it's this:
```ts
export type SourceCode<T = SourceCodeType> = T extends SourceCodeType
  ? {
      /** A normalized absolute path to the file. Assumes forwards slashes. */
      absolutePath: string;
      /** The type is used as a discriminant for type narrowing */
      type: T;
      /** The version is used by the Language Server to make sure the Client and Server are in sync */
      version?: number;
      /** The contents of the file */
      source: string;
      /** The AST representation of the file, or an Error instance when the file is unparseable */
      ast: AST[T] | Error;
    }
  : never;
```

If we were to preparse section files, we'd need a second ast attribute in there. But we'd need to have that attribute for all files, even non-section files.

PR #338, opened 2024-04-03
Add support for schema manifests depending on context

charlespwd:

Lots of internal breaking changes. See changesets.

- Add required `context: 'app' | 'theme'` property to `Config` type
- Unhardcode the `section` and `translations` schemas into a system that depends on the Shopify/theme-tools manifest files
  - [manifest_theme.json](https://github.com/Shopify/theme-liquid-docs/blob/main/schemas/manifest_theme.json)
  - [manifest_theme_app_extension.json](https://github.com/Shopify/theme-liquid-docs/blob/main/schemas/manifest_theme.json)
- We should be able to push new schemas without making new releases
  - (before it was edits without releases, now it's new _files_ without releases)
- Added support for relative `$ref` in the JSON schemas.
  - At the time of this writting, the `section.json` schema refers to `./input-settings.json` file in its rules.
  - This is setup so that I can reuse `input-settings.json` in the theme block JSON schema, theme app extension block JSON schema, and `config/settings_schema.json` JSON schema.
- Changed my mind about the `jsonValidationSet.schemas` being static.
  - For monorepo support, and to support theme app extension mode overload of the `blocks/*.liquid` JSON schema file match collision, we need that thing to be a `(context: 'app' | 'theme') => Promise<SchemaDefinition>[]`

## What did you learn?

- Got burned by trying to take a shortcut. The `getDefaultTranslationsForURI` style APIs are annoying, but they make monorepos work.

## Before you deploy


packages/theme-check-docs-updater/src/themeLiquidDocsDownloader.ts
```diff
@@ -1,44 +1,76 @@
...
 export type Resource = (typeof Resources)[number];
 export const Resources = [
   'filters',
   'objects',
   'tags',
-  'section_schema',
-  'translations_schema',
   'shopify_system_translations',
+  'manifest_theme',
+  'manifest_theme_app_extension',
```
charlespwd, line 21: Rather than hardcoded schema URLs, we have hardcoded manifests URLs. This means we can change the manifests without requiring new releases.

karreiro (approved):
Thank you, @charlespwd 🔥

PR #341, opened 2024-04-08
Prevent loading the same third party check more than once

### LINKED ISSUE #336: [Theme Check] Possible to apply the same plugin multiple times
**Describe the bug**
Plugins can be applied more than once. This produces duplicate offences in Theme Check and prevents auto fixes from being applied.

This issue is avoidable by removing the `require` key from the config, but may cause issues if users are extending multiple configs.

**Source**
```yaml
require:
  - theme-check-plugin-schema

Schema/RequireImageDimensions/Liquid:
  enabled: true
Schema/RequireImageDimensions/Json:
  enabled: true
```

**Expected behaviour**
Checks should only be run once

image

**Actual behaviour**
Checks are being run twice (once per time included in the `require` array, plus once for being auto discovered)

image

With `--auto-correct` flag

image

**Debugging information**
 - OS: macOS
 - OS Version: Sonoma 14.4
 - Theme Check Version: 2.2.2 & 2.4.0
 - Theme CLI Version: 3.58.0

I've included the Theme CLI version here because I first discovered the issue when testing Theme Check via the Theme CLI. I've since installed `@shopify/theme-check-node` and encountered the same issue.

**Additional context**
I discovered this during testing because I was unaware third party checks (for example [packages in node modules beginning `theme-check-`](https://github.com/Shopify/theme-tools/blob/main/packages/theme-check-node/src/config/load-third-party-checks.ts#L48)) [are auto discovered](https://github.com/Shopify/theme-tools/blob/main/packages/theme-check-node/src/config/load-config-description.ts#L31) and was using the `require` config option.

I may have just missed it, but auto discovery of plugins doesn't seem to be documented. It definitely seems worth shouting about as a handy feature, especially as it encourages people to following the naming convention for their packages.

**Replication Steps**
1. Clone [https://github.com/davidwarrington/theme-check-plugin-schema](https://github.com/davidwarrington/theme-check-plugin-schema/tree/feature/build)
2. Checkout to `feature/build` branch
3. Install dependencies for the above and run `pnpm tsup` to build so it can be imported into a project
4. Clone [https://github.com/davidwarrington/theme-check-duplicate-plugins](https://github.com/davidwarrington/theme-check-duplicate-plugins)
5. Install the plugin repo you cloned and built in steps 1, 2 and 3 (`pnpm install ../path/to/repo` if using pnpm)
6. Run `shopify theme check` and note duplicate error reports
7. Run `shopify theme check --auto-correct` and note an error because of overlapping fix ranges
8. Run `node ./check.js` and note duplicate offenses in the output

davidwarrington:

To prevent accidentally loading the same plugin more than once, which in turn prevents auto fixes from running.

Fixes #336 

## What did you learn?

Not exciting for anyone else, but I learned how the Theme Check config is resolved and loaded.

## Before you deploy

### DIFF
--- a/.changeset/flat-shoes-sparkle.md
+++ b/.changeset/flat-shoes-sparkle.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-check-node': patch
+---
+
+Prevent loading the same plugin more than once
--- a/packages/theme-check-node/src/config/load-third-party-checks.ts
+++ b/packages/theme-check-node/src/config/load-third-party-checks.ts
@@ -18,7 +18,7 @@ export function loadThirdPartyChecks(
    * */
   modulePaths: ModulePath[] = [],
 ): CheckDefinition<SourceCodeType>[] {
-  const checks = [];
+  const checks = new Set<CheckDefinition<SourceCodeType>>();
   for (const modulePath of modulePaths) {
     try {
       const moduleValue = require(/* webpackIgnore: true */ modulePath);
@@ -31,7 +31,7 @@ export function loadThirdPartyChecks(
 
       for (const check of moduleChecks) {
         if (isCheckDefinition(check)) {
-          checks.push(check);
+          checks.add(check);
         } else {
           console.error(`Expected ${check} to be a CheckDefinition, but it looks like it isn't`);
         }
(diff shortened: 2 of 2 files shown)


charlespwd:
Ran the tests locally. Should be good! TY for contribution :)

PR #344, opened 2024-04-11
Add toggle-able Vim mode to CodeMirror playground

jamesmengo:
## Before you deploy


.changeset/silly-trees-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/codemirror-language-client': minor
+---
+
+Add toggle-able Vim mode to CodeMirror playground
```
charlespwd, line 5: This doesn't require a new release of the CMLC to npm, no need for a changeset.

packages/codemirror-language-client/playground/src/playground.ts
```diff
@@ -120,10 +121,41 @@ async function main() {
     params: exampleTranslations,
   } as SetDefaultTranslationsNotification.type);
 
-  new EditorView({
+  class VimConfigurator {
+    private static vimStateFacet = Facet.define<boolean, boolean>({
+      combine: (values) => values[0],
+    });
```
charlespwd, line 127: I think that in this setup, you don't even need the facet. The state could be held "outside" codemirror. OW we're kind of forcing a contract that the state must be sync'ed across all three states. 

I think we can simplify it a little more. You don't have to, but if you feel like it I feel like you can :P 

The dispatch could simply do a reconfigure across all three without the internal state. Just put the state on this singleton :P 

Actually, you might not even need the class. You could do it inline with scoping.

jamesmengo: good point! 
I wanted to group this code somehow and ended up making it more complex than needed. I've simplified this a bit and pushed up those changes

charlespwd (approved):
There you go! Simple!

Issue #345, opened 2024-04-12
="" not automatically added with HTML attributes


stijns96:
**Describe the bug**
When I type `clas`, the editor will suggest me to auto complete `class`. Normally it's smart enough to add `=""`` as well where your cursor will be placed within `""`. This is not the case. It only autocompletes the word, but without `=""`.

**Source**

```liquid
<div class></div>
```

**Expected behaviour**
When I hit <kbd>tab</kbd> when a suggestion pops up, I expect that it will be completed with `=""` at the end with my cursor inside `""`.

**Actual behaviour**
When I hit <kbd>tab</kbd> when a suggestion pops up, it only autocompletes the word.

**Debugging information**
 - OS Mac
 - OS 14.2.1 (23C71)
 - Theme Check Version v2.3.0 (extension)

**Additional context**
I created my [own extension](https://github.com/ways-agency/shopify-liquid-snippets/tree/main) ones which is working correctly. At that time, HTML autocompletion was not working yet with the Shopify extension. Since it is kind of working now, I got the suggestions twice.

Might be worth to checkout `vscode-html-languageservice` as well.

stijns96:
Hi @charlespwd,

I try to contribute here and figured out what to do. The only thing is that debugging is quite hard... I added the following (WIP), but this should do the easy trick.

[HtmlAttributeCompletionProvider](https://github.com/Shopify/theme-tools/blob/main/packages/theme-language-server-common/src/completions/providers/HtmlAttributeCompletionProvider.ts#L47)

```javascript
function toCompletionItem(tag: Attribute): CompletionItem {
  return {
    label: tag.name,
    kind: CompletionItemKind.Value,
    insertText: tag.valueSet !== 'v' ? `${tag.name}="$1"` : tag.name,
    documentation: {
      kind: 'markdown',
      value: renderHtmlEntry(tag),
    },
  };
}
```

I did a format -> build -> test and have it running with pressing <kbd>F5</kbd>. I create a simple liquid file and type `<button t></button>`. In my debugger I get my consoles, but somehow the server is erroring.

Any idea why? When I remove my edits, it works fine.

<img width="1092" alt="image" src="https://github.com/Shopify/theme-tools/assets/47820758/4d89bf93-f486-4823-ac8a-ba816953a8db">

<img width="1079" alt="image" src="https://github.com/Shopify/theme-tools/assets/47820758/dfc28b3b-bed3-45b9-aeb9-1694fa541a99">

PR #350, opened 2024-04-17
Add a prettier configuration option for captureWhitespace

### LINKED ISSUE #348: Add a prettier configuration option for `captureWhitespace`

`{% capture var %}` is _technically_ whitespace sensitive. It captures its contents to a variable. Because of this and because we should not to change the semantic meaning of the file with default options, we were forced to use "paste-as-is" logic for the contents of capture tags. 

It has been reported by partners that this level of strictness is not always necessary. You can turn that feature off by using a `white-space: normal` comment above the capture tag, but some folks would want this by default and the other option on-demand. 

@stijns96 proposed having a prettier config that switch the default for that. I like that idea. 

- Add a new prettier configuration setting that changes the css-whitespace property of the capture tag. Something like this:

```
"captureWhitespace": "strict" | "ignore"
```

Reusing the nomenclature for `htmlWhitespaceSensitivity`. 
- `strict` (default), whitespace is considered significant and will be maintained as pre
- `ignore`, whitespace will not be considered significant. Same as `{% # white-space: normal %}` before capture tags

**Additional context**
Breadcrumbs:
- [Where we define our configuration variables](https://github.com/Shopify/theme-tools/blob/main/packages/prettier-plugin-liquid/src/plugin.ts#L17-L47)
- [Where we would want to switch to a different default for capture tags depending on the config value](https://github.com/Shopify/theme-tools/blob/main/packages/prettier-plugin-liquid/src/printer/preprocess/augment-with-css-properties.ts#L180). Note that that function doesn't accept options like the CssDisplay one above, we should change that.

stijns96:

This PR closes #348. You can now add `"captureWhitespace": "ignore|strict"` to the prettier config to set the global whitespace sensitivity.

Fix the HTML attribute snippet completion (`=""`) - ([345](https://github.com/Shopify/theme-tools/issues/345))

## Before you deploy

  - [x ] I ran `yarn build` and committed the updated configuration files

- [x ] My feature is backward compatible

### DIFF
--- a/packages/prettier-plugin-liquid/src/printer/preprocess/augment-with-css-properties.ts
+++ b/packages/prettier-plugin-liquid/src/printer/preprocess/augment-with-css-properties.ts
@@ -134,7 +134,10 @@ function getCssDisplay(node: AugmentedNode<WithSiblings>, options: LiquidParserO
   }
 }
 
-function getNodeCssStyleWhiteSpace(node: AugmentedNode<WithSiblings>): string {
+function getNodeCssStyleWhiteSpace(
+  node: AugmentedNode<WithSiblings>,
+  options: LiquidParserOptions,
+): string {
   if (node.prev && node.prev.type === NodeTypes.HtmlComment) {
     // <!-- white-space: normal -->
     const whitespace = getCssWhitespaceFromComment(node.prev.body);
@@ -177,7 +180,13 @@ function getNodeCssStyleWhiteSpace(node: AugmentedNode<WithSiblings>): string {
       return 'pre';
 
     case NodeTypes.LiquidTag:
-      return CSS_WHITE_SPACE_LIQUID_TAGS[node.name] || CSS_WHITE_SPACE_DEFAULT;
+      console.log(options.captureWhitespace);
+      switch (options.captureWhitespace) {
+        case 'strict':
+          return CSS_WHITE_SPACE_LIQUID_TAGS[node.name] || 'pre';
+        case 'ignore':
+          return CSS_WHITE_SPACE_LIQUID_TAGS[node.name] || 'normal';
+      }
 
     case NodeTypes.LiquidBranch:
     case NodeTypes.LiquidVariableOutput:
@@ -222,7 +231,7 @@ function getNodeCssStyleWhiteSpace(node: AugmentedNode<WithSiblings>): string {
 export const augmentWithCSSProperties: Augment<WithSiblings> = (options, node) => {
   const augmentations: WithCssProperties = {
     cssDisplay: getCssDisplay(node, options),
-    cssWhitespace: getNodeCssStyleWhiteSpace(node),
+    cssWhitespace: getNodeCssStyleWhiteSpace(node, options),
   };
 
   Object.assign(node, augmentations);
--- a/packages/prettier-plugin-liquid/src/printer/preprocess/augment-with-css-properties.ts
+++ b/packages/prettier-plugin-liquid/src/printer/preprocess/augment-with-css-properties.ts
@@ -180,15 +180,11 @@ function getNodeCssStyleWhiteSpace(
       return 'pre';
 
     case NodeTypes.LiquidTag:
-      console.log('LiquidTag', NodeTypes.LiquidTag);
-      console.log('captureWhitespace', options.captureWhitespace);
-      console.log(node.name, CSS_WHITE_SPACE_LIQUID_TAGS[node.name]);
-      console.log('---');
       switch (options.captureWhitespace) {
         case 'strict':
-          return CSS_WHITE_SPACE_LIQUID_TAGS[node.name] || 'pre';
+          return (node.name === 'capture') ? 'pre' : 'normal';
         case 'ignore':
-          return CSS_WHITE_SPACE_LIQUID_TAGS[node.name] || 'normal';
+          return (node.name === 'capture') ? 'normal' : 'pre';
       }
 
     case NodeTypes.LiquidBranch:
--- a/packages/prettier-plugin-liquid/src/printer/preprocess/augment-with-css-properties.ts
+++ b/packages/prettier-plugin-liquid/src/printer/preprocess/augment-with-css-properties.ts
@@ -180,7 +180,10 @@ function getNodeCssStyleWhiteSpace(
       return 'pre';
 
     case NodeTypes.LiquidTag:
-      console.log(options.captureWhitespace);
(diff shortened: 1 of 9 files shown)


packages/prettier-plugin-liquid/src/printer/preprocess/augment-with-css-properties.ts
```diff
@@ -177,7 +180,12 @@ function getNodeCssStyleWhiteSpace(node: AugmentedNode<WithSiblings>): string {
       return 'pre';
 
     case NodeTypes.LiquidTag:
-      return CSS_WHITE_SPACE_LIQUID_TAGS[node.name] || CSS_WHITE_SPACE_DEFAULT;
+      switch (options.captureWhitespace) {
+        case 'strict':
+          return (node.name === 'capture') ? 'pre' : 'normal';
+        case 'ignore':
+          return (node.name === 'capture') ? 'normal' : 'pre';
+      }
```
charlespwd, line 188: Here I think that—since this is an option for `capture` tags only—I'd do a switch on node name first, and then on the option. Also note that we also want `pre` for `raw` tags so I don't think the switch on options.captureWhitespace works well here. We should defer to CSS_WHITE_SPACE_LIQUID_TAGS[node.name] instead. 
```suggestion
      switch (node.name) {
        case NamedTags.capture: {
          switch (options.captureWhitespace) {
            case 'strict':
              return 'pre';
            case 'ignore':
              return 'normal';
            default: {
              throw assertNever(options.captureWhitespace);
            }
          }
        }

        default: {
          return CSS_WHITE_SPACE_LIQUID_TAGS[node.name] || CSS_WHITE_SPACE_DEFAULT;
        }
```

packages/prettier-plugin-liquid/src/test/liquid-tag-capture-whitespace/index.liquid
```diff
@@ -0,0 +1,37 @@
...
+{% # white-space: pre %}
+{% capture var %}
+      <div>{{ x }}</div>
+{% endcapture %}
+
+It should print as is without the white-space pre comment
+printWidth: 1
+{% capture var %}
+      <div>{{ x }}</div>
+{% endcapture %}
```
charlespwd, line 37: There should be a test with the `captureWhitespace` option, you can do that like this:
```
It should bla bla bla
captureWhitespace: 'ignore'
...
```

To verify that the code works as expected.

charlespwd (commented):
Thanks for your contribution! Would need a couple of modifs to merge directly. If you don't have time, someone from our team could pick this up.

packages/prettier-plugin-liquid/src/types.ts
```diff
@@ -23,6 +23,7 @@ export type LiquidParserOptions = ParserOptions<LiquidHtmlNode> & {
   liquidSingleQuote: boolean;
   embeddedSingleQuote: boolean;
   indentSchema: boolean;
+  captureWhitespace: string;
```
charlespwd, line 26: This should be `strict` | `ignore`

packages/prettier-plugin-liquid/src/test/liquid-tag-capture-whitespace/index.liquid
charlespwd: my bad should be a JSON parseable value on the right hand side 
```
captureWhitespace: "ignore"
```

stijns96: Hi @charlespwd,

Do you mean like this?

```liquid
It should never break on capture var name
captureWhitespace: "ignore"
{% capture varName %}{% endcapture %}
{%   capture   varName   %}{% endcapture %}
{%capture varName %}{% endcapture %}
{%- capture
varName
-%}{% endcapture %}

It should be indentation sensitive
<p>{% capture var %}
    # heading

    - list
      * elem
  {% endcapture %}
</p>

It should behave like a pre tag if the capture tag is set white-space pre
captureWhitespace: "ignore"
<!-- white-space: pre -->
{% capture var %}
      <div>{{ x }}</div>
{% endcapture %}

It should behave like a pre tag if the capture tag is set white-space pre (Liquid)
captureWhitespace: "ignore"
{% # white-space: pre %}
{% capture var %}
      <div>{{ x }}</div>
{% endcapture %}

It should print as is without the white-space pre comment
captureWhitespace: "ignore"
{% capture var %}
      <div>{{ x }}</div>
{% endcapture %}

```

Sorry if it's a stupid question. This is actually the first time I'm ever contributing. Also have to get used a little bit on how it works.

charlespwd: > Sorry if it's a stupid question.

It absolutely isn't! I'm impressed you managed to pick it up without more guidance. 

So the way those tests are setup is a bit weird because we want to test before/after. So 
- the index.liquid file is the before
- the fixed.liquid file is the after

Each tests are setup in "groups" which are the following:
- A "It/when/should" statement that describes what the test is supposed to do
- (Optional) list of prettier settings overrides (e.g. `captureWhitespace: "ignore"`, `printWidth: 1`, etc.)
- A block of code that doesn't include three consecutive new line characters. 

So something like this in `index.liquid`
```liquid
It should pretty print the contents of a capture tag with captureWhitespace ignore
printWidth: 1
captureWhitespace: "ignore"
{% capture _ %}
{%for element in col%}
{{element}}
{%endfor%}
{% endcapture %}  

It should print as is when strict
printWidth: 1
captureWhitespace: "strict"
{% capture _ %}
{%for element in col%}
{{element}}
{%endfor%}
{% endcapture %}
```

And this in `fixed.liquid`
```liquid
It should pretty print the contents of a capture tag with captureWhitespace ignore
printWidth: 1
captureWhitespace: "ignore"
{% capture _ %}
  {%for element in col%}
    {{ element }}
  {% endfor %}
{% endcapture %}  

It should print as is when strict
printWidth: 1
captureWhitespace: "strict"
{% capture _ %}
{%for element in col%}
{{element}}
{%endfor%}
{% endcapture %}
```

stijns96: Did my best! It already helped that I also created my own extension for Shopify, haha.

Makes sense! Thanks for the explanation. I'll fix it tomorrow!

stijns96: All test are passing now.

image

PR #359, opened 2024-05-16
Add `VariableName` check

madsenmm:

Adds a `VariableName` check, for a set naming convention to own liking, from the formats below.
Supports the following formats: `camelCase`, `PascalCase`, `snake_case` **(Default)** and `kebab-case`

## What did you learn?

The ins and outs of the theme-tool repo, and how it works.

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/checks/variable-name/index.ts
+++ b/packages/theme-check-common/src/checks/variable-name/index.ts
@@ -16,10 +16,6 @@ const pascalCase = (string: string) => {
   return camelCased.charAt(0).toUpperCase() + camelCased.slice(1);
 };
 
-const stringGetStartEnd = (str: string, sub: string) => {
-  return [str.indexOf(sub), str.indexOf(sub) + sub.length];
-};
-
 const isLiquidTagAssign = (node: LiquidTag): node is LiquidTagAssign => {
   return node.name === 'assign' && typeof node.markup !== 'string';
 };


tmmgrafikr:
I have signed the CLA!

packages/theme-check-common/src/checks/variable-name/index.ts
```diff
@@ -0,0 +1,116 @@
+import {
+  LiquidHtmlNode,
+  LiquidTag,
+  LiquidTagAssign,
+  LiquidTagCapture,
+  NodeTypes,
+} from '@shopify/liquid-html-parser';
+import { LiquidCheckDefinition, SchemaProp, Severity, SourceCodeType } from '../../types';
+
+import { camelCase, kebabCase, snakeCase } from 'lodash';
```
karreiro, line 10: May we use `lodash-es`, I've tried it locally and it works as expected :)

```suggestion
import { camelCase, kebabCase, snakeCase } from 'lodash-es';
```

karreiro (commented):
Thanks a lot for this PR, @tmmgrafikr! This is an excellent idea for a check indeed!

I've noticed that we're not reporting an offense for the first variable when we have two occurrences of the same variable. Also, I've noticed an issue when applying the suggestion in that scenario:

demo-suggestion

I wonder if removing the `wronglyFormattedVariables` `Map` in favor of reporting the offenses on the `LiquidTag` function might be the path to fix that.

Please let me know what you think, and thanks again for this PR!

packages/theme-check-common/src/checks/variable-name/index.ts
```diff
@@ -0,0 +1,118 @@
...
+
+const pascalCase = (string: string) => {
+  const camelCased = camelCase(string);
+  return camelCased.charAt(0).toUpperCase() + camelCased.slice(1);
+};
+
+const stringGetStartEnd = (str: string, sub: string) => {
+  return [str.indexOf(sub), str.indexOf(sub) + sub.length];
+};
+
```
karreiro, line 22: I believe we're no longer using this method, so we may remove it :)

```suggestion
```

karreiro (approved):
Thank you for the updates, @tmmgrafikr! I noticed just one last minor detail, and with that applied, I believe we're ready to merge 🚀

tmmgrafikr:
> Documentation for this [check](https://github.com/Shopify/theme-check/tree/main/docs/checks) is missing!

I've added an issue regarding this: https://github.com/Shopify/theme-tools/issues/395
I don't know how the docs site works, thought it was automated based on the contents of this repo, but guess not 😅

tmmgrafikr:
> @tmmgrafikr Why numbers are not accepted in `variable_name`?

That one is a bug, will look into it in near future.
Else you're more than welcome to add a PR fix on this, it simply uses `lodash` different casing formatters, as validators

nikitaourazbaev:
@thagxt @tmmgrafikr I’ve opened an issue about numbers, in my opinion `first_3d_model` should be accepted as correct, as it follows a long-standing naming convention in Shopify themes:

https://github.com/Shopify/theme-tools/issues/397

I looked into customizing it however as the current implementation relies on lodash, we would need to change the RegEx that matches against unicode words, which is beyond my RegEx skills personally:

https://github.com/lodash/lodash/blob/main/src/.internal/unicodeWords.ts

Interestingly, the `words` function switches to using `unicodeWords` once it encounters a number in front of a word or a number following a word, otherwise it uses the simpler `asciiWords` function. If I edit `words` to force `asciiWords`, the words array returned is `['first', '3d', 'model' ]` and the snakeCase result does become `first_3d_model`. However, the return value of `unicodeWords` is `['first', '3', 'd', 'model']`, which results in `first_3_d_model`.

So there’s a mismatch in results there between the two lodash functions. `words` can also take a custom RegEx pattern as the second argument, I wonder if the way around this is to write custom casing functions that pass a simpler pattern to `words`? 

I’m not sure as to the extent of unicode support in Liquid variable names is / should be.

PR #363, opened 2024-06-07
Adds new check for uniqueness of ids for blocks

MateusJabour:

Add a new theme check called `UniqueStaticBlockId`, which ensures the uniqueness of the ids of static blocks defined in a block or section.

Read more here: https://docs.google.com/document/d/1d1EpwfhbgSubLjEqMePZ30ZWJovuCt2mJQslgWSwex0/edit#heading=h.p7g70mn8jka8

## Before you deploy

### DIFF
--- a/.changeset/silly-sheep-happen.md
+++ b/.changeset/silly-sheep-happen.md
@@ -1,5 +1,5 @@
 ---
-'@shopify/theme-check-common': major
+'@shopify/theme-check-common': minor
 ---
 
 Add `UniqueStaticBlockId` check which ensures the uniqueness of ids for static blocks defined in the same section/block file


packages/theme-check-common/src/checks/unique-static-block-id/index.ts
```diff
@@ -0,0 +1,64 @@
...
+      recommended: true,
+    },
+    type: SourceCodeType.LiquidHtml,
+    severity: Severity.ERROR,
+    schema: {},
+    targets: [],
+  },
+
+  create(context) {
+    const usedIds: string[] = [];
```
albchu, line 20: A `Set` here should be more ideal for your usecase. Sets are faster than arrays for searching.

packages/theme-check-common/src/checks/unique-static-block-id/index.ts
```diff
@@ -0,0 +1,64 @@
...
+
+        const idRegex = /id: ["'](\S+)["']/;
+
+        const idValueMatch = idRegex.exec(node.markup);
+
+        if (idValueMatch == null) {
+          return;
+        }
+
+        const filteredIdValue = idValueMatch[1];
```
albchu, line 41: nit: At this point in the code, you could destructure the two match elements out of `idValueMatch` to be a bit more concise when you use each value. 
Something like:
```
const [entireIdTerm, idValue] = idValueMatch
```

packages/theme-check-common/src/checks/unique-static-block-id/index.ts
```diff
@@ -0,0 +1,64 @@
...
+          return;
+        }
+
+        const [blockType] = node.markup.split(',');
+
+        if (blockType.replace(/["']/g, '') !== 'block') {
+          return;
+        }
+
+        const idRegex = /id: ["'](\S+)["']/;
```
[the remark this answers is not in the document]
albchu: The idRegex is redefined each time this node type is encountered for any opened liquid file. It would be more efficient to move the definition out of the function.

albchu (commented):
Its working nicely on my end. Some comments for consideration.

albchu (approved):
Looks great, thanks!

packages/theme-check-node/configs/all.yml
```diff
@@ -85,6 +85,9 @@ UnclosedHTMLElement:
 UndefinedObject:
   enabled: true
   severity: 1
+UniqueStaticBlockId:
+  enabled: true
+  severity: 0
```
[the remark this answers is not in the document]
MateusJabour: There are 3 types of `Severity`, and the comment on the type answers your question:
```
/** The severity determines the icon and color of diagnostics */
export enum Severity {
  ERROR = 0,
  WARNING = 1,
  INFO = 2,
}
```

bahar-p (approved):
LGTM! 🚀🚀

.changeset/silly-sheep-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/theme-check-common': major
```
karreiro, line 2: Personally, I believe we could update this to be `minor`:

```suggestion
'@shopify/theme-check-common': minor
```

Otherwise, we will need to bump Theme Check to Theme Check 3, and I don't think we want to do that. What do you think @MateusJabour @albchu?

karreiro (commented):
Thank you, @MateusJabour! I just left one minor comment about the changesets (as they impact the release) :)

PR #418, opened 2024-07-15
Space-after-classlist

geddski:

This theme check ensures that a space is used after the `class_list` filter. Otherwise the content after the filter will be appended to the class_list's generated class names.

For example it catches this problematic case:

```html
<div class="{{ block.settings.spacing | class_list }}other-class">
    content
</div>
```

## Before you deploy

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -104,7 +104,9 @@ If ever you want to see how the VS Code extension or playground would behave bef
     theme-docs download
     ```
 
-6. Proceed to debug the VS Code extension or playground as usual.
+6. Note that you may have to delete the cached data directory (`~/Library/Caches/theme-liquid-docs-nodejs`) in order to see your changes.
+
+7. Proceed to debug the VS Code extension or playground as usual.
 
 ## Submitting a Pull Request
 
--- a/packages/theme-check-common/package.json
+++ b/packages/theme-check-common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/theme-check-common",
-  "version": "2.8.0",
+  "version": "2.9.0",
   "license": "MIT",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
--- a/packages/theme-check-common/src/checks/space-after-class_list/index.ts
+++ b/packages/theme-check-common/src/checks/space-after-class_list/index.ts
@@ -1,0 +1,75 @@
+import {
+  NodeTypes,
+} from '@shopify/liquid-html-parser';
+import { LiquidCheckDefinition, Severity, SourceCodeType } from '../../types';
+
+export const SpaceAfterClassList: LiquidCheckDefinition = {
+  meta: {
+    code: 'SpaceAfterClassList',
+    aliases: ['SpaceAfterClassList'],
+    name: 'Space After Class List',
+    docs: {
+      description: 'Warns you when there is no space after using the class_list filter',
+      recommended: true,
+      url: 'https://shopify.dev/docs/themes/tools/theme-check/checks/space-after-class_list',
+    },
+    type: SourceCodeType.LiquidHtml,
+    severity: Severity.ERROR,
+    schema: {},
+    targets: [],
+  },
+
+  create(context) {
+    return {
+
+      async LiquidFilter(node, ancestors) {
+        if (node.name !== 'class_list') {
+          return;
+        }
+
+        const classAttribute = ancestors.find((ancestor) => ancestor.type === NodeTypes.AttrDoubleQuoted || ancestor.type === NodeTypes.AttrSingleQuoted);
+
+        if (!classAttribute) {
+          return;
+        }
+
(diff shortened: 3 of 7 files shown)


packages/theme-check-common/src/checks/space-after-class_list/index.ts
```diff
@@ -0,0 +1,84 @@
...
+    code: 'SpaceAfterClassList',
+    aliases: ['SpaceAfterClassList'],
+    name: 'Space After Class List',
+    docs: {
+      description: 'Warns you when there is no space after using the class_list filter',
+      recommended: true,
+      url: 'https://shopify.dev/docs/themes/tools/theme-check/checks/space-after-class_list',
+    },
+    type: SourceCodeType.LiquidHtml,
+    severity: Severity.ERROR,
```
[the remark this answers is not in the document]
geddski: hmm it feels like error to me since it breaks the style settings, but I don't feel strongly either way.

packages/theme-check-common/src/checks/space-after-class_list/index.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export const SpaceAfterClassList: LiquidCheckDefinition = {
+  meta: {
+    code: 'SpaceAfterClassList',
+    aliases: ['SpaceAfterClassList'],
+    name: 'Space After Class List',
+    docs: {
+      description: 'Warns you when there is no space after using the class_list filter',
+      recommended: true,
+      url: 'https://shopify.dev/docs/themes/tools/theme-check/checks/space-after-class_list',
```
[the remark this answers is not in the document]
geddski: knew I was forgetting something! I'll see how to add that one sec

packages/theme-check-common/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/theme-check-common",
-  "version": "2.8.0",
+  "version": "2.9.0",
```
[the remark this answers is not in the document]
geddski: fixed

docs/contributing.md
```diff
@@ -104,7 +104,9 @@ If ever you want to see how the VS Code extension or playground would behave bef
     theme-docs download
     ```
 
-6. Proceed to debug the VS Code extension or playground as usual.
+6. Note that you may have to delete the cached data directory (`~/Library/Caches/theme-liquid-docs-nodejs`) in order to see your changes.
```
albchu, line 107: Since this is general contribution docs, we should probably be a bit more specific about what this instruction is alluding to.
```suggestion
6. Note when working on changes to theme-liquid-docs: You will have to delete the cached data directory (`~/Library/Caches/theme-liquid-docs-nodejs`) in order to see changes made to `packages/theme-check-docs-updater/data`.
```

packages/theme-check-common/src/checks/space-around-class_list/index.ts
```diff
@@ -0,0 +1,124 @@
...
+
+          context.report({
+            message: `Missing a space after using the class_list filter: '${match[0]}'`,
+            startIndex: errorPosition,
+            endIndex: errorPosition + 1,
+            suggest: [
+              {
+                message: 'Add a space after the class_list filter',
+                fix(corrector) {
+                  corrector.insert(errorPosition, ' ');
```
albchu, line 67: Nice!

albchu (approved):
LGTM. One minor doc change suggestion but then youre good to go! Thanks Dave :)

packages/theme-check-common/src/checks/space-around-class_list/index.ts
```diff
@@ -0,0 +1,124 @@
...
+        if (!classAttribute) {
+          return;
+        }
+
+        const classAttributeContent =
+          classAttribute.source.slice(classAttribute.position.start, classAttribute.position.end) ||
+          '';
+
+        // check for missing space after class_list
+        const afterRegex = /([a-zA-Z0-9._-]+)\s*\|\s*class_list\s*}}([a-zA-Z0-9._-]+)/gm;
```
karreiro, line 42: I believe we have valid cases, such as the following one where we want to raise an offense:
```liquid
<div class="{{ block.settings.spacing | class_list }}{{ shop.name }}other-class">content</div>
```

It's a bit changeling to cover that case relying on the regex tho. Instead, I believe we may navigate in the AST and can check if the sibling node starts with a space:

Screenshot 2024-07-22 at 08 48 34

What do you think?

karreiro (commented):
Thank you for this PR, @geddski! I've left only one minor suggestion about how we capture the absence of the space character. Please let me know what you think about that :)

packages/theme-check-common/src/checks/space-around-class_list/index.ts
geddski: ah, another interesting case. I'm glad y'all know your liquid! Would it be correct then to say the filter's next sibling (if any) must be a TextNode that starts with whitespace?

Issue #397, opened 2024-06-28
VariableName check too strict around numbers


nikitaourazbaev:
**Describe the bug**
`first_3d_model` results in a `VariableName` check warning

**Source**

```liquid
{% assign first_3d_model = product.media | where: 'media_type', 'model' | first %}
```

**Expected behaviour**
`VariableName` check should pass

**Actual behaviour**
`VariableName` check doesn’t pass, with the warning message “The variable 'first_3d_model' uses wrong naming format”

**Debugging information**
 - macOS
 - 14.5
 - Shopify Liquid VS Code Extension v2.4.1

**Additional context**
I understand that `first_3_d_model` would be the correct casing according to the library used for the check, but in my opinion this goes against a long history of naming variables in Liquid, including in Shopify’s own themes. Adjusting the check so it’s more lenient about numbers would be preferred, especially as this check is on by default.

madsenmm:
@nikitaourazbaev Added a fix in this PR:
https://github.com/Shopify/theme-tools/pull/462

nikitaourazbaev:
> @nikitaourazbaev now merged: [#462 (review)](https://github.com/Shopify/theme-tools/pull/462#pullrequestreview-2208719136) Not sure how long before the update is available in the CLI.

@madsenmm Amazing, thank you so much!

PR #462, opened 2024-07-26
Fix number formatting

madsenmm:

Changes the formatter handling of `VariableName` check, to default to more regular standards, based on general theme developers naming of variables.

Related to this issue: https://github.com/Shopify/theme-tools/issues/397

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/checks/variable-name/index.ts
+++ b/packages/theme-check-common/src/checks/variable-name/index.ts
@@ -62,10 +62,9 @@ export const VariableName: LiquidCheckDefinition<typeof schema> = {
         };
       }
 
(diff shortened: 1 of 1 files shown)


madsenmm:
> 👋🏻 Hi @madsenmm, thank you for your contribution!
> 
> The approach you've taken in this PR looks great, just a couple of small comments. Would you also be able to include a [new test scenario](https://github.com/Shopify/theme-tools/blob/89cb122b1f52fae135d1da862fb27dcf8e0e7f8c/packages/theme-check-common/src/checks/variable-name/index.spec.ts) to cover this behaviour?
> 
> Thanks!

Updated ☺️

PR #468, opened 2024-08-06
Update VariableName docs url

jeffgreendesign:

The previous link to the VariableName docs in Shopify theme check is going to a 404 page in /docs/themes/. I've updated it to the correct link in /docs/storefronts/.

There are many other /docs/themes/ urls that should be updated to their new location on the /docs/storefronts/.


jeffgreendesign:
I have signed the CLA!

karreiro (approved):
Thank you, @jeffgreendesign!

PR #478, opened 2024-08-30
We should have HTML attribute snippet completion of the =""

### LINKED ISSUE #345: ="" not automatically added with HTML attributes

stijns96:
HTML attribute autocompletion (`=""`)

- `insertText` to add the right SnippetString
- `insertTextFormat` to determine if it's a snippet or plain text

closes #345

## What did you learn?

Struggled quite some time to get the tab stops working. Went back to the original issue where @charlespwd already mentioned something about the `insertTextFormat`. Forgot about it, but then (little late) it came up again.

I was already familiar with snippets and extensions, but the Shopify theme-tools repo can be quite overwhelming.

If someone has good knowledge about testing/debugging extension (F5), please let me know how to make change (e.g. add a console) without the need to fully restart everything, let me know :)

### DIFF
--- a/packages/theme-language-server-common/src/completions/providers/HtmlAttributeCompletionProvider.ts
+++ b/packages/theme-language-server-common/src/completions/providers/HtmlAttributeCompletionProvider.ts
@@ -45,9 +45,13 @@ function getOptions(partial: string, parentNodeName: string): Attribute[] {
 }
 
 function toCompletionItem(tag: Attribute): CompletionItem {
+  const plainText = !tag.valueSet || tag.valueSet !== 'v';
+
   return {
     label: tag.name,
     kind: CompletionItemKind.Value,
+    insertText: plainText ? `${tag.name}="$1"` : tag.name,
+    insertTextFormat: plainText ? 2 : 1,
     documentation: {
       kind: 'markdown',
       value: renderHtmlEntry(tag),


stijns96:
Any news on this?

charlespwd:
Hey @stijns96. Was on medical leave for a while and now I'm back. Problem with the approach used in the PR is that it doesn't account for "existing code." If you're re-accepting the completion in `<div loading|="eager">`, you'd end up with `<div loading="eager"="eager">` which we don't want. 

I have a PR coming up for Liquid tag completion that uses textEdits to optionally replace code and I think we'd want to have similar guards here.

charlespwd:
Yeah I don't think we can merge this as is. We'd unit tests that cover the case I just mentioned. 

I'll probably have the Liquid end tag completion with snippets ready next week, feel free to spy on it :) (But I gotta warn ya, it's not the easiest code to follow 😅 ... lots of edge cases...)

charlespwd:
Looks like @aswamy beat you to it :) Closing in favour of #520

stijns96:
Fine by me! Haha. As long as it will work nicely in the end :)

PR #480, opened 2024-09-05
Fix broken parsing of liquid attribute names

### LINKED ISSUE #335: LiquidHTMLSyntaxError false positive with dynamic attributes
**Describe the bug**
It seems like liquid with dynamic attribute name is considered an error

**Source**

```liquid

<img {{ src_attr }}="{{ image | image_url: width: fallback_width }}" />


```

Screenshot 2024-03-28 at 5 33 02 PM

If i change the attribute name to a fixed string, it doesn't error (`img src="..."`)

**Expected behaviour**
Be able to dynamically set html attribute names in liquid

**Actual behaviour**
Theme-check flags it as an error state

**Debugging information**
  -Current Shopify CLI version: 3.58.0

lukeh-shopify:

Resolves #335 

This was occurring because the liquid variable was being parsed as a liquid node, rather than a double quoted attribute. 

## What did you learn?

@charlespwd Gave a great explanation on how ohmjs is used to build the CST prior to parsing things into an AST. He also showed me how to use https://ohmjs.org/editor/ to debug issues with the CST parsing. Thank you!

image

## Before you deploy

### DIFF
--- a/.changeset/wicked-balloons-reply.md
+++ b/.changeset/wicked-balloons-reply.md
@@ -1,0 +1,5 @@
+---
+'@shopify/liquid-html-parser': patch
+---
+
+Fix broken parsing of liquid attribute names
--- a/packages/liquid-html-parser/grammar/liquid-html.ohm
+++ b/packages/liquid-html-parser/grammar/liquid-html.ohm
@@ -430,7 +430,7 @@ LiquidHTML <: Liquid {
   AttrList = Attr*
 
   Attr =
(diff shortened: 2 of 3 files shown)


mgmanzella (commented):
just for my own learning -- reordering parsing for attributes worked because we need to parse attribute quotes first and then liquid nodes within attributes if we find them? is that right?

graygilmore (approved):
🎉

PR #486, opened 2024-09-16
Add LinkedEditingRanges support for HTML tag names

charlespwd:

- Add [linked editing](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_linkedEditingRange) support for HTML element names

https://github.com/user-attachments/assets/0b53f4c2-6f09-405b-9047-d157cd4620fd

Just like for [HTML in VS Code](https://code.visualstudio.com/updates/v1_52#_html), this feature is enabled by the `editor.linkedEditing` VS Code setting:

```json
"editor.linkedEditing": true
```

When enabled, this will make it so you can rename open/close pairs as you are typing.

## How it works

VS Code and other language clients can send `client->server` [Linked Editing Range][ler] requests
to language servers to figure out what pieces of content should change together.

Our job in the language server is to provide those ranges when they are requested for.

- Add DocumentHighlight support
- Add Rename and PrepareRename support
- Add codemirror language client support

## Before you deploy

[ler]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_linkedEditingRange

### DIFF
--- a/packages/theme-language-server-common/src/linkedEditingRanges/providers/HtmlTagNameLinkedRangesProvider.spec.ts
+++ b/packages/theme-language-server-common/src/linkedEditingRanges/providers/HtmlTagNameLinkedRangesProvider.spec.ts
@@ -1,0 +1,97 @@
+import { describe, beforeEach, it, expect, assert } from 'vitest';
+import { LinkedEditingRangesProvider } from '../LinkedEditingRangesProvider';
+import { DocumentManager } from '../../documents';
+import { LinkedEditingRangeParams } from 'vscode-languageserver';
+import { Position } from 'vscode-languageserver-protocol';
+import { htmlElementNameWordPattern } from '../wordPattern';
+
+describe('Module: HtmlTagNameLinkedRangesProvider', () => {
+  let documentManager: DocumentManager;
+  let provider: LinkedEditingRangesProvider;
+
+  beforeEach(() => {
+    documentManager = new DocumentManager();
+    provider = new LinkedEditingRangesProvider(documentManager);
+  });
+
+  it('should return null for non-existent documents', async () => {
+    const params: LinkedEditingRangeParams = {
+      textDocument: { uri: 'file:///path/to/non-existent-document.liquid' },
+      position: Position.create(0, 0),
+    };
+
+    const result = await provider.linkedEditingRanges(params);
+    expect(result).toBeNull();
+  });
+
+  it('should return null for non-HTML documents', async () => {
+    const params: LinkedEditingRangeParams = {
+      textDocument: { uri: 'file:///path/to/document.liquid' },
+      position: Position.create(0, 0),
+    };
+
+    documentManager.open(params.textDocument.uri, 'Sample text content', 1);
+
+    const result = await provider.linkedEditingRanges(params);
+    expect(result).toBeNull();
+  });
+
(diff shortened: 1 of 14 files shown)


packages/theme-language-server-common/src/linkedEditingRanges/providers/HtmlTagNameLinkedRangesProvider.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (!node || !ancestors) return null;
+
+    const textDocument = this.documentManager.get(params.textDocument.uri)?.textDocument;
+    if (!textDocument) return null;
+
+    const ranges = getHtmlElementNameRanges(node, ancestors, params, textDocument);
+    if (!ranges) return null;
+    return {
+      ranges,
+      wordPattern: '[a-zA-Z0-9-_]+',
```
aswamy, line 24: The `_` char is acting funky when i tophat it.

Once you add a `_` to the tag name, it just doesn't duplicate the rest.

E.g.

```
<!-- works -->
<foo></foo>
<foo-bar></foo-bar>

<!-- halts after typing _ -->
<foo_bar></foo_>
```

`getHtmlElementNameRanges` doesn't return `ranges` when there's an underscore in it.

charlespwd: Removed `_` from the word pattern :+1: Nice catch :)

packages/theme-language-server-common/src/linkedEditingRanges/providers/HtmlTagNameLinkedRangesProvider.spec.ts
```diff
@@ -0,0 +1,97 @@
+import { describe, beforeEach, it, expect, assert } from 'vitest';
+import { LinkedEditingRangesProvider } from '../LinkedEditingRangesProvider';
+import { DocumentManager } from '../../documents';
+import { LinkedEditingRangeParams } from 'vscode-languageserver';
+import { Position } from 'vscode-languageserver-protocol';
+import { htmlElementNameWordPattern } from '../wordPattern';
+
+describe('Module: LinkedEditingRangesProvider', () => {
```
aswamy, line 8:
```suggestion
describe('Module: HtmlTagNameLinkedRangesProvider', () => {
```

packages/theme-language-server-common/src/linkedEditingRanges/providers/HtmlTagNameLinkedRangesProvider.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+    assert(result);
+    assert(result.ranges[0]);
+    assert(result.ranges[1]);
+
+    const startTagName = document.getText(result.ranges[0]);
+    const endTagName = document.getText(result.ranges[1]);
+
+    expect(startTagName).toBe('div');
+    expect(endTagName).toBe('div');
+    expect(result.wordPattern).toBe('[a-zA-Z0-9-_]+');
```
aswamy, line 83:
```suggestion
    expect(result.wordPattern).toBe(htmlElementNameWordPattern);
```
Think you forgot one

aswamy (approved):
Looks great! 🎩 went well too. Thanks for the pair 🍐.

PR #487, opened 2024-09-16
Add DocumentHighlight support for HTML tags and Liquid blocks

charlespwd:

- Adds support for paired highlighting of HTML element names and Liquid blocks

https://github.com/user-attachments/assets/20b54ccf-a5e9-40ee-83eb-9cab3204e1e6

When you hover over a HTML tag open, the close tag is highlighted and vice-versa.

```html
<div> <!-- this div gets highlighted -->
   <div></div>
</div> <!-- with this one-->
```

When you hover over a Liquid block open, the close block is highlighted (and branches if any).

```liquid
{% # this if, elsif, else, endif all get highlighted together %}
{% if cond %}

{% elsif cond %}

{% else %}

{% endif %}
```

## How it works

In the language server, we start listening for [Document Highlight `client->server` requests][dh].

When we receive these, we may respond with an array of [`DocumentHighlight`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#documentHighlight) representing the stuff that should be highlighted together:

- HTML tag open/close pairs
- Liquid block open/branch/close tuples (if/elsif/else/endif, form/endform, etc)

- Merge https://github.com/Shopify/theme-tools/pull/486 (Linked Editing)
- Add Rename support
- Add codemirror language client support for this feature

## Before you deploy

[dh]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentHighlight

### DIFF
--- a/.changeset/thick-ravens-peel.md
+++ b/.changeset/thick-ravens-peel.md
@@ -1,0 +1,27 @@
+---
+'@shopify/theme-language-server-common': minor
+'theme-check-vscode': minor
+---
+
+Add tupled highlighting of HTML element names and Liquid blocks
+
+When you hover over a HTML tag open, the close tag is highlighted and vice-versa.
+
+```html
+<div> <!-- this div gets highlighted -->
+   <div></div>
+</div> <!-- with this one-->
+```
+
+When you hover over a Liquid block open, the close block is highlighted (and branches if any).
+
+```liquid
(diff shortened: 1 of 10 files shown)


EvilGenius13 (approved):
Thank you for the added context during pairing

Issue #189, opened 2023-03-28
liquid format error


RhinoLu:
**Describe the bug**
The minus sign should stick with the percentage sign.

**Unformatted source**
```liquid
{%-if customer.name==some_name-%}{%-endif-%}
```

**Expected output**
```liquid
{%- if customer.name == some_name -%}{%- endif -%}
```

**Actual output**
```liquid
{%- if customer.name == some_name- %}{%- endif -%}
```

**Debugging information**
 - Mac
 - Version 1.0.6

**Additional context**
<img width="1291" alt="Prettier_Liquid_Plugin_-_Playground" src="https://user-images.githubusercontent.com/349025/228113931-67dcb5e8-05c6-44f6-be15-7e5c569ddd33.png">

PR #495, opened 2024-09-17
[Fix] - Improve parsing of whitespace stripping characters in Liquid tags and variables

jamesmengo:
Fixes https://github.com/Shopify/theme-tools/issues/189

## What did you learn?

Learned that this is intentionally valid: `{% echo 'a' -%}`.

This will strip whitespace on the RHS of the output. This can be useful if used inside of whitespace sensitive HTML such as a `<span>{% echo 'a' -%}</span>`

### Demo
Input: `{%-echo someVariableName-%}`
### Before
image

### After
image

## Before you deploy

### DIFF
--- a/.changeset/perfect-wasps-crash.md
+++ b/.changeset/perfect-wasps-crash.md
@@ -1,0 +1,5 @@
+---
+'@shopify/liquid-html-parser': patch
+---
+
+Fix bug where whitespace stripping character was assigned to variable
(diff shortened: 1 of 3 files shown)


graygilmore (approved):
🎩 👍🏻

charlespwd (approved):
Yeehaw! 

Like gray said it's probably in the wrong section of the tests. But that's the fix indeed :)

Issue #260, opened 2023-12-18
Bug - Snippet autocomplete suggestions cutting off periods


sdn90:
**Describe the bug**

![Screenshot 2023-12-18 at 3 23 37 PM](https://github.com/Shopify/theme-tools/assets/2491713/574ccfd2-7087-4067-8945-6f473b360b88)

I have snippets named

```
block-product.installation.liquid
block-product.specs-table.liquid
block-product.ugc.liquid
```

When triggering the Intellisense box with:
```liquid
{% render 'block-' %}
```

The top results are all `block-product`.

Selecting the result returns `block-product` as well

**Expected behaviour**
Autocomplete suggestions should return the full snippet filename

**Debugging information**
VS Code Extension v1.15.0
`shopifyLiquid.themeCheckNextDevPreview` Enabled

PR #496, opened 2024-09-17
Fix bug preventing snippets with periods to be autocompleted

graygilmore:

- Fixes https://github.com/Shopify/theme-tools/issues/260

Previously this was matching against the first period it found in the snippet file and removing everything up until the end of the file name.

This meant that if you had a file like `foo.bar.baz.liquid` it would only offer `foo` as the suggestion. This change allows for any number of periods to be added to the file name and only removes the final `.liquid`.

## Before you deploy


packages/theme-language-server-common/src/server/startServer.ts
```diff
@@ -123,7 +123,7 @@ export function startServer(
         x
           .replace(/\\/g, '/')
           .replace(/^snippets\//, '')
-          .replace(/\..*$/, ''),
+          .replace(/\.liquid$/, ''),
```
graygilmore, line 126: This isn't under any tests right now. Because it's inside `startServer` it's a little hard to pull out and test like a unit. Going to spend a bit of time exploring pulling this logic out but didn't want to block opening a PR to fix the bug.

charlespwd: Oops 🙈 !

I mean, I'm not 100% sure a test would bring value / protect against refactors here. Plus I feel like extracting this combined dependency injection would just make the code harder to understand. Moving it out of scope would require you to do something like this and blerh 😅 

```ts
// You'd be forced to do something like this
const makeGetSnippetNamesForURI : ... = (filesForURI: Dependencies['fileForURI']) => async(uri: string) => {
  // ...
}

// I feel like that's where the value gained is smaller than the readability cost. 
// IMO it makes the thing harder to use / read at first glance.
// Am I crazy? WDYT?
```

Issue #165, opened 2023-10-11
Add `syntax` blurb in completion/hover windows. 


charlespwd:
**Is your feature request related to a problem? Please describe.**
Most of the DocsetEntry have a `syntax` tag. 

**Describe the solution you'd like**
Would be cool to print that in a helpful way as part of the completion items and hover items.

![Image](https://github.com/user-attachments/assets/c9d5edac-7885-4ef4-8cfe-618582f3e891)

Right now, our Completion and Hover items for Liquid tags and filters don't use that information. They only show the return value of the tag. 

![Image](https://github.com/user-attachments/assets/554d70b0-6c3c-4084-a2e2-7c2558f8f99b)

Code to look at:

- https://github.com/Shopify/theme-tools/blob/main/packages/theme-language-server-common/src/completions/providers/LiquidTagsCompletionProvider.ts
- https://github.com/Shopify/theme-tools/blob/main/packages/theme-language-server-common/src/completions/providers/FilterCompletionProvider.ts
- https://github.com/Shopify/theme-tools/blob/main/packages/theme-language-server-common/src/hover/providers/LiquidTagHoverProvider.ts
- https://github.com/Shopify/theme-tools/blob/main/packages/theme-language-server-common/src/hover/providers/LiquidFilterHoverProvider.ts
- https://github.com/Shopify/theme-tools/blob/main/packages/theme-language-server-common/src/completions/providers/common/CompletionItemProperties.ts#L41-L56
- https://github.com/Shopify/theme-tools/blob/main/packages/theme-language-server-common/src/docset/MarkdownRenderer.ts#L10

PR #508, opened 2024-09-27
Add an entry's syntax declaration to hover info

graygilmore:

- Closes https://github.com/Shopify/theme-tools/issues/165

We currently have this information but aren't displaying it in any way. If available, this commit adds the syntax under the name of the tag/filter when hovering so that developers can more easily see how it can be used.

I've attached some screenshots showing the before/after (ignore the slight differences in styling as I'm taking the `before` from Online Store's code editor and the `after` from my local VSCode):

#### Tags

| Before | After |
|--------|--------|
| image | image | 

#### Filters

| Before | After |
|--------|--------|
| image | image | 

## Before you deploy

### DIFF
--- a/packages/theme-language-server-common/src/docset/MarkdownRenderer.ts
+++ b/packages/theme-language-server-common/src/docset/MarkdownRenderer.ts
@@ -1,14 +1,14 @@
-import { DocsetEntry, FilterEntry, ObjectEntry } from '@shopify/theme-check-common';
+import { DocsetEntry, FilterEntry, ObjectEntry, TagEntry } from '@shopify/theme-check-common';
 import { ArrayType, PseudoType, docsetEntryReturnType, isArrayType } from '../TypeSystem';
 import { Attribute, Tag, Value } from './HtmlDocset';
 
 const HORIZONTAL_SEPARATOR = '\n\n---\n\n';
 
-type HtmlEntry = Tag | Attribute | Value;
+export type HtmlEntry = Tag | Attribute | Value;
 export type DocsetEntryType = 'filter' | 'tag' | 'object';
 
 export function render(
-  entry: DocsetEntry,
+  entry: DocsetEntry | FilterEntry | TagEntry,
   returnType?: PseudoType | ArrayType,
   docsetEntryType?: DocsetEntryType,
 ) {
@@ -51,6 +51,7 @@ function docsetEntryBody(
   docsetEntryType?: DocsetEntryType,
 ) {
   return [
+    syntax(entry),
     entry.deprecation_reason,
     entry.summary,
     entry.description,
@@ -65,6 +66,22 @@ function htmlEntryBody(entry: HtmlEntry, parentEntry?: HtmlEntry) {
   return [description(entry), references(entry), references(parentEntry)]
     .filter(Boolean)
     .join(HORIZONTAL_SEPARATOR);
+}
+
+function syntax(entry: DocsetEntry | FilterEntry | TagEntry) {
+  if (!('syntax' in entry) || !entry.syntax) {
+    return undefined;
+  }
+
+  // TagEntry entries already have liquid tags as a part of the syntax
+  // explanation so we can return them directly.
+  if (entry.syntax.startsWith('{%')) {
+    return `\`\`\`liquid\n${entry.syntax}\n\`\`\``;
+  }
+
+  // Wrap the syntax in liquid tags to ensure we get proper syntax highlighting
+  // if it's available.
+  return `\`\`\`liquid\n{{ ${entry.syntax} }}\n\`\`\``;
 }
 
 function description(entry: HtmlEntry) {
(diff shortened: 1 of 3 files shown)


packages/theme-language-server-common/src/docset/MarkdownRenderer.ts
```diff
@@ -67,6 +68,22 @@ function htmlEntryBody(entry: HtmlEntry, parentEntry?: HtmlEntry) {
     .join(HORIZONTAL_SEPARATOR);
 }
 
+function syntax(entry: DocsetEntry | FilterEntry | TagEntry) {
+  if (!('syntax' in entry) || !entry.syntax) {
```
graygilmore, line 72: This line could be a little confusing since it looks like I'm doing the same check twice but it's a little more nuanced than that! The first check of `!('syntax' in entry)` gives us type narrowing. Only `FilterEntry` and `TagEntry` have the `syntax` attribute but this could be a `DocsetEntry`. If only checked for `entry.syntax` TypeScript would complain because that attribute doesn't exist on `DocsetEntry`. Once we've narrowed the type *then* we can check `!entry.syntax` (the attribute could be there but `undefined`).

karreiro (commented):
Thank you for this PR, @graygilmore!

When the language server starts, the client may specify the [capabilities it supports](https://github.com/Shopify/theme-tools/blob/9badd60640cf5be8e24ec1fa084f66785b4f6536/packages/theme-language-server-common/src/server/startServer.ts#L191). The [capabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#clientCapabilities) are a bit of a large data structure, but if we navigate from there, we will reach the [hover capabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#hoverClientCapabilities) where clients may specify if they support `markdown` or `plaintext`.

Thus, what do you think about using that here? We could apply the markdown syntax only if the client supports it. For clients that do not support it, we'd still return a plain text version.

I personally don't feel strongly about this because the markdown syntax is quite clean and simple to understand, so it's still relatively readable for plaintext clients. What do you think?

Thanks again for this PR!

packages/theme-language-server-common/src/docset/MarkdownRenderer.ts
```diff
@@ -1,14 +1,14 @@
...
 
 const HORIZONTAL_SEPARATOR = '\n\n---\n\n';
 
-type HtmlEntry = Tag | Attribute | Value;
+export type HtmlEntry = Tag | Attribute | Value;
 export type DocsetEntryType = 'filter' | 'tag' | 'object';
 
 export function render(
-  entry: DocsetEntry,
+  entry: DocsetEntry | FilterEntry | TagEntry,
```
charlespwd, line 11: I believe both FilterEntry and TagEntry extends DocsetEntry, don't they?

packages/theme-language-server-common/src/docset/MarkdownRenderer.ts
charlespwd: Ah that's why you need FilterEntry | TagEntry. Because the props are _extra_.

packages/theme-language-server-common/src/docset/MarkdownRenderer.ts
```diff
@@ -67,6 +68,22 @@ function htmlEntryBody(entry: HtmlEntry, parentEntry?: HtmlEntry) {
...
+  // TagEntry entries already have liquid tags as a part of the syntax
+  // explanation so we can return them directly.
+  if (entry.syntax.startsWith('{%')) {
+    return `\`\`\`liquid\n${entry.syntax}\n\`\`\``;
+  }
+
+  // Wrap the syntax in liquid tags to ensure we get proper syntax highlighting
+  // if it's available.
+  return `\`\`\`liquid\n{{ ${entry.syntax} }}\n\`\`\``;
+}
```
charlespwd, line 85: Given that we have fallback tag.json and filter.json files in packages/theme-liquid-docs-updater/data. It feels like we could benefit from having a test in there that validates the assumption.

charlespwd (approved):
Nicely done. Minor nits to maybe catch assumption errors.

charlespwd:
@karreiro 

> I personally don't feel strong about this because the markdown syntax is quite clean and simple to understand, so it's still relatively readable for plaintext clients. What do you think?

I've initially thought the same. We could start having a bunch of if statements for every capability out there, but since the admin, vscode and neovim's language clients all support markdown, I don't care so much about the capability problem. We don't have to make our lives complicated.

packages/theme-language-server-common/src/docset/MarkdownRenderer.ts
graygilmore: We chatted about this IRL. Going to skip this for now but it's good to keep in mind anytime we make assumptions about the data generated by theme docs. Fortunately for this usage the worst case scenario is syntax highlighting not quite working.

Issue #479, opened 2024-09-04
Should not complete `endliquid` in `{% liquid %}` blocks


charlespwd:
**Describe the bug**
Should not complete `endliquid` in `{% liquid %}` blocks

**Source**

```liquid
{% liquid
  if cond
    # the following completed "endliquid"
  end|
%}
```

**Expected behaviour**
![Image](https://github.com/user-attachments/assets/14c58fa7-2b66-4433-9bac-fa86c354d683)

**Actual behaviour**
Not happening

PR #509, opened 2024-09-27
Remove `endliquid` as suggested auto-completion

aswamy:

Fixes https://github.com/Shopify/theme-tools/issues/479

## Before you deploy

### DIFF
--- a/packages/theme-language-server-common/src/completions/providers/LiquidTagsCompletionProvider.spec.ts
+++ b/packages/theme-language-server-common/src/completions/providers/LiquidTagsCompletionProvider.spec.ts
@@ -44,6 +44,22 @@ describe('Module: LiquidTagsCompletionProvider', async () => {
     await expect(provider).to.complete('{% form "cart", cart %} ... {% end', ['endform']);
   });
 
+  it('should not complete literal `liquid` tag', async () => {
+    await expect(provider).to.complete(
+      `{% liquid
+        echo 'hello'
+        e`,
+      ['echo'],
+    );
+
+    await expect(provider).not.to.complete(
+      `{% liquid
+        echo 'hello'
+        e`,
+      ['endliquid'],
+    );
+  });
+
   it('should offer the proper end tags in context', async () => {
     await expect(provider).to.complete('{% comment %} hello there {% e', ['echo', 'endcomment']);
     await expect(provider).to.complete('{% if cond %} hello {% else %} then {% e', [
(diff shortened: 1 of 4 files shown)


packages/theme-language-server-common/src/completions/providers/LiquidTagsCompletionProvider.spec.ts
```diff
@@ -44,6 +44,15 @@ describe('Module: LiquidTagsCompletionProvider', async () => {
     await expect(provider).to.complete('{% form "cart", cart %} ... {% end', ['endform']);
   });
 
+  it('should not complete literal `liquid` tag', async () => {
+    await expect(provider).to.complete(
+      `{% liquid
+        echo 'hello'
+        e`,
+      ['echo'],
```
graygilmore, line 52: Would it make sense to do something like `expect(provider).not.to.complete('code', ['endliquid'])`? Or maybe that would end up with a false positive. Almost wish there was a slightly different assertion that was like:

```ts
const completionOptions = provider.complete('code');
expect(completionOptions).not.toContain('endliquid')`
```

yarn.lock
```diff
@@ -4259,11 +4259,6 @@ ms@2.0.0:
   resolved "https://registry.yarnpkg.com/ms/-/ms-2.0.0.tgz#5608aeadfc00be6c2901df5f9861788de0d597c8"
   integrity sha512-Tpp60P6IUJDTuOq/5Z8cdskzJujfwqfOTkrwIwj7IRISpnkJnT6SyJ4PCPnGMoFjC9ddhal5KVIYtAt97ix05A==
 
-ms@2.1.2:
```
graygilmore, line 4262: I've noticed this while working locally, too. Probably safe to commit but thought I'd flag it to double check.

graygilmore (approved):

karreiro (approved):
Thank you, @aswamy! :)

packages/theme-language-server-common/src/completions/providers/LiquidTagsCompletionProvider.spec.ts
charlespwd: Funny you say this, in my Liquid Tag snippet completion PR, I modified the assertion to work with top level async matchers. I think that would make it possible with the not to complete with an expect.arrayContaining().

https://github.com/Shopify/theme-tools/pull/507/files#diff-77716827f5743abd97b0ab800c5b78475604beb65b8b5f21a56a5f18d8b82765

But, I mean... this test looks fine to me. The list of tags here is hardcoded at the top of the file, so this test would only break by manual intervention.

graygilmore: > The list of tags here is hardcoded at the top of the file, so this test would only break by manual intervention.
nice! didn't realize that

PR #516, opened 2024-10-01
Fix HTML element close tag auto-insertion inside Liquid branches

### LINKED ISSUE #515: HTML Element close tag auto-insertion doesn't work in else branches where the other branches have children
**Describe the bug**
HTML element close tag auto-insertion doesn't work in else branches when the previous branch has content in it. It's a findNodeAtCursor bug. 

**Source**

```liquid
{% if condition %}
  <div></div>
{% else %}
  <div>|   // cursor here
{% endif %}
```

**Expected behaviour**
`</div>` is auto-inserted

**Actual behaviour**
Nothing happens

charlespwd:
There was a bug with our findCurrentNode algorithm that made it so we'd go in the wrong node when a branch had child nodes.

The fix was to have more information on the AST and flag LiquidBranch as "unclosed" when their blockEndPosition is -1, -1.

It's kind of a nasty bug tbh.

Fixes #515

## Before you deploy

### DIFF
--- a/packages/theme-language-server-common/src/hover/providers/LiquidTagHoverProvider.spec.ts
+++ b/packages/theme-language-server-common/src/hover/providers/LiquidTagHoverProvider.spec.ts
@@ -18,16 +18,17 @@ describe('Module: LiquidTagHoverProvider', async () => {
   });
 
   it('should return the hover description of the correct tag', async () => {
+    // cursor always points at character before █
     await expect(provider).to.hover(`{%█ if cond %}{% endif %}`, expect.stringContaining('if'));
     await expect(provider).to.hover(`{% i█f cond %}{% endif %}`, expect.stringContaining('if'));
     await expect(provider).to.hover(`{% if█ cond %}{% endif %}`, expect.stringContaining('if'));
+    await expect(provider).to.hover(`{% if █cond %}{% endif %}`, expect.stringContaining('if'));
     await expect(provider).to.hover(`{% if cond █%}{% endif %}`, expect.stringContaining('if'));
     await expect(provider).to.hover(`{% if cond %}{% █ endif %}`, expect.stringContaining('if'));
     await expect(provider).to.hover(`{% echo█ 'hi' %}`, expect.stringContaining('echo'));
   });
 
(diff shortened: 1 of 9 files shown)


packages/theme-language-server-common/src/hover/providers/LiquidTagHoverProvider.spec.ts
```diff
@@ -18,16 +18,17 @@ describe('Module: LiquidTagHoverProvider', async () => {
...
     await expect(provider).to.hover(`{% i█f cond %}{% endif %}`, expect.stringContaining('if'));
     await expect(provider).to.hover(`{% if█ cond %}{% endif %}`, expect.stringContaining('if'));
+    await expect(provider).to.hover(`{% if █cond %}{% endif %}`, expect.stringContaining('if'));
     await expect(provider).to.hover(`{% if cond █%}{% endif %}`, expect.stringContaining('if'));
     await expect(provider).to.hover(`{% if cond %}{% █ endif %}`, expect.stringContaining('if'));
     await expect(provider).to.hover(`{% echo█ 'hi' %}`, expect.stringContaining('echo'));
   });
 
   it('should not return the tag hover description when hovering over anything else in the tag', async () => {
-    await expect(provider).to.not.hover(`{% if █cond %}{% endif %}`, expect.stringContaining('if'));
```
charlespwd, line 30: The cursor points to the the thing _left_ of the cursor, so you're not technically _in cond_ when your cursor is before the `c`.

packages/theme-language-server-common/src/visitor.ts
```diff
@@ -1,9 +1,10 @@
+import { NodeTypes } from '@shopify/liquid-html-parser';
 import {
   AST,
   LiquidHtmlNode,
   NodeOfType,
-  NodeTypes,
   SourceCodeType,
+  NodeTypes as TCNodeTypes,
```
charlespwd, line 7: No longer need this part. I ended up getting rid of the switch statement.

aswamy (approved):
🎩 ✅ 

Nice find + fix!

PR #520, opened 2024-10-03
Support HTML attribute auto-completion

aswamy:

- Support HTML attribute auto-completion
- Override entire attribute even if the cursor is  in the middle of the attribute
- Add `=""` at the end of an attribute with a value
- Allow attribute completion when cursor is not on any character
- NOTE: Does NOT replace Liquid Code when auto-completing HTML attribute
  - Why? This seems like one of those trade-offs where we _could_ try to parse the string and assume the whole attribute (including any liquid code) is part of the tag, but we'd be overriding code which i assume isn't what the developer wants. If they really wish to remove the liquid code, they must do so manually.

### What we support

| before code completion| after code completion (once you select it from dropdown) |
| - | - |
| `<a hr█` |  `<a href=""` |
| `<button disab█` |  `<button disabled` |
| `<a █` |  `<a href=""` |
| `<button s█tyle=""` |  `<button spellcheck=""` |
| `<button d█ata--{{ echo "demo" }}=""` |  `<button dir{{ echo "demo" }}=""` |

## What did you learn?

- Edges and tradeoffs on what to support and what to ignore 🤕

## Before you deploy

### DIFF
--- a/.changeset/quiet-rice-brake.md
+++ b/.changeset/quiet-rice-brake.md
@@ -1,0 +1,8 @@
+---
+'@shopify/theme-language-server-common': minor
+---
+
+Support HTML attribute auto-completion
+
+- Add `=""` to attributes that have values
+- Override whole attribute even if cursor is in the middle of it
--- a/.changeset/quiet-rice-brake.md
+++ b/.changeset/quiet-rice-brake.md
@@ -2,7 +2,7 @@
 '@shopify/theme-language-server-common': minor
 ---
 
-Support HTML attribute auto-completion
+Improve HTML attribute auto-completion
 
-- Add `=""` to attributes that have values
-- Override whole attribute even if cursor is in the middle of it
+- Add `=""` to attributes that should have a value when relevant
+- Override entire attribute even if the cursor was in the middle of it
--- a/packages/theme-language-server-common/src/completions/CompletionsProvider.ts
+++ b/packages/theme-language-server-common/src/completions/CompletionsProvider.ts
@@ -49,7 +49,7 @@ export class CompletionsProvider {
 
     this.providers = [
       new HtmlTagCompletionProvider(),
-      new HtmlAttributeCompletionProvider(),
+      new HtmlAttributeCompletionProvider(documentManager),
       new HtmlAttributeValueCompletionProvider(),
       new LiquidTagsCompletionProvider(themeDocset),
       new ObjectCompletionProvider(typeSystem),
--- a/packages/theme-language-server-common/src/completions/providers/HtmlAttributeCompletionProvider.spec.ts
+++ b/packages/theme-language-server-common/src/completions/providers/HtmlAttributeCompletionProvider.spec.ts
@@ -38,14 +38,21 @@ describe('Module: HtmlAttributeCompletionProvider', async () => {
   });
 
   it('should replace the existing attribute partial with the full attribute tag', async () => {
-    await expect(provider).to.complete('<text titl█', [
+    const attributePartial = 'titl';
+    const source = `<text ${attributePartial}█>`;
+    const sourceWithoutCursor = source.replace('█', '');
+
+    await expect(provider).to.complete(source, [
       expect.objectContaining({
         label: 'title',
         textEdit: {
           newText: 'title="$1"$0',
           range: {
-            start: { line: 0, character: 6 },
-            end: { line: 0, character: 10 },
+            start: { line: 0, character: sourceWithoutCursor.indexOf(attributePartial) },
+            end: {
+              line: 0,
+              character: sourceWithoutCursor.indexOf(attributePartial) + attributePartial.length,
+            },
           },
         },
       }),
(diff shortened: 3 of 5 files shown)


packages/theme-language-server-common/src/test/CompletionItemsAssertion.spec.ts
```diff
@@ -19,12 +19,12 @@ describe('Module: CompletionItemsAssertion', () => {
...
-  it('should assert a list of labels', () => {
-    expect(provider).to.complete('{% rend', ['render']);
+  it('should assert a list of labels', async () => {
+    await expect(provider).to.complete('{% rend', ['render']);
   });
 
-  it('should assert a list of completion items', () => {
-    expect(provider).to.complete('{% rend', [
+  it('should assert a list of completion items', async () => {
+    await expect(provider).to.complete('{% rend', [
```
aswamy, line 27: without the `await`, the test was always passing even when i changed the expected result

packages/theme-language-server-common/src/completions/CompletionsProvider.ts
```diff
@@ -49,7 +49,7 @@ export class CompletionsProvider {
 
     this.providers = [
       new HtmlTagCompletionProvider(),
-      new HtmlAttributeCompletionProvider(),
+      new HtmlAttributeCompletionProvider(documentManager),
```
aswamy, line 52: I won't have to pass in the `documentManager` once this change is introduced
https://github.com/Shopify/theme-tools/pull/507/files#diff-d0fc601312d376218785407c3e38c31fda0c22b4adec2702bdc3aee7ff24eb92R34

packages/theme-language-server-common/src/completions/providers/HtmlAttributeCompletionProvider.spec.ts
```diff
@@ -36,4 +36,83 @@ describe('Module: HtmlAttributeCompletionProvider', async () => {
...
+        return expect.objectContaining({
+          textEdit: {
+            newText: expect.stringContaining(attr),
+            range: {
+              start: { line: 0, character: 3 },
+              end: { line: 0, character: 9 },
+            },
+          },
+        });
+      }),
```
charlespwd, line 115: Actually I think we _want_ to replace the liquid part of the attribute name in this case. 🤔

packages/theme-language-server-common/src/completions/providers/HtmlAttributeCompletionProvider.ts
```diff
@@ -32,7 +41,49 @@ export class HtmlAttributeCompletionProvider implements Provider {
...
+      // E.g. <a href="" █>
+      return {
+        start: document.textDocument.positionAt(node.position.start),
+        end: document.textDocument.positionAt(node.position.start),
+      };
+    }
+
+    let attributeEndOffset =
+      document.source.slice(node.position.end).match(/[\s=]|{%|{{|>/)?.index ??
+      document.source.length;
```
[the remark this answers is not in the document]
aswamy: The reason i chose offset to be `document.source.length` instead of `0` is for this specific edge case:

```
<script d█efer
```

When we are trying to find the end of an attribute, we rely on space-like character, `=`, `{%`, or `{{`. The example above has none of those; the end of the document IS the attribute.

Im not sure there are any other ways i can do this 🤔

packages/theme-language-server-common/src/completions/providers/HtmlAttributeCompletionProvider.spec.ts
```diff
@@ -36,4 +36,83 @@ describe('Module: HtmlAttributeCompletionProvider', async () => {
     const expected = [...aTagAttributeNames, ...globalAttributeNames].sort();
     await expect(provider).to.complete('<a {% if cond %}█', expected);
   });
+
+  it('should replace the existing attribute partial with the full attribute tag', async () => {
```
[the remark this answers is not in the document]
aswamy: Hmm good catch. I thought i fixed it here, but i can add a test to make sure it's actually doing it properly 🤔 

https://github.com/Shopify/theme-tools/pull/520/files#diff-de8c51ef0e3e9766a3f1feaa4396111123079aecde36d33e69e3fedd0b2eddd7R69-R77

packages/theme-language-server-common/src/completions/providers/HtmlAttributeCompletionProvider.ts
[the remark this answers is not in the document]
aswamy: OH i think i see what you mean...

It should probably be this

```
    const sourcePartialPastCursor = document.source.slice(node.position.end);

    let attributeEndOffset =
      sourcePartialPastCursor.match(/[\s=]|{%|{{|>/)?.index ??
      sourcePartialPastCursor.length;
```

packages/theme-language-server-common/src/completions/providers/HtmlAttributeCompletionProvider.spec.ts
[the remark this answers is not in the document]
aswamy: I updated the existing test cases to include characters after the attribute partial. I tophatted it again, and couldn't replicate what you did 🤔

.changeset/quiet-rice-brake.md
```diff
@@ -2,7 +2,8 @@
 '@shopify/theme-language-server-common': minor
 ---
 
-Support HTML attribute auto-completion
+Improve HTML attribute auto-completion
 
 - Add `=""` to attributes that have values
-- Override whole attribute even if cursor is in the middle of it
+  - Ignore if it attribute also has value before auto-completion is triggered
```
charlespwd, line 8:
```suggestion
```

I don't think we need to document this in the changelog. Feels more like "this is how it's supposed to work" thing than it is a feature (I think?) 

Maybe reword to `Snippet complete HTML attributes (Add `=""`) when relevant`?

packages/theme-language-server-common/src/completions/CompletionsProvider.ts
charlespwd: 507 is merged if you want to rebase and do that. But honestly not a big deal.

Issue #19, opened 2023-08-10
ObjectCompletionProvider and filter arguments


charlespwd:
**Is your feature request related to a problem? Please describe.**
filters.json gives us the expected types of arguments and named parameters.

**Describe the solution you'd like**
- We offer completion options for the name of the named arguments of filters
- We offer completion options of the appropriate type (e.g. width only accepts variables of type number, etc.)

![Image](https://github.com/user-attachments/assets/25726660-4336-45b3-a876-073acf400299)

PR #522, opened 2024-10-07
Add completion support for filter parameters

graygilmore:

- Closes https://github.com/Shopify/theme-tools/issues/19

This adds two new features to the liquid language server!

- Automatically adds required parameter to filter completions
- Adds support for completion options for filter parameters

Before we could fully ship these features we needed first to add support to the Liquid docs generator to differentiate between named and positional parameters for filters. This has since been completed and we're now good to go! You can see the updated schema here: https://github.com/Shopify/theme-liquid-docs/pull/824

### Automatically add required parameters to filter completions

If you use completion for a filter that has required parameters we'll automatically add them with tab support.

completion

If the required parameter is a string it'll automatically add the necessary quotes:

colorcompletion

### Completion support for filter parameters

If you have something like the following:

```liquid
{{ product | image_url: █ }}
```

We'll show completion options for any of the named parameters for the filter.

image

> [!NOTE]
> We cannot add completion support for positional parameters since there'd be nothing for us to "complete". We'll be adding better type support for these filters in a later PR that will allow developers to see which parameters are available which should help fill this gap.

## Before you deploy

### DIFF
--- a/packages/theme-language-server-common/src/completions/providers/FilterCompletionProvider.spec.ts
+++ b/packages/theme-language-server-common/src/completions/providers/FilterCompletionProvider.spec.ts
@@ -171,11 +171,26 @@ describe('Module: FilterCompletionProvider', async () => {
 
   describe('when there are no required parameters', () => {
     it('should not include parameters in the insertText of the completion', async () => {
-      await expect(provider).to.complete('{{ string | defa█ }}', [
+      //                  char 12 ⌄   ⌄ char 16
+      const liquid = '{{ string | defa█ }}';
+
+      await expect(provider).to.complete(liquid, [
         expect.objectContaining({
           label: 'default',
-          insertText: 'default',
           insertTextFormat: 1,
+          textEdit: expect.objectContaining({
+            newText: 'default',
+            range: {
+              end: {
+                line: 0,
+                character: 16,
+              },
+              start: {
+                line: 0,
+                character: 12,
+              },
+            },
+          }),
         }),
       ]);
     });
@@ -183,11 +198,26 @@ describe('Module: FilterCompletionProvider', async () => {
 
   describe('when there are required positional parameters', () => {
     it('should include parameters in the insertText of the completion', async () => {
-      await expect(provider).to.complete('{{ string | h█ }}', [
+      //                  char 12 ⌄  ⌄ char 15
+      const liquid = '{{ string | hig█ }}';
+
+      await expect(provider).to.complete(liquid, [
         expect.objectContaining({
           label: 'highlight',
-          insertText: "highlight: '${1:highlighted_term}'",
           insertTextFormat: 2,
+          textEdit: expect.objectContaining({
+            newText: "highlight: '${1:highlighted_term}'",
+            range: {
+              end: {
+                line: 0,
+                character: 15,
+              },
+              start: {
+                line: 0,
+                character: 12,
+              },
+            },
+          }),
         }),
       ]);
     });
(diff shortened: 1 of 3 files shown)


packages/theme-language-server-common/src/completions/providers/ObjectCompletionProvider.ts
```diff
@@ -20,6 +20,13 @@ export class ObjectCompletionProvider implements Provider {
       return [];
     }
 
+    const currentContext = params.completionContext.ancestors.at(-1);
+    // We don't want to show generic variable completions when we're inside a
+    // filter. This case will be handled by FilterNamedParameterCompletionProvider.
+    if (currentContext?.type === NodeTypes.LiquidFilter) {
```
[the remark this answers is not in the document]
charlespwd: We probably could use some kind of `sortText` `CompletionItem` stuff to make sure that the properties are shown before the variables?

packages/theme-language-server-common/src/completions/providers/FilterNamedParameterCompletionProvider.ts
```diff
@@ -0,0 +1,57 @@
...
+          entry: {
+            ...parameter,
+          },
+          types: parameter.types,
+        };
+      });
+
+    return options.map(({ entry, types }) =>
+      createCompletionItem(
+        entry,
```
aswamy, line 50: Is there any limitations today that prevent us from putting a ":" after the param? It would be great if we can do something like `${param}: $0`.

packages/theme-language-server-common/src/completions/providers/FilterNamedParameterCompletionProvider.ts
```diff
@@ -0,0 +1,57 @@
...
+        return {
+          entry: {
+            ...parameter,
+          },
+          types: parameter.types,
+        };
+      });
+
+    return options.map(({ entry, types }) =>
+      createCompletionItem(
```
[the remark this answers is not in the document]
graygilmore: Interesting! Good call out. That's annoying to have to handle but makes sense in terms of total completeness of this (and matching other behavior).

packages/theme-language-server-common/src/completions/providers/FilterNamedParameterCompletionProvider.ts
graygilmore: There is! Not all params are named and will require a `:`. I'm working on updating the source of the docs to include this so that we can do that!

packages/theme-language-server-common/src/completions/providers/FilterCompletionProvider.spec.ts
```diff
@@ -133,6 +168,42 @@ describe('Module: FilterCompletionProvider', async () => {
...
+          insertText: "highlight: '${1:highlighted_term}'",
+          insertTextFormat: 2,
+        }),
+      ]);
+    });
+  });
+
+  describe('when there are required named parameters', () => {
+    it('should include parameters in the insertText of the completion', async () => {
+      await expect(provider).to.complete('{{ string | pre█ }}', [
```
charlespwd, line 198: Can we add a test to make sure the required args are not added if text follows?

I typically use `TextEdit` instead of `insertText` in order to maybe replace stuff past the cursor. It might not apply here, but we probably don't want `width: 100, height: 100` to be appended the second time you accept the image_url filter 🤔

graygilmore: I should have added this to the PR description that I'm aware of this bug but specifically not addressing it here. I didn't want to balloon this PR any more and it's an outstanding bug. For example, even on `main` we get the duplicate filter:

bug

This change does make it *worse*, though:

badbug

graygilmore: I wonder if I can abstract https://github.com/Shopify/theme-tools/pull/522/files#diff-7256ee5482c911787e4800568fe9ea339c55f5a946fa72793694f4b2cdfe3b9aR70 into some shared code 🤔 

I'll have a look at this

graygilmore: This was hard! https://github.com/Shopify/theme-tools/pull/522/commits/4b2dec31b946b8e92d8601390699c0af94d0274c

charlespwd (approved):
🔥 This is amazing. 🎩 looks great! Don't let the test comment stop you :)

packages/theme-language-server-common/src/completions/providers/FilterCompletionProvider.spec.ts
```diff
@@ -133,6 +168,161 @@ describe('Module: FilterCompletionProvider', async () => {
...
+              start: {
+                line: 0,
+                character: 12,
+              },
+            },
+          }),
+        }),
+      ]);
+    });
+  });
```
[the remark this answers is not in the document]
graygilmore: 💀 I am dead inside haha

On the plus side: doing it my way was a really good way to figure out why a particular bug was happening in another branch!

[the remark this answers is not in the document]
graygilmore: Played around with this a bit and I'll definitely make the changes! I'll do it in a follow up PR.

PR #524, opened 2024-10-07
Fix broken tests

aswamy:

The [last commit](https://github.com/Shopify/theme-tools/pull/520) found a typo where `CompletionItemsAssertion` tests weren't running, but after fixing it, CI didn't warn us that the tests were broken. Patching the broken test.

CI on merge: https://github.com/Shopify/theme-tools/actions/runs/11223315187/job/31197681873


graygilmore (approved):
> also seems to be broken on my local environment, but doesn't appear as broken in the CI

might need a `yarn build` before running the tests

aswamy:
> might need a yarn build before running the tests

Yep that did it! Thanks @graygilmore

PR #529, opened 2024-10-08
Virtual File Systems & Shopify Liquid VS Code extension for Web

charlespwd:

- Major breaking changes in most packages (mostly internal APIs)
   - `vscode-extension`:
       - We now have a `browser` entry in our `package.json`
       - The `src` folder has been split into `node`, `common` and `browser` folders
       - The `browser` module starts the language server in a Web Worker
       - The `node` module starts the language server in a subprocess 
       - Dropped support for CLI based language servers (legacy extension)
       - There's two new ways to test/debug the VS Code extension in browser contexts (see tophatting section below).
   - `theme-check-*`, and `theme-language-server-*`:
     - Replace `AbsolutePath` concerns with `Uri`s. 
         - Absolute paths don't make sense in browser
         - Absolute paths don't make sense with remote files
   - Replace all "fs-based" `Dependencies` with a single `fs: AbstractFileSystem` dependency
       - The old injections are now common implementations based on the abstract `fs`
       - For the CLI theme check, a `NodeFileSystem` is available (and used by default)
       - For the CLI language server, a `NodeFileSystem` is available (and used by default)
       - For the VS Code extension, a `VsCodeFileSystem` is available
           - This file system implementation queries the [VS Code FileSystem API](https://code.visualstudio.com/api/references/vscode-api#FileSystem) (only available in the extension host thread) for file contents. This query is done on the language server via non-standard LSP messages:
               - `fs/readFile` -> a TextDecoded `vscode.workspace.fs.readFile(uri)`
               - `fs/readDirectory` -> `vscode.workspace.fs.readDirectory(uri)`
               - `fs/stat` -> `vscode.workspace.fs.stat(uri)`
           - In `node`, the same happens but short circuits to the `NodeFileSystem` for `file:///` URIs.
               - We don't need to send file contents over STDIO when we can query the FS directly
   - The `codemirror-language-client`'s playground implements a mock "MainThreadFileSystem" that we could use as inspiration for online-store-web

## Demo

https://github.com/user-attachments/assets/bade2bd3-ccf3-4f63-a8f8-15dceab0dda9

   
## Top-hatting

- There's 2 _new_ ways you can do this. They are documented in docs/contributing. But here's a video. 

### Debugging in Chromium VS Code for Web

The "true" E2E way. Demo worthy version of "showing it works in browser"

https://github.com/user-attachments/assets/f556dda1-11a6-4647-b353-cd70c8351b5c

### Debugging in the Desktop app

A faster, debuggable way that runs the browser extension in the desktop app. This disables all the Node APIs but it lets you debug the code.

https://github.com/user-attachments/assets/6601c569-6f54-4c8d-a2f1-867a31a532f3

### Debugging the CodeMirror Language client

(Not new, but you should know that this flow is also updated to virtual file systems). 

https://github.com/user-attachments/assets/90effdf0-6ee8-4771-8cb1-1417854344a8

- Make `loadConfig` work in-browser
- [x] Make the prettier plugin work in-browser

## What did you learn?

- A couple of neat things that the vscode libraries already offer. 
   - The web worker stuff is offered via `BrowserMessageReader(worker)` and `BrowserMessageWriter(worker)`. 
   - `vscode.workspace.fs` is pretty cool. It does a pretty clever the URIScheme -> FileSystemProvider mapping that makes it so we could eventually have a "ShopifyBackendFileSystem" that would write directly to the storefront. 

## Before you deploy

- [x] I included a bunch major bump `changeset`


packages/vscode-extension/package.json
```diff
@@ -40,6 +40,7 @@
     "dev:build": "webpack --mode development",
     "dev:syntax": "yarn --cwd ./syntaxes dev",
     "dev:watch": "webpack --mode development --watch",
+    "dev:web": "yarn build && vscode-test-web --permission=clipboard-read --permission=clipboard-write --browserType=chromium --extensionDevelopmentPath=.",
```
charlespwd, line 43: vscode-test-web is the one that lets you run chrome and have vscode run in it with your extension loaded. It was a real breeze ngl!

packages/vscode-extension/package.json
```diff
@@ -83,7 +85,8 @@
   "activationEvents": [
     "workspaceContains:**/.theme-check.yml"
   ],
-  "main": "./dist/extension.js",
+  "main": "./dist/node/extension.js",
+  "browser": "./dist/browser/extension.js",
```
charlespwd, line 89: Major difference is here: We have a different build for web that uses `theme-language-server-browser`

packages/vscode-extension/package.json
```diff
@@ -98,6 +101,18 @@
...
+            "string"
+          ],
+          "description": "Traces the communication between the client and the server.",
+          "default": "off",
+          "enum": [
+            "off",
+            "messages",
+            "verbose"
+          ]
+        },
```
charlespwd, line 115: Added that stuff in there because I was tired of not having code completion in settings.json to enable the LSP logging

packages/vscode-extension/src/browser/extension.ts
```diff
@@ -0,0 +1,117 @@
...
+  const worker = new Worker(serverMain.toString(true));
+
+  // create the language server client to communicate with the server running in the worker
+  return new LanguageClient(
+    'shopifyLiquid',
+    'Theme Check Language Server',
+    clientOptions,
+    worker,
+  );
+}
```
charlespwd, line 117: Major difference is that here we start the language server in a Web Worker instead of settingup STDIO and a subprocess.

packages/vscode-extension/src/node/extension.ts
```diff
@@ -92,6 +92,8 @@ async function startServer(context: ExtensionContext) {
     ],
   };
 
+  workspace.fs.readDirectory
```
charlespwd, line 95: oops shouldn't be there.

packages/theme-language-server-common/src/server/startServer.ts
```diff
@@ -46,21 +53,15 @@ const defaultLogger = () => {};
...
   {
-    fileExists,
-    fileSize,
-    filesForURI,
-    findRootURI: findConfigurationRootURI,
-    getDefaultLocaleFactory,
-    getDefaultTranslationsFactory,
-    getDefaultSchemaLocaleFactory,
-    getDefaultSchemaTranslationsFactory,
-    getThemeSettingsSchemaForRootURI,
```
charlespwd, line 57: :)

packages/vscode-extension/src/node/extension.ts
```diff
@@ -81,14 +73,14 @@ async function startServer(context: ExtensionContext) {
...
-      { scheme: 'file', language: 'json' },
-      { scheme: 'file', language: 'jsonc' },
+      { language: 'liquid' },
+      { language: 'plaintext' },
+      { language: 'html' },
+      { language: 'javascript' },
+      { language: 'css' },
+      { language: 'scss' },
+      { language: 'json' },
+      { language: 'jsonc' },
```
charlespwd, line 83: This thing took me a while to figure out since I wrote this 3 years ago. We need to remove `scheme` in order for completion requests from `vscode-vfs://` URIs to trigger.

packages/vscode-extension/webpack.config.js
```diff
@@ -94,4 +95,136 @@ const config = {
...
+    mode: 'none',
+    entry: {
+      server: './src/browser/server.ts',
+    },
+    output: {
+      path: path.resolve(__dirname, 'dist', 'browser'),
+      filename: '[name].js',
+      libraryTarget: 'var',
+      library: 'serverExportVar',
+      devtoolModuleFilenameTemplate: '../../[resource-path]',
```
charlespwd, line 180: The only reason I need 2x a config that should look like one is because of this difference here. The worker has to be "import-less", but the extension needs a `require('vscode')` call in the output that gets shimmed.

packages/vscode-extension/src/browser/extension.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  const textDecoder = new TextDecoder();
+  client.onRequest('fs/readFile', async (uriString: string): Promise<string> => {
+    const bytes = await workspace.fs.readFile(Uri.parse(uriString));
+    return textDecoder.decode(bytes);
+  });
+
+  client.onRequest('fs/stat', async (uriString: string): Promise<FileStat> => {
+    return workspace.fs.stat(Uri.parse(uriString));
+  });
```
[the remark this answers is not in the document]
charlespwd: This is the one that calls the [VS Code FileSystem API](https://code.visualstudio.com/api/references/vscode-api#FileSystem)

```ts
vscode.workspace.fs.readFile() // returns a UInt8Array (bytes)
// AbstractFileSystem.readFile is Promise<string>, this is why we do textDecoder.decode(bytes) on the result. 
```

packages/theme-check-common/src/checks/translation-key-exists/index.spec.ts
```diff
@@ -169,7 +169,7 @@ describe('Module: TranslationKeyExists', () => {
       check: TranslationKeyExists.meta.code,
       message:
         "'shopify.this.does.not.exist' does not have a matching entry in 'locales/en.default.json'",
-      absolutePath: '/code.liquid',
+      // uri: 'file:///code.liquid',
```
aswamy, line 172:
```suggestion
      uri: 'file:///code.liquid',
```

packages/theme-check-node/src/temp.ts
```diff
@@ -0,0 +1 @@
+export type AbsolutePath = string;
```
aswamy, line 1: Is this file created on accident? I think you wanted to place this code in `path.ts` since that file is empty (and probably wanted to delete this file)

packages/theme-check-common/src/test/MockFileSystem.spec.ts
```diff
@@ -0,0 +1,71 @@
+import { beforeEach, describe, expect, it } from 'vitest';
+import { MockFileSystem } from './MockFileSystem';
+import { FileType } from '../AbstractFileSystem';
+
+describe('MockFileSystem', () => {
+  let fs: MockFileSystem;
```
aswamy, line 6: Idk why but testing a mock file system entirely made for testing with more tests is pretty funny to me 😄

packages/theme-check-common/src/context-utils.ts
```diff
@@ -0,0 +1,142 @@
...
+}
+
+function cached<T>(fn: () => Promise<T>): () => Promise<T>;
+function cached<T>(fn: (...args: any[]) => Promise<T>): (...args: any[]) => Promise<T> {
+  let cachedPromise: Promise<T>;
+  return async (...args) => {
+    if (!cachedPromise) cachedPromise = fn(...args);
+    return cachedPromise;
+  };
+}
```
aswamy, line 111:
```suggestion
function cached<T>(fn: (...args: any[]) => Promise<T>): (...args: any[]) => Promise<T> {
  let cachedPromise: Promise<T>;
  return async (...args) => {
    if (!cachedPromise) cachedPromise = fn(...args);
    return cachedPromise;
  };
}
```

aswamy (approved):
Sorry took some time, but we should be good. Minor comments + concerns, but nothing blocking atm.

Tophat complete:
- Able to run VScode web, able to edit files, save it
- Able to run VScode web in "sandbox" mode (vscode-test-web), able to edit files - does not persist file as expected
- Able to run desktop VSCode with web extension, able to edit files, save it
- Able to run Codemirror Client, able to edit files, save it

⚠️ Noticed that when you open any files in the repo, there is a weird error where we look for `.theme-check.yml` file in the wrong directory. It could be one of the rootPath functions are messed up because i saw this in all VScode scenarios above. However, Node extension had it as a thrown error, while Browser Extension had it as a info-message.

image

package.json
```diff
@@ -23,6 +23,7 @@
     "build": "yarn workspaces run build",
     "build:ci": "yarn workspaces run build:ci",
     "build:ts": "yarn workspaces run build:ts",
+    "dev:web": "yarn --cwd packages/vscode-extension dev:web",
```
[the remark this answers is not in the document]
charlespwd: Durr I have a `.nvmrc` but I git ignored/excluded it? Yeah def should do

packages/theme-language-server-common/src/translations.ts
```diff
@@ -1,6 +1,12 @@
-import { SourceCodeType, Translations, isError, parseJSON } from '@shopify/theme-check-common';
+import {
+  SourceCodeType,
+  Translations,
+  isError,
+  makeGetDefaultSchemaTranslations,
+  makeGetDefaultTranslations,
+  parseJSON,
+} from '@shopify/theme-check-common';
```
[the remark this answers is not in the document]
charlespwd: That _would_ be nice indeed. I love the prettier solution and it's already caught by CI.

packages/theme-language-server-common/src/server/startServer.ts
```diff
@@ -354,14 +355,32 @@ export function startServer(
   connection.workspace.onDidCreateFiles((params) => {
     const triggerUris = params.files.map((fileCreate) => fileCreate.uri);
     runChecks.force(triggerUris);
+    for (const { uri } of params.files) {
+      fs.readDirectory.invalidate(path.dirname(uri));
+      fs.readFile.invalidate(uri);
+      fs.stat.invalidate(uri);
```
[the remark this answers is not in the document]
charlespwd: if `fs.stat` returned nothing before (because it didn't exist), then the cache should be invalidated because now it returns something!

packages/theme-language-server-common/src/documentLinks/DocumentLinksProvider.spec.ts
```diff
@@ -16,7 +16,7 @@ describe('DocumentLinksProvider', () => {
     const uriString = 'file:///path/to/non-liquid-html-document.txt';
     const rootUri = 'file:///path/to/project';
 
-    documentManager.open(URI.parse(uriString).toString(), 'Sample plain text content', 1);
+    documentManager.open(path.normalize(uriString), 'Sample plain text content', 1);
```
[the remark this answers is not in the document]
charlespwd: You're probably right. The normalization happens because of Windows URIs with the `\` in them. URI.parse().toString() without the `true` encodes the `\` as a `%` encoded string and it makes the URIs unreadable. Plus any logic that would have used the replace `\` by `/` would not work. 

This diff happened because I was looking for all the URI...toString() in the code and was replacing it with normalize instead.

packages/theme-check-node/src/temp.ts
charlespwd: absolutely correct haha. That's a leftover from the refactor.

Issue #502, opened 2024-09-25
Metafield completion


charlespwd:
**Is your feature request related to a problem? Please describe.**
Accessing metafield values in a theme is [_kind of complicated_.](https://shopify.dev/docs/api/liquid/objects#metafield-access-metafields). There's probably something smart we can do that would make it super easy to have code completion for those in a shop-aware scenario. 

**Describe the solution you'd like**
- Use dependency injection of some kind of data structure so that the browser and node environments have access to a shop's metafield config
- Wire the metafield definitions into the type system so that code completion works as though they were normal types. 

**Describe alternatives you've considered**
- Using dependency injection so that the core does the graphql calls to get the stuff. I don't like that because it creates a network dependency and error handling can get weird. I'd like to be able to keep running theme-check in CI without making API calls. Probably something like dumping the definitions to a file and providing those. In web, we could probably do the graphql query to get them.

**Additional context**

PR #534, opened 2024-10-18
Auto-complete metafields depending on shop data

aswamy:

Part 1 for https://github.com/Shopify/theme-tools/issues/502
- Adding support for metadata autocompletion on our editors
- listens to file changes for `.shopify` folder
- This should work for VScode only

- We need to make it work on OSE next
- We will create a CLI command to fetch the metafields.json
- We need to automatically call this CLI command in VSCode on startup

## What did you learn?

- registering for `DidChangeWatchedFilesNotification` took some time, but had to read-up on client capabilities and how to subscribe to notifications after initialization

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles

## Tophat

Run the following query and dump the contents into `.shopify/metafields.json` in the theme repository. Remove the `nodes` array, and place it directory under each group.

```
query {
	article: metafieldDefinitions(ownerType: ARTICLE, first: 250) {
		nodes {
			...MetafieldDefinitionFields
		}
	}
	blog: metafieldDefinitions(ownerType: BLOG, first: 250) {
		nodes {
			...MetafieldDefinitionFields
		}
	}
	brand: metafieldDefinitions(ownerType: BRAND, first: 250) {
		nodes {
			...MetafieldDefinitionFields
		}
	}
	collection: metafieldDefinitions(ownerType: COLLECTION, first: 250) {
		nodes {
			...MetafieldDefinitionFields
		}
	}
	company: metafieldDefinitions(ownerType: COMPANY, first: 250) {
		nodes {
			...MetafieldDefinitionFields
		}
	}
	company_location: metafieldDefinitions(ownerType: COMPANY_LOCATION, first: 250) {
		nodes {
			...MetafieldDefinitionFields
		}
	}
	location: metafieldDefinitions(ownerType: LOCATION, first: 250) {
		nodes {
			...MetafieldDefinitionFields
		}
	}
	market: metafieldDefinitions(ownerType: MARKET, first: 250) {
		nodes {
			...MetafieldDefinitionFields
		}
	}
	order: metafieldDefinitions(ownerType: ORDER, first: 250) {
		nodes {
			...MetafieldDefinitionFields
		}
	}
	page: metafieldDefinitions(ownerType: PAGE, first: 250) {
		nodes {
			...MetafieldDefinitionFields
		}
	}
	product: metafieldDefinitions(ownerType: PRODUCT, first: 250) {
		nodes {
			...MetafieldDefinitionFields
		}
	}
	variant: metafieldDefinitions(ownerType: PRODUCTVARIANT, first: 250) {
		nodes {
			...MetafieldDefinitionFields
		}
	}
	shop: metafieldDefinitions(ownerType: SHOP, first: 250) {
		nodes {
			...MetafieldDefinitionFields
		}
	}
}

fragment MetafieldDefinitionFields on MetafieldDefinition {
	name
	namespace
	description
	type {
		category
		name
	}
}
```

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/index.ts
+++ b/packages/theme-check-common/src/index.ts
@@ -7,6 +7,7 @@ import {
   makeGetDefaultSchemaLocale,
   makeGetDefaultSchemaTranslations,
   makeGetDefaultTranslations,
+  makeGetMetafieldDefinitions,
 } from './context-utils';
 import { createDisabledChecksModule } from './disabled-checks';
 import { isIgnored } from './ignore';
@@ -74,6 +75,8 @@ export async function check(
     getDefaultTranslations: makeGetDefaultTranslations(fs, theme, rootUri),
     getDefaultSchemaLocale: makeGetDefaultSchemaLocale(fs, rootUri),
     getDefaultSchemaTranslations: makeGetDefaultSchemaTranslations(fs, theme, rootUri),
+    getMetafieldDefinitions:
+      injectedDependencies.getMetafieldDefinitions ?? makeGetMetafieldDefinitions(fs),
   };
 
   const { DisabledChecksVisitor, isDisabled } = createDisabledChecksModule();
--- a/packages/theme-language-server-common/src/TypeSystem.ts
+++ b/packages/theme-language-server-common/src/TypeSystem.ts
@@ -14,6 +14,8 @@ import {
   ArrayReturnType,
   DocsetEntry,
   FilterEntry,
+  MetafieldDefinitionMap,
+  MetafieldDefinition,
   ObjectEntry,
   ReturnType,
   SourceCodeType,
@@ -21,6 +23,7 @@ import {
   isError,
   parseJSON,
   path,
+  FETCHED_METAFIELD_CATEGORIES,
 } from '@shopify/theme-check-common';
 import {
   GetThemeSettingsSchemaForURI,
@@ -35,6 +38,7 @@ export class TypeSystem {
   constructor(
     private readonly themeDocset: ThemeDocset,
     private readonly getThemeSettingsSchemaForURI: GetThemeSettingsSchemaForURI,
+    private readonly getMetafieldDefinitions: (rootUri: string) => Promise<MetafieldDefinitionMap>,
   ) {}
 
   async inferType(
@@ -109,9 +113,10 @@ export class TypeSystem {
    * e.g. objectMap['product'] returns the product ObjectEntry.
    */
   public objectMap = async (uri: string, ast: LiquidHtmlNode): Promise<ObjectMap> => {
-    const [objectMap, themeSettingProperties] = await Promise.all([
+    const [objectMap, themeSettingProperties, metafieldDefinitionsObjectMap] = await Promise.all([
       this._objectMap(),
       this.themeSettingProperties(uri),
+      this.metafieldDefinitionsObjectMap(uri),
     ]);
 
     // Here we shallow mutate `settings.properties` to have the properties made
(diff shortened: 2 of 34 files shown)


packages/theme-check-common/src/index.ts
```diff
@@ -74,6 +75,7 @@ export async function check(
     getDefaultTranslations: makeGetDefaultTranslations(fs, theme, rootUri),
     getDefaultSchemaLocale: makeGetDefaultSchemaLocale(fs, rootUri),
     getDefaultSchemaTranslations: makeGetDefaultSchemaTranslations(fs, theme, rootUri),
+    getMetafieldDefinitions: () => Promise.resolve({} as MetafieldDefinitionMap),
```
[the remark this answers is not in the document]
aswamy: Theme check doesn't do anything yet, so we stubbed the method. But i can change it even though it doesn't get used.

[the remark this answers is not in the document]
charlespwd: Bah might as well do it. Would be easy to forget :P

packages/theme-check-common/src/index.ts
```diff
@@ -74,6 +75,7 @@ export async function check(
     getDefaultTranslations: makeGetDefaultTranslations(fs, theme, rootUri),
     getDefaultSchemaLocale: makeGetDefaultSchemaLocale(fs, rootUri),
     getDefaultSchemaTranslations: makeGetDefaultSchemaTranslations(fs, theme, rootUri),
+    getMetafieldDefinitions: makeGetMetafieldDefinitions(fs),
```
aswamy, line 78: Since `makeGetMetafieldDefinitions` reads the file on the filesystem, wouldn't this actually run every time we run check?

I feel like we need to memoize it too (like we did in startServer)

charlespwd: hmm yeah. I think you'll need to do two things here:
- to support caching in language server, we inject the memoized version of the function as an optional dep
- to support CLI use cases, we do makeGetMetafieldDefinitions(fs) as a fallback

this means:
```ts
interface Dependencies {
  getMetafieldDefinitions?: () => Promise<MetafieldDefinition[]>
}

// ... 

const dependencies: AugmentedDependencies = {
   ...injectedDependencies,
   // ...
   getmetafieldDefinitions = injectedDependencies.getMetafieldDefinitions ?? makeGetMetafieldDefinitions(fs),
}
```

aswamy:
- updated to remove unused `getMetafieldDefinitions` in unit tests
- updated usage of `findThemeRootURI(...)` within `onDidChangeWatchedFiles`
- changed `getMetafieldDefinitions` inside theme-check so it can be injected

packages/theme-language-server-common/src/TypeSystem.ts
```diff
@@ -122,8 +127,24 @@ export class TypeSystem {
...
+    for (let category of FETCHED_METAFIELD_CATEGORIES) {
+      if (!result[category]) continue;
+
+      let metafieldsProperty = result[category].properties?.find(
+        (prop) => prop.name === 'metafields',
+      );
+
+      if (!metafieldsProperty) continue;
+
+      metafieldsProperty.return_type = [{ type: `${category}_metafields`, name: '' }];
```
charlespwd, line 145: Hmm. Thinking about this, are we sure we're not mutating the original object here? Should we clone/replace metafieldsProperty before doing this mutation? 

Remember that `...objectMap` gives you a _shallow_ clone of the object map. Looks like we're mutating the original object map here and this is no bueno. Means that if you'd get metafield data from one theme in another.

packages/theme-language-server-common/src/TypeSystem.ts
```diff
@@ -845,6 +931,101 @@ function settingReturnType(setting: InputSetting): ObjectEntry['return_type'] {
...
+// where x is the type of metafield inside the `value` property. The metafields ending with `x_array`
+// is where the value is an array of type x.
+const customMetafieldTypeEntries = memo((baseMetafieldEntry: ObjectEntry) => {
+  if (!baseMetafieldEntry) return {} as ObjectMap;
+
+  return [
+    ...new Set([...Object.values(METAFIELD_TYPE_TO_TYPE), ...FETCHED_METAFIELD_CATEGORIES]),
+  ].reduce((map, type) => {
+    {
+      const metafieldEntry = JSON.parse(JSON.stringify(baseMetafieldEntry)); // easy deep clone
```
charlespwd, line 991: Oh are we doing it already. I'll just dm you to verify :D

charlespwd (approved):
LGTM, didn't tophat yet :eyes:

packages/theme-language-server-common/src/TypeSystem.ts
aswamy: Yeah sorry there are three things happening to make the whole thing work. We can talk about it more today, but here are the things:
- `customMetafieldTypeEntries` --> we copy the `metafield` type and create one for each category we support. e.g. `product_metafield`, `collection_metafield`, etc.
- `metafieldDefinitionsObjectMap` --> reads the metafield definitions we fetched and converts it into types for auto-completion
- the for-loop you commented above --> we override existing types like `product` and `collection`, so their `metafields` properties don't return as `untyped`, but instead return `product_metafields` and `collection_metafields` respectively

packages/theme-language-server-common/src/TypeSystem.ts
aswamy: We talked about it in person - since this isn't unique to a theme, we should be fine with this change happening across themes within a workspace. Added a warning message in a comment above the block instead.

PR #541, opened 2024-10-25
Add OnSnippetRename theme refactor

charlespwd:

TODO

TODO

## What did you learn?

TODO

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/path.ts
+++ b/packages/theme-check-common/src/path.ts
@@ -24,6 +24,13 @@ export function dirname(uri: UriString): UriString {
   return normalize(Utils.dirname(URI.parse(uri)));
 }
 
+export function basename(uri: UriString, ext?: string): string {
+  return URI.parse(uri)
+    .path.split(/(\\|\/)/g)
+    .pop()!
+    .replace(ext ? new RegExp(`${ext}$`) : '', '');
+}
+
 export function fsPath(uri: UriString): string {
   return URI.parse(uri).fsPath;
 }
--- a/packages/theme-language-server-common/src/server/startServer.ts
+++ b/packages/theme-language-server-common/src/server/startServer.ts
@@ -1,3 +1,4 @@
+import { LiquidTag, NamedTags, NodeTypes } from '@shopify/liquid-html-parser';
 import {
   AugmentedThemeDocset,
   FileTuple,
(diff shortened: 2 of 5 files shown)


UglyJoevIne (approved):

packages/theme-language-server-common/src/renamed/RenameHandler.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * The RenameHandler is responsible for handling workspace/didRenameFiles notifications.
+ *
+ * Stuff we'll handle:
+ * - When a snippet is renamed, then we'll change all the render calls
+ * - When an asset is renamed, then we'll change the asset_url calls
+ * - etc.
+ */
+export class RenameHandler {
```
charlespwd, line 17: Called these handlers because they are not in a request/response cycle. They don't _provide_ anything, they deal with a notification. 

We also already have a RenameProvider and that's for renaming variables/html tags. Different feature similar name.

packages/theme-language-server-common/src/server/startServer.ts
```diff
@@ -387,8 +385,10 @@ export function startServer(
   });
   connection.workspace.onDidRenameFiles((params) => {
     const triggerUris = params.files.map((fileRename) => fileRename.newUri);
+    renameHandler.onDidRenameFiles(params);
```
aswamy, line 388: Stuck while tophatting:
- I have using "Debug node extension"
- When i rename a css or js file, i never see a `workspace/didRenameFiles` in the logs
- I placed a breakpoint here just to make sure, but never hit - it does work for liquid + json files

charlespwd: 23f980e1ae52210546d620adeea2a6c7053f1025

aswamy (approved):
🎩 complete (we already discussed about the limitations on auto-saving after renaming files on Node)

minor comments, but nothing blocking

packages/theme-check-common/src/path.ts
```diff
@@ -24,6 +24,13 @@ export function dirname(uri: UriString): UriString {
   return normalize(Utils.dirname(URI.parse(uri)));
 }
 
+export function basename(uri: UriString, ext?: string): string {
+  return URI.parse(uri)
+    .path.split(/(\\|\/)/g)
```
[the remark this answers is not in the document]
charlespwd: URI.parse().path removes the scheme(and params) from the URI . so you'll never see `vfs:` past that point.

packages/theme-language-server-common/src/renamed/handlers/AssetRenameHandler.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+      },
+      'mock-fs:',
+    );
+    documentManager = new DocumentManager(fs);
+    handler = new RenameHandler(connection, documentManager, makeFileExists(fs));
+  });
+
+  it('returns a needConfirmation: false workspace edit for renaming an asset', async () => {
+    await handler.onDidRenameFiles({
+      files: [
```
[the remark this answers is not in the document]
charlespwd: Tbh, I put that in to satisfy typescript. You'd get an array when you rename folders, but if you did then the condition isSnippet() for both the old and new uri would never be true. 

I'm not 100% sure how else you could get multiple snippet renames in one UI operation.

UglyJoevIne (approved):

PR #542, opened 2024-10-28
Ignore `git:` URI requests and notifications

charlespwd:
The `git:` VFS used by VS Code is the one for URLs shown in git diffs. It does not implement `fs.readDirectory` and makes any kind of complex call to the language server break for that reason.

So this commit makes it so we ignore `git:` requests and notifications.

## Before you deploy

No changeset because VFS hasn't been published yet and it's a VFS "bug."

### DIFF
--- a/packages/theme-language-server-common/src/server/startServer.ts
+++ b/packages/theme-language-server-common/src/server/startServer.ts
@@ -35,10 +35,27 @@ import { GetTranslationsForURI } from '../translations';
 import { Dependencies } from '../types';
 import { debounce } from '../utils';
 import { VERSION } from '../version';
-import { Configuration } from './Configuration';
 import { CachedFileSystem } from './CachedFileSystem';
+import { Configuration } from './Configuration';
 
 const defaultLogger = () => {};
+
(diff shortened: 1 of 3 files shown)


packages/theme-language-server-common/src/server/startServer.ts
```diff
@@ -35,11 +35,28 @@ import { GetTranslationsForURI } from '../translations';
...
 
+/**
+ * The `git:` VFS does not support the `fs.readDirectory` call and makes most things break.
+ * `git` URIs are the ones you'd encounter when doing a git diff in VS Code. They're not
+ * real files, they're just a way to represent changes in a git repository. As such, I don't
+ * think we want to sync those in our document manager or try to offer document links, etc.
+ *
+ * A middleware would be nice but it'd be a bit of a pain to implement.
+ */
+const hasUnsupportedDocument = (params: any) => {
```
graygilmore, line 51: Any chance we can type `params` here? Maybe the language server exports something?

graygilmore (approved):

Issue #540, opened 2024-10-25
Setup the prettier organize import plugin


charlespwd:
**Is your feature request related to a problem? Please describe.**
https://github.com/Shopify/theme-tools/pull/529#discussion_r1815801281

**Describe the solution you'd like**
https://github.com/Shopify/theme-tools/pull/529#discussion_r1815801281

Arman-Luthra:
hey, mind if i grab this? already have it running locally: plugin wired into the prettier config, the organize-on-save vscode setting from the #529 review comment, and the big reformat kept in a separate commit so it's easy to review. pr incoming

PR #551, opened 2024-11-04
Add ValidSchemaName check

miazbikowski:
This PR has 2 "pieces":

1. We needed to parse the `schema` such that we could get the start/end location of the property we're checking, so we used the `visit` logic and moved the `visitor` files (see all the updated paths / re-export)

2. A new Theme Check for the `name` property not going over 25 chars

Resolves https://github.com/Shopify/theme-tools/issues/556

The requested behaviour was: `When a schema name exceeds 25 characters, show an error message`

This allows early feedback ahead of trying to save the file.

The reason I did not use the JSON Schema for this check is that in a follow-up PR, I will be validating the length of `name` when the value is a translation key, which could not have been achieved with JSON schema.

[Note: I tested that this doesn't impact section groups, which have [a name length of 50.](https://shopify.dev/docs/storefronts/themes/architecture/section-groups)]

Follow-ups:
- Validate that name translation key is valid (exists + length)

## Before you deploy

🎩 in theme-tools debug:

image

### DIFF
--- a/.changeset/early-news-check.md
+++ b/.changeset/early-news-check.md
@@ -4,4 +4,4 @@
 '@shopify/theme-check-node': minor
 ---
 
-Add ValidateSchemaName check
+Add `ValidateSchemaName` check
--- a/packages/theme-check-common/src/checks/index.ts
+++ b/packages/theme-check-common/src/checks/index.ts
@@ -26,16 +26,16 @@ import { RequiredLayoutThemeObject } from './required-layout-theme-object';
 import { TranslationKeyExists } from './translation-key-exists';
 import { UnclosedHTMLElement } from './unclosed-html-element';
 import { UndefinedObject } from './undefined-object';
+import { UniqueStaticBlockId } from './unique-static-block-id';
 import { UnknownFilter } from './unknown-filter';
 import { UnusedAssign } from './unused-assign';
 import { ValidContentForArguments } from './valid-content-for-arguments';
 import { ValidHTMLTranslation } from './valid-html-translation';
 import { ValidJSON } from './valid-json';
 import { ValidSchema } from './valid-schema';
+import { ValidSchemaName } from './valid-schema-name';
 import { ValidStaticBlockType } from './valid-static-block-type';
 import { VariableName } from './variable-name';
-import { ValidSchemaName } from './valid-schema-name';
-import { UniqueStaticBlockId } from './unique-static-block-id';
 
 export const allChecks: (LiquidCheckDefinition | JSONCheckDefinition)[] = [
   AppBlockValidTags,
--- a/packages/theme-check-common/src/visitor.spec.ts
+++ b/packages/theme-check-common/src/visitor.spec.ts
@@ -1,8 +1,4 @@
-import {
-  LiquidHtmlNode,
-  LiquidHtmlNodeTypes,
-  SourceCodeType,
-} from './types';
+import { LiquidHtmlNode, LiquidHtmlNodeTypes, SourceCodeType } from './types';
 import { toSourceCode } from './to-source-code';
 import { expect, describe, it } from 'vitest';
 import { Visitor, visit } from './visitor';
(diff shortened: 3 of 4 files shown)


packages/theme-check-common/src/checks/valid-schema-name/index.ts
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        const schema = parseJSON(node.body.value);
+        if (isError(schema) && schema instanceof SyntaxError) return;
+
+        if (!schema.name.startsWith('t:') && schema.name.length > 25) {
+          context.report({
+            message: 'Schema name is too long (max 25 characters)',
+            startIndex: node.blockStartPosition.end,
+            endIndex: node.blockEndPosition.start,
```
albchu, line 34: For the error message indices, I think we should chop it down to only cover the `"name"` value area or the key/value pair. It becomes a bit intense if it highlights the entire schema.

image

miazbikowski: Fair. I'd thought about it, but there's currently no tracking of the location of schema value indices (only the indices of different pieces of liquid, like where the schema ends and begins). I'll have to build that out.

navdeep5: Currently working on a PR that allows us to create a `JSONDocument` within the check -- we can traverse the document and use `offset` + `length` to get valid start and end indices. I will ping you for review once it is ready and you can take a look 😄

.changeset/early-news-check.md
```diff
@@ -0,0 +1,7 @@
+---
+'@shopify/theme-language-server-common': minor
+'@shopify/theme-check-common': minor
+'@shopify/theme-check-node': minor
+---
+
+Add ValidateSchemaName check
```
charlespwd, line 7:
```suggestion
Add the `ValidSchemaName` check
```

packages/theme-check-common/src/checks/index.ts
```diff
@@ -35,6 +34,8 @@ import { ValidJSON } from './valid-json';
 import { ValidSchema } from './valid-schema';
 import { ValidStaticBlockType } from './valid-static-block-type';
 import { VariableName } from './variable-name';
+import { ValidSchemaName } from './valid-schema-name';
+import { UniqueStaticBlockId } from './unique-static-block-id';
```
charlespwd, line 38: Can you use the `Organize imports` vscode command to make sure these are sorted? IIRC the default keyboard shortcut is `ctrl+shift+o`

We have #540 to make this automatic, but never got around doing it.

packages/theme-check-common/src/checks/valid-schema-name/index.ts
```diff
@@ -0,0 +1,82 @@
...
+              // We will be handling the translation length in a follow up
+              if (!name.startsWith('t:') && name.length > MAX_SCHEMA_NAME_LENGTH) {
+                const startIndex = node.blockStartPosition.end + getLiteralLocStart(nameNode.value);
+                const endIndex = node.blockStartPosition.end + getLiteralLocEnd(nameNode.value);
+                context.report({
+                  message: 'Schema name is too long (max 25 characters)',
+                  startIndex: startIndex,
+                  endIndex: endIndex,
+                });
+              }
```
charlespwd, line 56: I was going to recommend to just use the [JSON schema `min/maxLength` rules](https://json-schema.org/understanding-json-schema/reference/string#length). But I could see how we'd want to centralize this logic given we can't really do `t:` with just JSON schemas...

image

What's the rule we want for `t:`? Same thing but query the file? 

```jsonc
{
  "properties": {
    "name": {
      "type": "string",
      "description": "The name attribute determines the block title that's shown in the theme editor.",
      "markdownDescription": "The `name` attribute determines the block title that's shown in the theme editor.\n\n---\n\n[Shopify reference](https://shopify.dev/docs/themes/architecture/blocks/theme-blocks/schema#name)",
      "allOf": [
        {
          "if": { "pattern": "^t:" },
          "else": {
            "maxLength": 25
          }
        }
      ]
    },
  }
}
```

PR #580, opened 2024-11-11
Don't do snippet/asset rename refactors when the client doesn't support it

### LINKED ISSUE #579: We should do a capability check before doing on rename refactors
**Describe the bug**
We already have a ClientCapabilities.hasApplyEditSupport support helper, we should gate the AssetRenameHandler & SnippetRenameHandler to not do anything if the client doesn't support workspace/applyEdit requests (e.g. our codemirror language client!)

charlespwd:

- I do a `workspace/applyEdit` capability gate in `{Asset,Snippet}RenameHandler`.
  - If the client doesn't support it, don't try to do it.
  - The problem is that `workspace/applyEdit` is a `Server->Client` request, and if the client doesn't handle it, the request will hang and never complete.
  - We already do that kind of feature gating in our ExecuteCommand providers, so I'm repeating the pattern here.

I specifically chose to do it in each handler because I don't know what kind of stuff we'll do in rename handlers and I'm not sure I can guarantee that we'll always do `workspace/applyEdit` requests in them.

I made that kind of assumption in execute command and it makes writing different providers/handlers complicated for no good reason.

Fixes #579

## Before you deploy


packages/theme-language-server-common/src/renamed/RenameHandler.ts
```diff
@@ -18,10 +19,14 @@ export class RenameHandler {
...
     connection: Connection,
+    capabilities: ClientCapabilities,
     private documentManager: DocumentManager,
     private fileExists: FileExists,
   ) {
-    this.handlers = [new SnippetRenameHandler(connection), new AssetRenameHandler(connection)];
+    this.handlers = [
+      new SnippetRenameHandler(connection, capabilities),
+      new AssetRenameHandler(connection, capabilities),
+    ];
```
aswamy, line 29: Instead of passing it into the specific handlers, couldn't we just add the `if (!this.capabilities.hasApplyEditSupport) return;` block inside of `onDidRenameFiles` below?

This would reduce the complexity of the individual handlers.

aswamy (approved):
1 suggestion

PR #581, opened 2024-11-11
Do not activate for non-theme-{json,liquid} files

### LINKED ISSUE #576: VS Shopify Liquid Extension > "Request textDocument/hover failed"
While hovering of unsaved JSON files an error is produced: "Request textDocument/hover failed"
Image

Latest verison of the CLI & Extension
Windows 11

See video here: 
https://streamable.com/6l9wqq

charlespwd:

I wrote this thing 3 or 4 years ago without really understanding what I was doing.

The DocumentSelector is what the client uses to tell if it should send requests/notifications to the language server for a given file.

We were sending messages for files that were not supported by us. `package.json` and so on were sent to the language server and we weren't even sure they were theme files.

Fixes #576

### DIFF
--- a/packages/vscode-extension/src/common/constants.ts
+++ b/packages/vscode-extension/src/common/constants.ts
@@ -1,0 +1,14 @@
+import { DocumentSelector } from 'vscode';
+
+/**
+ * The document selectors is the list of documents for which vscode needs to send requests/notifications to the language server.
+ *
+ * We specifically don't want to answer completion requests in package.json files and so on.
+ *
+ * Nor do we want to handle js,css,scss liquid files.
+ */
+export const documentSelectors: DocumentSelector = [
+  { language: 'liquid', pattern: '**/*.liquid' },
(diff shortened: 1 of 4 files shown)


packages/vscode-extension/src/common/constants.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * The document selectors is the list of documents for which vscode needs to send requests/notifications to the language server.
+ *
+ * We specifically don't want to answer completion requests in package.json files and so on.
+ *
+ * Nor do we want to handle js,css,scss liquid files.
+ */
+export const documentSelectors: DocumentSelector = [
+  { language: 'liquid', pattern: '**/*.liquid' },
```
charlespwd, line 11: We unfortunately can't do _negative_ glob patterns here. We rely on `language` to be set properly for the js/css/scss liquid files. 

Looking at packages/vscode-extension/package.json, it looks like I did a mapping for those files to a `language` that _isn't_ liquid. So I think we're good here.

aswamy (approved):

Issue #557, opened 2024-11-04
Verify schema name translation


miazbikowski:
Schema names can also be translated via locale files in the theme, as seen [here](https://github.com/Shopify/dawn/blob/main/sections/announcement-bar.liquid#L169). When the schema name starts with t:, we need to validate the name from the locale files [Already have bespoken locale [verification code](https://github.com/Shopify/theme-tools/blob/main/packages/theme-check-common/src/checks/translation-key-exists/index.ts#L73-L76)]

PR #584, opened 2024-11-11
Verify schema name translation

miazbikowski:
Addresses https://github.com/Shopify/theme-tools/issues/557

We want to verify that the schema name's translation exists, and that its value is less than 25 characters in length. Scope-wise, as aligned with @charlespwd and @albchu, we are doing this check against the default schema translation. 
 
I approached this by doing: if the name starts with `t:`, fetching the default schema translation file and verifying its length.

Video:
https://screenshot.click/12-16-h0bx0-ol5uv.mp4

Follow up:
PR in the `shopify-dev` repo to point out the existence / usage of this check

As a future progressive enhancement, when we can have quick and easy access to all schema locales, it would be great to check all translations for their existence and length.

## Before you deploy

Mia's note: I interpreted "changes of the configuration of a check" to include when you add new functionality to a check. LMK if that's not right and I'll remove the changeset file.

### DIFF
--- a/packages/theme-check-common/src/checks/valid-schema-name/index.spec.ts
+++ b/packages/theme-check-common/src/checks/valid-schema-name/index.spec.ts
@@ -1,5 +1,5 @@
 import { expect, describe, it } from 'vitest';
-import { highlightedOffenses, runLiquidCheck } from '../../test';
+import { highlightedOffenses, runLiquidCheck, check } from '../../test';
 import { ValidSchemaName } from './index';
 
 const DEFAULT_FILE_NAME = 'sections/file.liquid';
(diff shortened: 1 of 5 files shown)


packages/theme-check-common/src/checks/valid-schema-name/index.spec.ts
```diff
@@ -16,10 +16,70 @@ describe('Module: ValidSchemaName', () => {
...
+    expect(offenses[0].message).toEqual(
+      "'t:my_translation_key' does not have a matching entry in 'locales/en.default.schema.json'",
+    );
+  });
+
+  it('reports an offense with schema name translation that exists and is over 25 chars long', async () => {
+    const offenses = await check(
+      {
+        'locales/en.default.schema.json':
+          '{ "my_translation_key": "My translation is tooooooooo long."}',
```
charlespwd, line 69: Can you add a nested property test? `getTranslation` looks fine, but probably better to at least have 1 test.

charlespwd: I also think we probably have a similar util somewhere. Something called `deepGet` or something.

albchu (approved):

PR #585, opened 2024-11-12
Preload theme on boot with progress

charlespwd:

We'll do more stuff like the `{Snippet,Asset}RenameHandler`. We don't want those operations to feel slow the first time it runs.

So as soon as we open a text document in a theme, we'll preload all the files in it.

- Added new helpers in DocumentManager to handle create, delete, rename correctly and prevent requiring preload invalidation.

- Added a `Progress` class to help with progress bars. The LSP stuff feels very heavy and that class makes it a bit cuter/less disruptive in the code. Still gotta do percent calculations (which sucks), but with a couple of comments it ain't so bad.

- Running theme check _after_ the rename handler completes so that we no longer have a race condition that makes it so the file appears to not be there when it is.

Performance wise:
- We'll cache the preload. A theme is only preloaded once per theme root URI.

I'm making this a patch because it's a minor perf improvement and not necessarily something folks should care about externally.

https://github.com/user-attachments/assets/c0a875f7-d98e-4409-9a9d-4c47a5b6d07b

## What did you learn?

- Unit tests in startServer are freaking annoying to write. So much setup. `flushAsync` concerns. Yuk.

## Before you deploy

### DIFF
--- a/packages/theme-language-server-common/src/documents/DocumentManager.ts
+++ b/packages/theme-language-server-common/src/documents/DocumentManager.ts
@@ -1,18 +1,22 @@
 import {
   AbstractFileSystem,
+  memoize,
   path,
   recursiveReadDirectory,
   SourceCode,
   SourceCodeType,
   Theme,
   toSourceCode,
+  UriString,
 } from '@shopify/theme-check-common';
+import { Connection } from 'vscode-languageserver';
 import { TextDocument } from 'vscode-languageserver-textdocument';
-import { URI } from 'vscode-languageserver-types';
+import { ClientCapabilities } from '../ClientCapabilities';
+import { percent as percent, Progress } from '../progress';
 
 export type AugmentedSourceCode<SCT extends SourceCodeType = SourceCodeType> = SourceCode<SCT> & {
   textDocument: TextDocument;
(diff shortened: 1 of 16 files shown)


packages/theme-language-server-common/src/documents/DocumentManager.ts
```diff
@@ -93,15 +118,39 @@ export class DocumentManager {
...
-  }
+  public preload = memoize(
+    async (rootUri: UriString) => {
+      if (!this.fs) throw new Error('Cannot call preload without a FileSystem');
+      const { fs, connection, clientCapabilities } = this;
+      const progress = Progress.create(connection, clientCapabilities, `preload#${rootUri}`);
+
+      progress.start('Initializing Liquid LSP');
+
+      const missingFiles = await recursiveReadDirectory(
```
graygilmore, line 129: Maybe an opportunity to rename this variable that has confused me a couple of times:

```suggestion
      const filesToLoad = await recursiveReadDirectory(
```

graygilmore (approved):
Did a small tophat and things seemed to work! Progress bar worked and renaming was fast.

PR #587, opened 2024-11-12
Add ValidLocalBlocks Check

navdeep5:
## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/checks/valid-local-blocks/index.spec.ts
+++ b/packages/theme-check-common/src/checks/valid-local-blocks/index.spec.ts
@@ -1,0 +1,265 @@
+import { expect, describe, it } from 'vitest';
+import { ValidLocalBlocks } from './index';
+import { check, MockTheme } from '../../test';
+
+it('should report error when local blocks are defined in a block bucket', async () => {
+  const theme: MockTheme = {
+    'blocks/local-blocks.liquid': `
+      {% schema %}
+      {
+        "name": "Block name",
+        "blocks": [
+          {
+            "type": "local_block",
+            "name": "Local Block"
+          }
+        ]
+      }
+      {% endschema %}
+    `,
+  };
+
+  const offenses = await check(theme, [ValidLocalBlocks]);
+  expect(offenses).to.have.length(1);
+  expect(offenses[0].message).to.equal('Local scoped blocks are not supported in theme blocks.');
+});
+
+it('should report local block errors when preset defined local blocks are defined in a block bucket', async () => {
+  const theme: MockTheme = {
+    'blocks/text.liquid': '',
+    'blocks/local-blocks.liquid': `
+      {% schema %}
+      {
+        "name": "Block name",
+        "blocks": [
+          {
+            "type": "text"
+          }
+        ],
+        "presets": [
+          {
+            "name": "Default",
+            "blocks": [
+              {
+                "type": "local_block",
+                "name": "Local Block"
+              }
+            ]
+          }
+        ]
+      }
+      {% endschema %}
+    `,
+  };
+
+  const offenses = await check(theme, [ValidLocalBlocks]);
+  expect(offenses).to.have.length(1);
+  expect(offenses[0].message).to.equal('Local scoped blocks are not supported in theme blocks.');
+});
+
+it('should report errors when sections use theme blocks together with locally scoped blocks in root level', async () => {
+  const theme: MockTheme = {
+    'blocks/text.liquid': '',
+    'sections/local-blocks.liquid': `
+      {% schema %}
+      {
+        "name": "Section name",
+        "blocks": [
+          {
+            "type": "text"
+          },
+          {
+            "type": "local_block",
+            "name": "Local Block"
+          }
+        ]
+      }
+      {% endschema %}
+    `,
+  };
+  const offenses = await check(theme, [ValidLocalBlocks]);
+  expect(offenses).to.have.length(1);
+  expect(offenses[0].message).to.equal(
+    'Sections cannot use theme blocks together with locally scoped blocks.',
+  );
+});
+
+it('should report errors when sections use theme blocks together with locally scoped blocks in presets level', async () => {
+  const theme: MockTheme = {
+    'blocks/text.liquid': '',
+    'sections/local-blocks.liquid': `
+      {% schema %}
+      {
+        "name": "Section name",
+        "blocks": [
+          {
+            "type": "local_block",
+            "name": "Local Block"
+          }
+        ],
+        "presets": [
+          {
+            "name": "Default",
+            "blocks": [
+              {
+                "type": "text"
+              }
+            ]
+          }
+        ]
+      }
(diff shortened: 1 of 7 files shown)


packages/theme-check-common/src/checks/valid-local-blocks/index.spec.ts
```diff
@@ -0,0 +1,265 @@
...
+  };
+  const offenses = await check(theme, [ValidLocalBlocks]);
+  expect(offenses).to.have.length(1);
+
+  const content = theme['sections/local-blocks.liquid'];
+  const errorContent = content.slice(offenses[0].start.index, offenses[0].end.index);
+  expect(errorContent).to.equal('"static_block"');
+});
+
+it('should not report errors when sections mix static and local blocks with @theme', async () => {
```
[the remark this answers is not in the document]
navdeep5: Ohh, thank you for the clarification 😄

packages/theme-check-common/src/checks/valid-local-blocks/index.spec.ts
```diff
@@ -0,0 +1,265 @@
...
+          }
+        ]
+      }
+      {% endschema %}
+    `,
+  };
+
+  const offenses = await check(theme, [ValidLocalBlocks]);
+  expect(offenses).to.have.length(1);
+  expect(offenses[0].message).to.equal('Local scoped blocks are not supported in theme blocks.');
```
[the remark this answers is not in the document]
navdeep5: Originally (and still) we get a `name` error from `ValidSchema`. The idea is to add an additional error message so it is clear that having `name` is a local block pattern.

packages/theme-check-common/src/checks/valid-local-blocks/index.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+    `,
+  };
+  const offenses = await check(theme, [ValidLocalBlocks]);
+  expect(offenses).to.have.length(1);
+  expect(offenses[0].message).to.equal(
+    'Sections cannot use theme blocks together with locally scoped blocks.',
+  );
+});
+
+it('should report errors when sections use theme blocks together with locally scoped blocks in presets level', async () => {
```
miazbikowski, line 87:
```suggestion
it('should report errors when sections use theme blocks together with locally scoped blocks in presets', async () => {
```

super nit

packages/theme-check-common/src/checks/valid-local-blocks/index.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+            ]
+          }
+        ]
+      }
+      {% endschema %}
+    `,
+  };
+
+  const offenses = await check(theme, [ValidLocalBlocks]);
+  console.log(offenses);
```
miazbikowski, line 116:
```suggestion
```

packages/theme-check-common/src/checks/valid-local-blocks/index.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+        `,
+  };
+  const offenses = await check(theme, [ValidLocalBlocks]);
+  expect(offenses).to.have.length(1);
+  expect(offenses[0].message).to.equal(
+    'Sections cannot use static theme blocks together with locally scoped blocks.',
+  );
+});
+
+it('should report errors with correct indicies when sections use static theme blocks together with locally scoped blocks', async () => {
```
miazbikowski, line 158:
```suggestion
it('should report errors with correct indices when sections use static theme blocks together with locally scoped blocks', async () => {
```

nit

miazbikowski: nit: maybe a bit overkill to have a test just for the indices, could maybe merge this with the previous test?

packages/theme-check-common/src/checks/valid-local-blocks/index.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+  };
+  const offenses = await check(theme, [ValidLocalBlocks]);
+  expect(offenses).to.have.length(1);
+
+  const content = theme['sections/local-blocks.liquid'];
+  const errorContent = content.slice(offenses[0].start.index, offenses[0].end.index);
+  expect(errorContent).to.equal('"static_block"');
+});
+
+it('should report errors when static theme blocks have a name property', async () => {
```
miazbikowski, line 194: If this error is coming from the JSON schema, I wouldn't add this test just because if ever we change something in the `theme-liquid-docs` repo, this test could break.

packages/theme-check-common/src/checks/valid-local-blocks/index.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+          }
+        ]
+      }
+      {% endschema %}
+    `,
+  };
+  const offenses = await check(theme, [ValidLocalBlocks]);
+  expect(offenses).to.have.length(1);
+  expect(offenses[0].message).to.equal('Static theme blocks cannot have a name property.');
+});
```
miazbikowski, line 226: Fun fact: we recently decided to support preset blocks as a hash:

```
presets: [{
  "name": "My preset",
   "blocks": {
     "my_block_id": { "type": "text" }
   }
}]
```

That means we should ensure the check works in the case of preset blocks as hash, too.

miazbikowski (changes requested):
I'll re-review when we have test coverage for tests that include preset theme blocks, where we also need to have a version where those preset blocks are a hash version (don't hesitate to ping me if you have any questions about how to go from array to the hash version)

packages/theme-check-common/src/checks/valid-local-blocks/index.spec.ts
navdeep5: These tests are all testing for `ValidLocalBlocks` specific checks.

packages/theme-check-common/src/checks/valid-local-blocks/index.spec.ts
navdeep5: Oh my, more complexity 😆

albchu (dismissed):
lgtm

packages/theme-check-common/src/checks/valid-local-blocks/index.ts
```diff
@@ -0,0 +1,89 @@
...
+          }
+
+          if (hasLocalBlocks && hasThemeBlocks) {
+            themeBlockLocations.forEach(
+              reportError(
+                'Sections cannot use theme blocks together with locally scoped blocks.',
+                context,
+                node,
+              ),
+            );
```
albchu, line 77: Nit: it feels a bit noisy to blast an error on all preset theme blocks when a local block exists in the root level blocks. It would be nicer if we could limit the errors to only the root level blocks

albchu (commented):
image
Found a bug: we should not see an error on `@app` blocks for this check.

albchu (approved):
lgtm, thanks Nav!

PR #588, opened 2024-11-13
Fix hardcoded `file:` scheme in resolveConfig

charlespwd:
I think it should have been `'file://' + absolutePath` and is probably the cause of #586.

See the comments in 586 for some details of the investigation.

Can't confirm this fixes it because I couldn't figure out a way to reproduce the issue reported. But given that I know it should have been `file://`, let's see if that fixes it.

might solve #586 ?

## What did you learn?

- I only have a handwavy understanding of how URIs for files work
- I learned a bit of things by [reading the URI spec](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1):
    - A URI has 4 allowed syntaxes (rewritten in ohm syntax)
    
```ohmjs
Uri {
  Uri = scheme ":" hierPart ("?" query)? ("#" fragment)?
  scheme = letter (letter | digit | "-" | "+" | ".")*
  hierPart = 
    | "//" authority pathAbEmpty -- differentRule
    | pathAbsolute 
    | pathRootless
    | pathEmpty
    
  pathAbEmpty = ("/" pathSegment)*
  pathAbsolute = ("/" pathSegment)+
  pathRootless = pathSegment pathAbEmpty
  pathEmpty = "/"
  pathSegment = anyExceptStar<"/">
  authority = userInfo? host port? 
  userInfo = "..." "@"
  port = ":" digit+
  
  // host can be empty!
  host = (alnum | ".")*
 
  query = any* 
  fragment = any*
  
  anyExceptStar<lit> = (~ lit any)*
}
```

Key thing to realize is that you can only start with _one_ or _three_ slashes after the scheme. If you only have 2, then we have a problem.

The `authority` can be the empty string. Which is why `file:///` works.

    
image

          

## Before you deploy

### DIFF
--- a/.changeset/modern-squids-wait.md
+++ b/.changeset/modern-squids-wait.md
@@ -1,0 +1,6 @@
+---
+'@shopify/theme-check-common': patch
+'@shopify/theme-check-node': patch
+---
+
+Fix weird root URI loading bug
--- a/packages/theme-check-common/src/path.ts
+++ b/packages/theme-check-common/src/path.ts
@@ -1,36 +1,40 @@
 import { RelativePath, UriString } from './types';
 import { URI, Utils } from 'vscode-uri';
 
-export function relative(uri: UriString, rootUri: UriString): RelativePath {
-  return uri
+export { URI, Utils };
+
+export function relative(uri: UriString | URI, rootUri: UriString): RelativePath {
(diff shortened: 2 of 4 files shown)


charlespwd:
Failing unit tests... we're on to something 🤔

PR #589, opened 2024-11-13
Gate the "preload" functionality behind `themeCheck.preloadOnBoot` initialization option

charlespwd:

We probably don't want to preload all the files just yet.

We can now start the `CodeMirrorLanguageClient` with the following to disable full theme preload on theme file open.

```
const client = new CodeMirrorLanguageClient(worker, {
  initializationOptions: {
    'themeCheck.preloadOnBoot': false,
  }
});
```

## Before you deploy

### DIFF
--- a/packages/theme-language-server-common/src/renamed/handlers/AssetRenameHandler.ts
+++ b/packages/theme-language-server-common/src/renamed/handlers/AssetRenameHandler.ts
@@ -1,5 +1,5 @@
 import { LiquidVariable, NodeTypes } from '@shopify/liquid-html-parser';
-import { SourceCodeType } from '@shopify/theme-check-common';
+import { path, SourceCodeType, visit } from '@shopify/theme-check-common';
 import { Connection } from 'vscode-languageserver';
 import {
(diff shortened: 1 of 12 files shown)


packages/theme-language-server-common/src/renamed/handlers/AssetRenameHandler.ts
```diff
@@ -27,18 +26,29 @@ import { BaseRenameHandler } from '../BaseRenameHandler';
...
 
-    const liquidSourceCodes = theme.filter(isLiquidSourceCode);
     const relevantRenames = params.files.filter(
       (file) => isAsset(file.oldUri) && isAsset(file.newUri),
     );
 
+    // Only preload if you have something to do
+    if (relevantRenames.length === 0) return;
+    const rootUri = await this.findThemeRootURI(path.dirname(params.files[0].oldUri));
+    await this.documentManager.preload(rootUri);
```
charlespwd, line 48: Also further gated the preload behind the hasApplyEdit support here. We'll only preload if we know we must preload.

Issue #463, opened 2024-07-25
Theme Check rule to warn against using hard coded block id


bahar-p:
We've updated the [docs](https://shopify.dev/docs/api/liquid/objects/block#block-id) to warn against relying on hard coded block ids. In additional to that it'd still be very valuable to have a theme check rule to warn against this usage as people author liquid code. See this [document](https://docs.google.com/document/d/1KOi0U-4HKfuNtpaKIN_Phfo_S8_xxjlD8OVSlV0m8N8/edit#heading=h.lgeabch6zmdt) for examples of this.

PR #595, opened 2024-11-14
Warn against block.id usage

miazbikowski:
Part of https://github.com/Shopify/theme-tools/issues/463

We want to warn developers against the use of `block.id` in their liquid code in cases where the ID's instability could cause problems. This PR addresses usage in `if`, `elseif`, `unless` and `case` blocks. (I've got some comments in the code to demonstrate the cases)

Note: did not end up doing `assign` as a variable can have valid use cases (like those in my code comments)

## Follow ups
Dev docs [PR](https://github.com/Shopify/shopify-dev/pull/50646)

## What did you learn?
`console.log` will break your code and not tell you why. This is because of the language server.

## Before you deploy


packages/theme-check-common/src/checks/block-id-usage/index.ts
```diff
@@ -0,0 +1,68 @@
...
+      async VariableLookup(node, ancestors) {
+        const parentNode = ancestors.at(-1);
+        if (parentNode?.type === NodeTypes.LiquidTag && parentNode.name === 'case') {
+          if (isUsingBlockId(node)) {
+            reportWarning(context, node.position);
+          }
+        }
+      }
+    };
+  },
```
charlespwd, line 49: So tidy. So simple. So beautiful 🥹

.changeset/seven-tips-burn.md
```diff
@@ -0,0 +1,7 @@
+---
+'@shopify/liquid-html-parser': minor
```
charlespwd, line 2: I think this warrant a _separate_ changeset that is a `patch` change. Not really a new feature, so this is a patch.

charlespwd (approved):

navdeep5 (approved):
After discussion with Mia and CP, it looks like invalid liquid just returns `false` and in these cases, we will not see any squiggly lines (which I believe is okay since it is not a check-related error)

PR #597, opened 2024-11-15
Fetch metafield definitions on start-up

aswamy:

Part of https://github.com/Shopify/theme-tools/issues/502
- Adding the ability to fetch metafield definitions on VSCode startup
- Does not check the version of the CLI
- Timeouts if the CLI does not complete the action

## Tophat

Unfortunately, the CLI does not contain the metafield definitions pull command yet. To tophat the change override the `path` variable to use a locally built version of CLI of the main branch.

E.g.

```
const  path = '<path-to-cli-repo>/packages/cli/bin/dev.js';
```

UPDATE: CLI version 3.73 has `shopify theme metafields pull` command so you can just update your CLI and it should work if you're logged in.

- Should we be adding `.shopify/metafields.json` as part of gitignore for themes?

## Before you deploy

### DIFF
--- a/.changeset/plenty-flowers-eat.md
+++ b/.changeset/plenty-flowers-eat.md
@@ -1,0 +1,6 @@
+---
+'@shopify/theme-language-server-common': minor
+'@shopify/theme-language-server-node': minor
+---
+
+Fetch metafield definitions on start-up using CLI
--- a/packages/theme-language-server-common/src/server/startServer.ts
+++ b/packages/theme-language-server-common/src/server/startServer.ts
@@ -18,6 +18,7 @@ import {
   InitializeResult,
   ShowDocumentRequest,
   TextDocumentSyncKind,
+  WorkspaceFolder,
 } from 'vscode-languageserver';
 import { ClientCapabilities } from '../ClientCapabilities';
 import { CodeActionKinds, CodeActionsProvider } from '../codeActions';
@@ -79,6 +80,7 @@ export function startServer(
     log = defaultLogger,
     jsonValidationSet,
     themeDocset: remoteThemeDocset,
+    fetchMetafieldDefinitionsForURI,
   }: Dependencies,
 ) {
   const fs = new CachedFileSystem(injectedFs);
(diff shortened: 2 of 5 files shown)


packages/vscode-extension/src/node/metafieldDefinitions.ts
```diff
@@ -0,0 +1,36 @@
+import * as child_process from 'child_process';
+import { promisify } from 'node:util';
+
+const exec = promisify(child_process.exec);
+
+const isWin = process.platform === 'win32';
+
+export async function fetchMetafieldDefinitions() {
+  const path = getShopifyCliPath();
```
aswamy, line 9: The current CLI does not have the ability to pull metafields, so tophat this build the CLI PR [here](https://github.com/Shopify/cli/pull/4782) and hardcode this value to the path.

E.g.

`const path = '<where-ever-you-cloned-it>/cli/packages/cli/bin/dev.js'`

.changeset/plenty-flowers-eat.md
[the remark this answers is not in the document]
graygilmore: 👍🏻 

I think we should do two things:

- Add `.shopify` to `.gitignore` in our 1P themes
- Update the CLI to warn if there is a `.shopify` folder that isn't ignored during `theme dev`

What do you think?

graygilmore (approved):

jeffreyguenther:
>Update the CLI to warn if there is a .shopify folder that isn't ignored during theme dev

Yes, pls. A nice error is a great way to be hinted at how things work!

packages/theme-language-server-common/src/server/startServer.ts
```diff
@@ -321,6 +339,16 @@ export function startServer(
...
     });
+
+    connection.workspace.getWorkspaceFolders().then(async (folders) => {
+      if (!folders) return;
+
+      fetchMetafieldDefinitionsForWorkspaceFolders(folders);
+    });
+
+    connection.workspace.onDidChangeWorkspaceFolders(async (params) => {
+      fetchMetafieldDefinitionsForWorkspaceFolders(params.added);
```
charlespwd, line 350: What kind of event triggers this? Does this happen a lot a lot?

charlespwd: Ahh ok so not that often! Alrighty :)

charlespwd (approved):
LGTM! 🙏

PR #604, opened 2024-11-19
Check preset blocks_order

miazbikowski:
Note: I'm expecting a conflict with Nav's [PR](https://github.com/Shopify/theme-tools/pull/617/files/1083b2bc9892b1fc9f725955ca68b3538b228f65..9435f2661f3ccc751ac0e6835902e0a2e2497cc0) regarding changes to the types. I've volunteered to be the one who deals with the conflicts.

---- 
Dependent on https://github.com/Shopify/theme-liquid-docs/pull/849 

Solves https://github.com/Shopify/theme-tools/issues/590

1. Checks that block_order is present when it should be (if you have blocks and they aren't all static)

2. When presets have blocks as a hash, they must also be included in the `block_order` array... Unless they are a static block (a block with `static: true`), then they have to NOT be included in the `block_order`.

Note: this check doesn't validate the existence of `block_order`, it assumes it's there to run this check.

To test:
```
"presets": [
    {
      "name": "Slideshow",
      "settings": {
        "title": "Slideshow"
      },
      "blocks": [
        {
          "type": "slide",
          "blocks": [
            {
              "type": "test"
            }
          ]
        },
        {
          "type": "slide"
        }
      ]
    },
    {
      "name": "Preset test 1",
      "blocks": {
        "block1": {
          "type": "test"
        },
        "block2": {
          "type": "slide",
          "static": true
        }
      },
      "block_order": ["block2"]
    },
    {
      "name": "Preset test 1",
      "blocks": {
        "block1": {
          "type": "test",
          "blocks": {
            "nested_block2": {
              "type": "slide"
            }
          },
          "block_order": []
        }
      },
      "block_order": ["block1"]
    },
    {
      "name": "Block order is missing",
      "blocks": {
        "block1": {
          "type": "test"
        }
      }
    },
    {
      "name": "Block_order is correctl missing",
      "blocks": {
        "block1": {
          "type": "test",
          "static": true
        }
      }
    },
    {
      "name": "block_order for nested block is missing",
      "blocks": {
        "block1": {
          "type": "test",
          "static": true,
          "blocks": {
            "nested_block2": {
              "type": "slide"
            }
          }
        }
      }
    }
  ],
```

Note that I updated a lot of the types to better represent the possible cases in presets.

## What did you learn?
To use CP's new building blocks :)

## Before you deploy

### DIFF
--- a/.changeset/smooth-cycles-hear.md
+++ b/.changeset/smooth-cycles-hear.md
@@ -1,0 +1,7 @@
+---
+'@shopify/theme-check-common': minor
+'@shopify/theme-check-node': minor
+'@shopify/theme-check-vscode': minor
+---
+
+Add `SchemaPresetsBlockOrder` check
--- a/packages/theme-check-common/src/checks/schema-presets-block-order/index.ts
+++ b/packages/theme-check-common/src/checks/schema-presets-block-order/index.ts
@@ -1,0 +1,149 @@
+import { ArrayNode } from 'json-to-ast';
+import { getLocEnd, getLocStart, nodeAtPath } from '../../json';
+import { basename } from '../../path';
+import { isBlock, isSection } from '../../to-schema';
+import { JSONNode, LiquidCheckDefinition, Severity, SourceCodeType } from '../../types';
+import { Preset } from '../../types/schemas/preset';
+import { ThemeBlock } from '../../types/schemas/theme-block';
+import { Context } from '../../types';
+
+export const SchemaPresetsBlockOrder: LiquidCheckDefinition = {
+  meta: {
+    code: 'SchemaPresetsBlockOrder',
+    name: 'Gives recommendations and validations for block_order',
+    docs: {
+      description:
+        'Gives recommendations and validations for block_order for preset blocks as a hash.',
+      recommended: true,
+      url: 'https://shopify.dev/docs/storefronts/themes/tools/theme-check/checks/schema-presets-block-order',
+    },
+    type: SourceCodeType.LiquidHtml,
+    severity: Severity.WARNING,
+    schema: {},
+    targets: [],
+  },
+
+  create(context) {
+    function getSchema() {
+      const name = basename(context.file.uri, '.liquid');
+      switch (true) {
+        case isBlock(context.file.uri):
+          return context.getBlockSchema?.(name);
+        case isSection(context.file.uri):
+          return context.getSectionSchema?.(name);
+        default:
+          return undefined;
+      }
+    }
+
+    return {
+      async LiquidRawTag(node) {
+        if (node.name !== 'schema' || node.body.kind !== 'json') {
+          return;
+        }
+
+        const offset = node.blockStartPosition.end;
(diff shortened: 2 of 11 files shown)


miazbikowski:
@navdeep5 just warned me that copy+pasting the same schema code from my tests didn't work in the f5 vs code instance so I'm definitely going to look into that shortly

navdeep5:
> @navdeep5 just warned me that copy+pasting the same schema code from my tests didn't work in the f5 vs code instance (which is weird, since, tests pass with the same schema) so I'm definitely going to look into that shortly

Yeah, it early returns since `validSchema` is false and contains an error -- so the check does not run. It thinks `block_order` is not a valid property within nested blocks.
image

albchu (commented):
image
Should this message be an error instead of a warning? Our BE validation should hard reject this case.

miazbikowski:
> image Should this message be an error instead of a warning? Our BE validation should hard reject this case.

@albchu I had this reflection as well, the problem is that this check also checks that blocks be added to block_order but not having a block in the `block_order` is not an error, it's just not ideal. Since I have to set the severity for the check, and not for individual cases, I put warning for now. We could consider splitting the checks.

packages/theme-check-common/src/checks/schema-presets-block-order/index.ts
```diff
@@ -0,0 +1,150 @@
...
+  context: Context<SourceCodeType.LiquidHtml>,
+  offset: number,
+  ast: JSONNode,
+  ast_path: string[],
+) {
+  if ('blocks' in node && typeof node.blocks == 'object' && node.blocks) {
+    const blockOrder = 'block_order' in node ? node.block_order : undefined;
+    // check 1: is block_order missing?
+    if (!blockOrder) {
+      if (shouldHaveBlockOrder(node)) {
```
navdeep5, line 75: Super well done! Only nit is would it be easier on the eyes to place the warning for missing `block_order` on merely the hash key  as opposed to everywhere?
image

navdeep5 (approved):

.changeset/smooth-cycles-hear.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/theme-check-common': minor
+'@shopify/theme-check-node': minor
```
charlespwd, line 3: I know we technically didn't modify it, but I like to see those added to the vscode extension readme when they are added. 

Mind adding `theme-check-vscode: minor` in there?

albchu (approved):
Changes look good but I found that the following use-case curiously does not raise any offences. 

```json
{
  "name": "Test section",
  "presets": [
    {
      "name": "Preset 1",
      "blocks": {
        "block-1": {
          "type": "text"
        },
      },
      "block_order": ["block-1", "block-2"]
    }
  ]
}
```

I think it would look nice if we could highlight `"block-2"` with a warning that it is not defined within `"blocks"`.

This PR is already quite the set of changes though so it would be more than reasonable to hit that separately.

Issue #605, opened 2024-11-19
ValidBlock throwing false positives


charlespwd:
**Describe the bug**
Was testing the cli bump and found those false positives from dawn:

![Image](https://github.com/user-attachments/assets/8ce0e3e8-3c93-44c3-b570-131ec069c5bf)

Issue #608, opened 2024-11-19
ValidLocalBlock throwing false positives


charlespwd:
**Describe the bug**

![Image](https://github.com/user-attachments/assets/8b8f3bac-b933-4c5c-bc51-242a606dda1a)

Shouldn't be throwing on `blocks[].settings[].type` 🤔

PR #617, opened 2024-11-25
Refactor `ValidBlockTarget` and `ValidLocalBlocks` & Introduce Cross-File Nested Block Targeting Check

navdeep5:

Resolves https://github.com/Shopify/theme-tools/issues/605 and https://github.com/Shopify/theme-tools/issues/608

This PR tackles two things:
1) Refactoring https://github.com/Shopify/theme-tools/pull/561 and https://github.com/Shopify/theme-tools/pull/587
 - There was a bug in Dawn where a lot of unnecessary things were getting flagged as incorrect. In order to fix it using the old parsing code, I had to hard code ancestor numbers which, although logically made sense, were, frankly, ugly and could be easily broken should schema definitions change. Thus, these checks have been refactored using new parsing code that is hopefully cleaner and easier to follow.
2) Introduces cross-file checks for nested blocks
- Initial understanding was that all preset level blocks (including blocks nested in other blocks) merely had to adhere to root-level blocks. However, we learned that nested blocks instead have to adhere to their parent block which required cross-file checking. The new parsing code made this a lot more feasible so I added it into this PR.

With the checks being refactored and `recommended` being set back to `true`, I believe the checks will need to be re-deployed.

## What did you learn?

There was a lot to learn in this process:
1) The new parsing code. CP did an unbelievable job here making everything so much easier to parse and write!
2) How to correctly check for block targeting. Once again, initial belief was that everything in presets simply had to adhere to the root-level blocks, but, that is in fact false -- nested blocks need to adhere to the root-level blocks of their **parent block**.
3) Adding in the cross-file check itself -- something new to me.

## Before you deploy


packages/theme-check-common/src/checks/valid-local-blocks/index.ts
navdeep5, line 1: The logic here is the same as before, just switched over to the new parsing code.

packages/theme-check-common/src/checks/valid-block-target/index.spec.ts
```diff
@@ -479,6 +500,351 @@ describe('Module: ValidBlockTarget', () => {
...
+                }
+              ]
+            }
+            {% endschema %}
+          `,
+        };
+
+        const offenses = await check(theme, [ValidBlockTarget]);
+        expect(offenses).to.have.length(1);
+        expect(offenses[0].message).to.equal(
```
albchu, line 736: Given the cross file added complexity, can you also verify the offence `uri` surfaces this error for the right theme file?

packages/theme-check-common/src/checks/valid-block-target/index.spec.ts
```diff
@@ -479,6 +500,351 @@ describe('Module: ValidBlockTarget', () => {
...
+                }
+              ]
+            }
+            {% endschema %}
+          `,
+        };
+
+        const offenses = await check(theme, [ValidBlockTarget]);
+        expect(offenses).to.be.empty;
+      });
```
albchu, line 847: Can you add a unit test to verify that this check wont crash when we have a cyclical theme blocks relationship?

Something like this:
blockA has a preset with blockB that nests blockC
blockC has a preset that uses blockA and nests blockB

The thing we're testing is that `await check` does not crash. It should resolve in a reasonable amount of time and also with no offences.

Claude recommended this assert structure 
```typescript
test('resolves within timeout', async () => {
  await expect(async () => {
        const offenses = await check(theme, [ValidBlockTarget]);
        expect(offenses).to.be.empty;
  }).toMatchTimings({
    // Will fail if execution takes longer than 500ms
    maxMs: 500
  })
})
```

albchu (commented):
Some unit test requests. The reset of the code otherwise looks good. Will do a tophat tomorrow morning.

packages/theme-check-common/src/checks/valid-block-target/index.spec.ts
navdeep5: Good idea!

packages/theme-check-common/src/checks/valid-block-target/index.spec.ts
navdeep5: Sure sure, I can look into this!

miazbikowski (approved):
🎩 ed and reviewed - thanks Nav!!

packages/theme-check-common/src/checks/valid-local-blocks/index.spec.ts
```diff
@@ -329,3 +367,127 @@ describe('ValidLocalBlocks with hash-style presets', () => {
...
+        }
+        {% endschema %}
+      `,
+    };
+
+    const offenses = await check(theme, [ValidLocalBlocks]);
+    expect(offenses).to.be.empty;
+  });
+
+  it('should not report errors on older themes that define local blocks at the root level and use them in presets', async () => {
```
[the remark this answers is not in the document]
navdeep5: Good point!

packages/theme-check-common/src/checks/valid-local-blocks/index.ts
```diff
@@ -24,64 +33,171 @@ export const ValidLocalBlocks: LiquidCheckDefinition = {
...
+        default:
+          return undefined;
+      }
+    }
 
-    return {
-      async LiquidRawTag(node) {
-        if (node.name !== 'schema' || node.body.kind !== 'json') {
-          return;
+    function getBlocks(validSchema: ThemeBlock.Schema | Section.Schema): {
```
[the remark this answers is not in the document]
navdeep5: Sure thing, I initially left it since it would just be two functions but, you're right, doesn't hurt to make it a little cleaner!

Issue #602, opened 2024-11-18
Validate preset settings


charlespwd:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Checklist** 

- [ ] I have checked and made sure that the proposal adheres to this plugin's [principles](/docs/prettier-plugin-liquid/principles/index.md)

**Additional context**

PR #620, opened 2024-11-27
[Theme Checks] Added in a check for theme block settings validation

AribaRajput:

First part of https://github.com/Shopify/theme-tools/issues/602

Adding in a theme check for checking theme blocks for valid preset and default setting keys. The preset and default setting keys (and the presets of any theme block preset settings) must match the settings keys configured in either the preset settings or the nested theme block settings. 

## What did you learn?

A lot about presets, sections and theme blocks and how they are configured. That was the trickiest part of all of this (managing the types, knowing how to grab nested information, etc). 

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/checks/valid-preset-and-default-settings/index.ts
+++ b/packages/theme-check-common/src/checks/valid-preset-and-default-settings/index.ts
@@ -1,19 +1,19 @@
 import { isSection, isBlock } from '../../to-schema';
 import { basename } from '../../path';
 import {
-  ArrayNode,
+  JSONNode,
   LiquidCheckDefinition,
   LiteralNode,
-  ObjectNode,
-  Preset,
-  PropertyNode,
-  Setting,
   Severity,
   SourceCodeType,
-  ValueNode,
 } from '../../types';
 import { nodeAtPath } from '../../json';
-import { BlockNodeWithPath, getBlocks } from '../valid-block-target/block-utils';
+import {
+  BlockNodeWithPath,
+  getBlocks,
+  validateBlockFileExistence,
+} from '../valid-block-target/block-utils';
+import { Section } from '../../types/schemas/section';
 
 export const ValidPresetAndDefaultSettings: LiquidCheckDefinition = {
   meta: {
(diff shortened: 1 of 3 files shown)


packages/theme-check-common/src/checks/valid-block-preset-settings/index.ts
```diff
@@ -0,0 +1,115 @@
...
+import { Preset } from '../../types/schemas/preset';
+
+export const ValidBlockPresetSettings: LiquidCheckDefinition = {
+  meta: {
+    code: 'ValidBlockPresetSettings',
+    name: 'Reports invalid preset settings for a theme block',
+    docs: {
+      description: 'Reports invalid preset settings for a theme block',
+      recommended: true,
+      url: 'https://shopify.dev/docs/storefronts/themes/tools/theme-check/checks/preset-key-exists',
```
charlespwd, line 13:
```suggestion
      url: 'https://shopify.dev/docs/storefronts/themes/tools/theme-check/checks/valid-block-preset-settings,
```

packages/theme-check-common/src/checks/valid-block-preset-settings/index.ts
```diff
@@ -0,0 +1,115 @@
...
+import { basename } from '../../path';
+import { LiquidCheckDefinition, Severity, SourceCodeType, ThemeBlock } from '../../types';
+import { Preset } from '../../types/schemas/preset';
+
+export const ValidBlockPresetSettings: LiquidCheckDefinition = {
+  meta: {
+    code: 'ValidBlockPresetSettings',
+    name: 'Reports invalid preset settings for a theme block',
+    docs: {
+      description: 'Reports invalid preset settings for a theme block',
```
charlespwd, line 11: Is it only theme blocks or just blocks in general?

packages/theme-check-common/src/checks/valid-block-preset-settings/index.ts
```diff
@@ -0,0 +1,115 @@
...
+        const validSchema = schema.validSchema;
+        const settingsKeys = getInlineSettingsTypesandKeys(validSchema.settings);
+        const presetSettingsKeys = getPresetSettingsKeys(validSchema.presets ?? []);
+
+        for (const key of presetSettingsKeys) {
+          if (!settingsKeys.some((setting) => setting.id === key)) {
+            context.report({
+              message: `Preset setting "${key}" does not exist in the block's settings`,
+              startIndex: 0,
+              endIndex: 0,
```
charlespwd, line 86: We need to highlight the node. I don't think it's ok to have 0:0 for this check. 

`schema` has a `ast` property you can use to find the node. We also have the `nodeAtPath` util that can give you the AST node for a given "JSONPath"

```ts
const schema = await getSchema();
if (!schema) return;
if (isError(schema.validSchema) || isError(schema.ast)) return;

// find out that ["presets", 0, "settings", "unknown-id"] is invalid:
const settingsNode = nodeAtPath(schema.ast, ["presets", 0, "settings"]) as ObjectNode;
if (!settingsNode) {} // something went wrong

const unknownIdNode = settingsNode.
```

charlespwd (changes requested):
Position information can't be ignored for this check. We'll probably need to rewrite it and iterate over each blocks separately rather than push all of them in the same bucket.

.changeset/chilled-bugs-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/theme-check-common': minor
+---
+
+Added theme check for preset settings validation on theme blocks
```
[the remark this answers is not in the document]
AribaRajput: Will keep in mind going forward!

packages/theme-check-common/src/checks/valid-block-preset-settings/index.ts
AribaRajput: Or, whatever block is in the presets settings.

packages/theme-check-common/src/checks/valid-preset-and-default-settings/index.ts
```diff
@@ -0,0 +1,272 @@
...
+      blocks: Record<string, any>[],
+      path: string[] = ['default', 'blocks'],
+    ): { key: string; start: any; end: any }[] => {
+      let settingIds: { key: string; start: any; end: any }[] = [];
+
+      for (const [blockIndex, block] of blocks.entries()) {
+        const currentPath = [...path, blockIndex.toString()];
+
+        for (const [settingId, _] of Object.entries(block.settings ?? {})) {
+          const node = nodeAtPath(ast, [...currentPath, 'settings']);
```
AribaRajput, line 57: Need to nest this one level deeper - the node is highlighting all of settings rather than the specific id key

PR #634, opened 2024-12-02
Update filter param tests

graygilmore:

This is a follow up to https://github.com/Shopify/theme-tools/pull/522 where I'm just updating the filter completion tests that I added to be a little more robust.

We were previously asserting only against the fact that the text edits were what we expected but we can have a lot more confidence if we also test that the application of the text edit is what we expect.

This PR hooks into TextDocument and lets us use `applyEdits` to test that the modified string looks as we expect it to.

## Before you deploy

### DIFF
--- a/.changeset/wild-hornets-crash.md
+++ b/.changeset/wild-hornets-crash.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-language-server-common': patch
+---
+
+Use CURSOR variable instead of hardcoded string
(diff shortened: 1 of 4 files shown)


jamesmengo (approved):
🔥 This is an awesome change!

charlespwd (approved):
🤩 :)

EvilGenius13 (approved):
Awesome!

PR #636, opened 2024-12-03
Theme Check - Update theme-check-common to include doc tag

jamesmengo:
## Before you deploy

- [ ] This PR includes a new checks or changes the configuration of a check
  - [ ] I included a minor bump `changeset`
  - [ ] It's in the `allChecks` array in `src/checks/index.ts`
  - [ ] I ran `yarn build` and committed the updated configuration files
    
    
    
    - [ ] If applicable, I've updated the `theme-app-extension.yml` config

- [ ] I included a minor bump `changeset`
- [ ] My feature is backward compatible

- [ ] I included a patch bump `changeset`

PR #621, opened 2024-11-27
Add liquidDoc tag to liquidHTMLParser

jamesmengo:
--wip-- [skip ci]

--wip-- [skip ci]

## Before you deploy

- [ ] This PR includes a new checks or changes the configuration of a check
  - [ ] I included a minor bump `changeset`
  - [ ] It's in the `allChecks` array in `src/checks/index.ts`
  - [ ] I ran `yarn build` and committed the updated configuration files
    
    
    
    - [ ] If applicable, I've updated the `theme-app-extension.yml` config

- [ ] I included a minor bump `changeset`
- [ ] My feature is backward compatible

- [ ] I included a patch bump `changeset`

PR #646, opened 2024-12-03
Implement fallbackNode for LiquidDoc

jamesmengo:
## Before you deploy


jamesmengo:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/646?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#646** <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/646?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/646?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#638** <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/638?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#621** <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/621?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#636](https://github.com/Shopify/theme-tools/pull/636) <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/636?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

packages/liquid-html-parser/src/stage-2-ast.ts
```diff
@@ -1103,7 +1109,7 @@ export function cstToAst(
 }
 
 function buildAst(
-  cst: LiquidHtmlCST | LiquidCST | ConcreteAttributeNode[],
+  cst: LiquidHtmlCST | LiquidCST | LiquidDocCST | ConcreteAttributeNode[],
```
jamesmengo, line 1112: Should I just be adding this to `LiquidHtmlCST` ? My plan is to allow `LiquidDocCST` to evolve over time. I wanted to keep them separate as much as possible since `LiquidDoc` is kind of its own thing

cc: @charlespwd

packages/liquid-html-parser/grammar/liquid-html.ohm
```diff
@@ -389,7 +389,17 @@ LiquidStatement <: Liquid {
...
+  LiquidDocNode =
+    | paramNode
+    | fallbackNode
+
+  fallbackNode = "@" anyExceptStar<newline>
+  paramNode = "@param" space* (paramType)? space* (paramName)? space* "-"? paramDescription?
+  paramType = "{" anyExceptStar<"}"> "}"
+  paramName = identifierCharacter+
+  paramDescription = (~newline identifierCharacter | space | punctuation)+
+  punctuation = singleQuote | doubleQuote | "." | "`"
```
jamesmengo, line 402: This will likely continue to evolve over time as we discover more edge cases. I don't expect this to be perfect, though I want to fix anything smelly now rather than later.

This is one of those scenarios where it's hard to know where I might shoot myself in the foot until the bullet is already in my shoe imo

packages/liquid-html-parser/grammar/liquid-html.ohm
```diff
@@ -389,7 +389,17 @@ LiquidStatement <: Liquid {
...
 
 LiquidDoc <: Helpers {
-  Node := (TextNode)*
+  Node := (LiquidDocNode | TextNode)*
+  LiquidDocNode =
+    | paramNode
+    | fallbackNode
+
+  fallbackNode = "@" anyExceptStar<newline>
+  paramNode = "@param" space* (paramType)? space* (paramName)? space* "-"? paramDescription?
```
jamesmengo, line 398: according to the JSDOc [spec](https://jsdoc.app/tags-param), `paramName` seems to be required.

I'm making it optional to enable formatting formatting to apply even when paramName isn't present.

packages/liquid-html-parser/grammar/liquid-html.ohm
jamesmengo: I think we can make paramDescription more permissive - just need to figure out how to do that without bleeding over elsewhere @charlespwd  - any suggestions?

packages/liquid-html-parser/src/stage-2-ast.ts
charlespwd: Given those nodes can be subtree nodes of a LiquidHtml or Liquid tree, LiquidDocCST should be added to the  LiquidCST union. The union should have all the nodes that you could find in the tree.

packages/liquid-html-parser/grammar/liquid-html.ohm
charlespwd: Make it grab everything until the end of line? `anyExceptStar<newline | end>` ? (IIRC end is end of string, might be `eof` or something.

packages/liquid-html-parser/grammar/liquid-html.ohm
```diff
@@ -389,7 +389,17 @@ LiquidStatement <: Liquid {
 }
```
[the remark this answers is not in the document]
jamesmengo: Because LiquidDoc is kind of a special tag (contents are not liquid, but the tag is), I chose to write this explicitly rather than using `blockName` - [ref](https://github.com/Shopify/theme-tools/blob/411d67a0990f1dff8ee71948e16b1628155ef79e/packages/liquid-html-parser/grammar/liquid-html.ohm#L220-L227)

We can revisit this decision if we find value in adding this to `blockName` at a later point!

aswamy (approved):
✅ Tophat went great (including non-param node)
✅ Code looks good

1 question + non-blocking comments

packages/liquid-html-parser/grammar/liquid-html.ohm
```diff
@@ -389,7 +389,16 @@ LiquidStatement <: Liquid {
...
 
 LiquidDoc <: Helpers {
-  Node := (TextNode)*
+  Node := (LiquidDocNode | TextNode)*
+  LiquidDocNode =
+    | paramNode
+    | fallbackNode
+
+  fallbackNode = "@" anyExceptStar<newline>
```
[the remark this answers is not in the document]
jamesmengo: currently the fallbackNode is a catch-all to treat everything that isn't matched as a TextNode, so we'd want it to capture the entire input there

packages/liquid-html-parser/src/stage-1-cst.spec.ts
```diff
@@ -981,53 +981,130 @@ describe('Unit: Stage 1 (CST)', () => {
...
+
+          expectPath(cst, '0.children.0.paramType.type').to.equal('TextNode');
+          expectPath(cst, '0.children.0.paramType.value').to.equal('String');
+          expectPath(cst, '0.children.0.paramType.locStart').to.equal(testStr.indexOf('{String}'));
+          expectPath(cst, '0.children.0.paramType.locEnd').to.equal(
+            testStr.indexOf('{String}') + '{String}'.length,
+          );
+        });
+
+        it('should parse @param with description seperated by a dash', () => {
```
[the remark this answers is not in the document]
jamesmengo: good call!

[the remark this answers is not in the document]
jamesmengo: This is already covered in the test for  unsupported nodes, but I updated the name and test to make it more clear. Good callout!

packages/liquid-html-parser/grammar/liquid-html.ohm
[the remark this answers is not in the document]
charlespwd: We might have to revisit that if we support stuff like `@link` but that's something for later. It's fine for now!

packages/liquid-html-parser/src/stage-1-cst.ts
```diff
@@ -105,6 +107,21 @@ export interface ConcreteBasicNode<T> {
   locEnd: number;
 }
 
+export interface ConcreteLiquidDocParamNode
+  extends ConcreteBasicNode<ConcreteNodeTypes.LiquidDocParamNode> {
+  name: string;
+  value: string;
+  paramName: ConcreteTextNode;
+  paramDescription: ConcreteLiquidDocParamDescription;
+  paramType: ConcreteTextNode;
```
charlespwd, line 116: Isn't this optional? Shouldn't this be ConcreteTextNode | null for when `{}` is not in the input string?

packages/liquid-html-parser/src/stage-1-cst.ts
```diff
@@ -105,6 +107,21 @@ export interface ConcreteBasicNode<T> {
   locEnd: number;
 }
 
+export interface ConcreteLiquidDocParamNode
+  extends ConcreteBasicNode<ConcreteNodeTypes.LiquidDocParamNode> {
+  name: string;
```
charlespwd, line 112: Shouldn't this be `param` or `@param` as a const rather than an arbitrary string?

packages/liquid-html-parser/src/stage-1-cst.ts
```diff
@@ -105,6 +107,21 @@ export interface ConcreteBasicNode<T> {
...
+  value: string;
+  paramName: ConcreteTextNode;
+  paramDescription: ConcreteLiquidDocParamDescription;
```
charlespwd, line 115: Same deal, isn't this optional? Make it part of the type! `| null`

packages/liquid-html-parser/src/stage-2-ast.ts
```diff
@@ -754,6 +755,19 @@ export interface TextNode extends ASTNode<NodeTypes.TextNode> {
   value: string;
 }
 
+export interface LiquidDocParamNode extends ASTNode<NodeTypes.LiquidDocParamNode> {
+  name: string;
+  value: string;
+  paramDescription: LiquidDocParamDescription;
+  paramName: TextNode;
+  paramType: TextNode;
```
charlespwd, line 763: Same deal, please add `/** */` to explain. Also why isn't `name` the `'param'` const?

packages/prettier-plugin-liquid/src/printer/print/liquid.ts
```diff
@@ -490,6 +496,45 @@ export function printLiquidRawTag(
...
+    parts.push(' ', node.paramName.value);
+  }
+
+  if (node.paramDescription.value) {
+    const normalizedDescription = node.paramDescription.value.replace(/\s+/g, ' ').trim();
+    if (node.paramDescription.dashSeparated) {
+      parts.push(' - ', normalizedDescription);
+    } else {
+      parts.push(' ', normalizedDescription);
+    }
```
charlespwd, line 532: IMO this is usually the place where it should be _configurable_ and homogenized by prettier. Forgot to put the dash? We'll put it for you. The goal is to have a consistent result, _not_ to repeat what the input had.

packages/liquid-html-parser/src/stage-1-cst.ts
```diff
@@ -440,10 +457,13 @@ export type LiquidConcreteNode =
   | ConcreteTextNode
   | ConcreteYamlFrontmatterNode;
 
-export type LiquidHtmlCST = LiquidHtmlConcreteNode[];
+export type LiquidHtmlCST = LiquidHtmlConcreteNode[] | LiquidDocCST;
```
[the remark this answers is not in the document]
jamesmengo: Can you lmk if the newest change is what you had in mind?

packages/prettier-plugin-liquid/src/printer/print/liquid.ts
jamesmengo: https://github.com/Shopify/theme-tools/pull/646/commits/f028514e5196b62db1eb0dbfec068cc0e568699b

packages/liquid-html-parser/src/stage-1-cst.ts
```diff
@@ -1306,7 +1319,7 @@ function toLiquidDocAST(source: string, matchingSource: string, offset: number)
 
   const LiquidDocMappings: Mapping = {
     Node: 0,
-    TextNode: {
+    textNode: {
```
charlespwd, line 1322: Hmm. Looks like an unintentional change! The capital letter at the start of a rule is meaningful in ohm so this needs to change! It would make it so the source string would start on a space rather than a non-space character.

> When naming rules, case matters: whitespace is implicitly skipped inside a rule application if the rule name begins with an uppercase letter. For further information, see [Syntactic vs. Lexical Rules](https://ohmjs.org/docs/syntax-reference#syntactic-lexical).

[source](https://ohmjs.org/docs/syntax-reference#:~:text=When%20naming%20rules%2C%20case%20matters%3A%20whitespace%20is%20implicitly%20skipped%20inside%20a%20rule%20application%20if%20the%20rule%20name%20begins%20with%20an%20uppercase%20letter.%20For%20further%20information%2C%20see%20Syntactic%20vs.%20Lexical%20Rules.)

packages/liquid-html-parser/src/stage-1-cst.ts
```diff
@@ -1315,6 +1328,54 @@ function toLiquidDocAST(source: string, matchingSource: string, offset: number)
...
+      paramDescription: function (nodes: Node[]) {
+        const dashNode = nodes[6];
+        const descriptionNode = nodes[7];
+        return {
+          type: ConcreteNodeTypes.TextNode,
+          value: descriptionNode.sourceString.trim(),
+          source,
+          locStart: offset + descriptionNode.source.startIdx,
+          locEnd: offset + descriptionNode.source.endIdx,
+          dashSeparated: dashNode.sourceString.trim() === '-',
```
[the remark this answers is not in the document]
jamesmengo: yep you're right! This is an artifact from when this used to be captured.

charlespwd:
image

image

image

jamesmengo:
### Merge activity

* **Jan 16, 5:17 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/theme-tools/646).

PR #648, opened 2024-12-04
Update contentFor tag grammar to properly extract arguments

veken1199:
It seems that the grammar we had for content_for did not properly extract the contentFor liquid tag argument and convert them into array. 

In this PR we have:
1. Updated the grammar for contentFor liquid tag to support arguments that contain `.`
2. Updated ValidContentForArguments check to support `closest.` instead of `context.`
3. Introduced a new `DeprecatedContentForArgument` check that reports a warning when contentFor tag contains `context.`

### DIFF
--- a/packages/liquid-html-parser/grammar/liquid-html.ohm
+++ b/packages/liquid-html-parser/grammar/liquid-html.ohm
@@ -127,8 +127,13 @@ Liquid <: Helpers {
   liquidTagLiquidMarkup = tagMarkup
 
   liquidTagContentFor = liquidTagRule<"content_for", liquidTagContentForMarkup>
+  
   liquidTagContentForMarkup =
-    contentForType (argumentSeparatorOptionalComma tagArguments) (space* ",")? space*
+    contentForType (argumentSeparatorOptionalComma contentForTagArgument) (space* ",")? space*
+    
+  contentForTagArgument = listOf<contentForNamedArgument<delimTag>, argumentSeparatorOptionalComma>
+  contentForNamedArgument<delim> = (variableSegment ("." variableSegment)*) space* ":" space* (liquidExpression<delim>)
+
   contentForType = liquidString<delimTag>
 
   liquidTagInclude = liquidTagRule<"include", liquidTagRenderMarkup>
--- a/packages/liquid-html-parser/src/stage-1-cst.ts
+++ b/packages/liquid-html-parser/src/stage-1-cst.ts
@@ -81,6 +81,7 @@ export enum ConcreteNodeTypes {
   RenderMarkup = 'RenderMarkup',
   PaginateMarkup = 'PaginateMarkup',
   RenderVariableExpression = 'RenderVariableExpression',
+  ContentForNamedArgument = 'ContentForNamedArgument',
 }
 
 export const LiquidLiteralValues = {
@@ -873,11 +874,21 @@ function toCST<T>(
     },
     simpleArgument: 0,
     tagArguments: 0,
+    contentForTagArgument: 0,
     positionalArgument: 0,
     namedArgument: {
       type: ConcreteNodeTypes.NamedArgument,
       name: 0,
       value: 4,
+      locStart,
+      locEnd,
+      source,
+    },
+
+    contentForNamedArgument: {
+      type: ConcreteNodeTypes.NamedArgument,
+      name: (node) => node[0].sourceString + node[1].sourceString,
+      value: 6,
       locStart,
       locEnd,
       source,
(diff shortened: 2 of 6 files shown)


packages/liquid-html-parser/src/stage-1-cst.ts
```diff
@@ -369,14 +370,20 @@ export interface ConcreteLiquidFilter extends ConcreteBasicNode<ConcreteNodeType
...
   extends ConcreteBasicNode<ConcreteNodeTypes.NamedArgument> {
   name: string;
   value: ConcreteLiquidExpression;
 }
 
+export interface ConcreteLiquidContentForNamedArgument
+  extends ConcreteBasicNode<ConcreteNodeTypes.ContentForNamedArgument> {
+  name: string;
+  value: ConcreteLiquidExpression;
+}
```
charlespwd, line 385: Does this really need a new concrete node type? I see we fold it back into NamedArgument in stage 2. Why not fold it in the contentForTagArgument mapping instead?

packages/liquid-html-parser/grammar/liquid-html.ohm
```diff
@@ -127,8 +127,13 @@ Liquid <: Helpers {
   liquidTagLiquidMarkup = tagMarkup
 
   liquidTagContentFor = liquidTagRule<"content_for", liquidTagContentForMarkup>
+  
   liquidTagContentForMarkup =
-    contentForType (argumentSeparatorOptionalComma tagArguments) (space* ",")? space*
+
+    contentForType (argumentSeparatorOptionalComma contentForTagArgument) (space* ",")? space*
+    contentForTagArgument = listOf<contentForNamedArgument<delimTag>, argumentSeparatorOptionalComma>
+    contentForNamedArgument<delim> = (variableSegment "."? variableSegment?) space* ":" space* (liquidExpression<delim>)
```
charlespwd, line 135: Is this because the `.` is only accepted in content_for named argument parsing?

veken1199: Yes indeed. I don't believe other tag attributes need to support `.` atm. This is why I had to create new contentForTagArgument

packages/theme-check-common/src/checks/deprecated-content-for-arguments/index.ts
```diff
@@ -0,0 +1,71 @@
...
+        if (!validate) {
+          return;
+        }
+
+      
+        validate(node.markup);
+      },
+    };
+  },
+};
```
charlespwd, line 71: Given how transient this thing is and the thing hasn't been released. We can probably not ship this. Just let the one that accepts `closest` complain about `context` not being supported.

charlespwd: Thing is this could live in this repo forever. People will start being able to have that in their `.theme-check.yml` and I can't remember what the behaviour is for when you have settings for checks that don't exist. 

Which would probably mean that removing the check would require a major version bump of theme-check. 

Do we have another check that we could use instead? Something that wouldn't go away? Or maybe something like DeprecatedFeature?

charlespwd: That's probably a nice place for it yeah!

veken1199: dropped this check

albchu (approved):
LGTM!

packages/liquid-html-parser/grammar/liquid-html.ohm
```diff
@@ -127,8 +127,13 @@ Liquid <: Helpers {
   liquidTagLiquidMarkup = tagMarkup
 
   liquidTagContentFor = liquidTagRule<"content_for", liquidTagContentForMarkup>
+  
   liquidTagContentForMarkup =
-    contentForType (argumentSeparatorOptionalComma tagArguments) (space* ",")? space*
+
+    contentForType (argumentSeparatorOptionalComma contentForTagArgument) (space* ",")? space*
+    contentForTagArgument = listOf<contentForNamedArgument<delimTag>, argumentSeparatorOptionalComma>
+    contentForNamedArgument<delim> = (variableSegment ("." variableSegment)*) space* ":" space* (liquidExpression<delim>)
```
charlespwd, line 135: Can we de-indent those two lines? They are new rules and should be at the same indentation level as the other rule declarations. 
```suggestion
  contentForTagArgument = listOf<contentForNamedArgument<delimTag>, argumentSeparatorOptionalComma>
  contentForNamedArgument<delim> = (variableSegment ("." variableSegment)*) space* ":" space* (liquidExpression<delim>)
```

charlespwd (approved):
Other than the nitpick, this LGTM :) Feel free to merge after!

PR #652, opened 2024-12-05
Add basic prettier support for doc tag

jamesmengo:
## Before you deploy

### DIFF
--- a/packages/prettier-plugin-liquid/src/printer/printer-liquid-html.ts
+++ b/packages/prettier-plugin-liquid/src/printer/printer-liquid-html.ts
@@ -1,5 +1,6 @@
 import {
   getConditionalComment,
+  LiquidDocParamNode,
   NodeTypes,
   Position,
   RawMarkupKinds,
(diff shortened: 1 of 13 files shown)


packages/prettier-plugin-liquid/src/printer/printer-liquid-html.ts
```diff
@@ -242,6 +242,15 @@ function printNode(
           ? join(hardline, reindent(lines, shouldSkipFirstLine))
           : softline;
 
+      if (parentNode && parentNode.name === 'doc') {
+        // here we know that we are in a liquidDocNode
+        // Is it better to implement this here or in LiquidRawTag?
+        // One of the benefits of leaving it here is that we can use the indentation and other logic from the raw markup.
+        // The benefit of doing it in LiquidRawTag is that we can keep the logic for doc separate
+        // I still want to understand how embeddings would work in this scenario. Maybe that's a complete different approach that we could take.
+        return [indent([hardline, body]), hardline];
```
charlespwd, line 251: LiquidRawTag exists mostly because we want to paste the body back as it was ("raw"). So `body` here is actually just a string and not a "recursive call to print". This is why the asdf thing at the bottom never gets called. You need to do some kind of `path.call(p => print(p), 'children')` to make it recursive.

Issue #161, opened 2023-10-10
Look into using `@changeset/action` on merge to main instead of running `yarn release` manually.


charlespwd:
**Is your feature request related to a problem? Please describe.**
Turns out @changeset/action can be pretty smart about updating existing PRs and stuff. Might be worth a look.

charlespwd:
Ugh so changeset/action doesn't publish tags for private repos (such as theme-check-vscode). Ugh.

PR #653, opened 2024-12-05
update syntaxes submodule to latest commit

EvilGenius13:

We have recently introduced syntax highlighting for the liquid doc tag in the `liquid-tm-grammar` [repo](https://github.com/Shopify/liquid-tm-grammar/pull/29). This PR updates the submodule commit in `/vscode-extension/syntaxes` to point to the latest commit

## Before you deploy

### DIFF
--- a/.changeset/great-mangos-wait.md
+++ b/.changeset/great-mangos-wait.md
@@ -1,0 +1,5 @@
+---
+'theme-check-vscode': minor
+---
+
+Update vscode-extension syntaxes module to latest commit in order to use liquid doc syntax highlighting


jamesmengo:
I think we're ok without a changeset since we don't have official support for LiquidDoc yet, but will defer to CP on that

charlespwd:
I think the rule should be to always have a changeset. 

Anything on `main` that implicitly depends on something without a changeset is a bug-in-production risk. See #161 

it won't happen in this _particular_ case, but it shouldn't become a habit.

charlespwd:
Also, I think we should wait till the tag is supported to syntax highlight it? OW we might give the impression that it's supported but it isn't.

EvilGenius13:
> Also, I think we should wait till the tag is supported to syntax highlight it? OW we might give the impression that it's supported but it isn't.

That makes sense. I'll add a changeset and we can ship it when the tag itself is ready.

jamesmengo (approved):

PR #659, opened 2024-12-06
static blocks content_for check

miazbikowski:
Addresses https://github.com/Shopify/shopify/issues/559551

The static blocks found at the root level of the preset blocks (meaning, not the nested blocks) need to have a `content_for "block"` for each of those blocks, so that the preset will render correctly. 

image

My section (could be a block) for tophatting:
```
{% content_for 'block', type: 'text', id: 'block1' %}
{% content_for 'block', type: 'text', id: 'block3' %}

{% schema %}
{
  "name": "Slideshow",
  "tag": "section",
  "class": "slideshow",
  "limit": 1,
  "settings": [
    {
      "type": "text",
      "id": "title",
      "label": "Slideshow"
    }
  ],
  "max_blocks": 5,
  "blocks": [
    {
      "type": "text"
    },
    {
      "type": "slide"
    }
  ],
  "presets": [
    {
      "name": "Two blocks, one error",
      "blocks": [
        {
          "type": "text",
          "static": true,
          "id": "block1"
        },
        {
          "type": "slide",
          "static": true,
          "id": "block2"
        }
      ]
    },
    {
      "name": "Single block with an error",
      "blocks": [
        {
          "type": "slide",
          "static": true,
          "id": "block2"
        }
      ]
    },
    {
      "name": "One block, no errors",
      "blocks": [
        {
          "type": "text",
          "static": true,
          "id": "block3"
        }
      ]
    },
    {
      "name": "Blocks Hash: Two blocks, no errors part 2",
      "blocks": {
        "block1": {
          "type": "text",
          "static": true
        },
        "block2": {
          "type": "slide",
          "static": true
        }
      }
    },
    {
      "name": "Blocks Hash: One block, no errors",
      "blocks": {
        "block1": {
          "type": "text",
          "static": true
        }
      }
    },
    {
      "name": "Blocks Hash: One block, one error",
      "blocks": {
        "block2": {
          "type": "slide",
          "static": true
        }
      }
    }
  ]
}
{% endschema %}
```

Note about the approach:
- I debugged and the `LiquidTag` callback worked for `content_for` and not for `schema`
- Vice versa, the `LiquidRawTag` callback worked for `schema` and not for `content_for`

But in the end, it seems to be working well. Both callbacks run before `onCodePathEnd`, at which point we can compare the two lists we've gathered to suss out any missing static block `content_for`s.

### DIFF
--- a/.changeset/silly-parrots-behave.md
+++ b/.changeset/silly-parrots-behave.md
@@ -1,0 +1,7 @@
+---
+'@shopify/theme-check-common': minor
+'@shopify/theme-check-node': minor
+'theme-check-vscode': minor
+---
+
+Add `SchemaPresetsStaticBlocks` check
--- a/packages/theme-check-common/src/checks/index.ts
+++ b/packages/theme-check-common/src/checks/index.ts
@@ -7,7 +7,6 @@ import { AssetSizeAppBlockJavaScript } from './asset-size-app-block-javascript';
 import { AssetSizeCSS } from './asset-size-css';
 import { AssetSizeJavaScript } from './asset-size-javascript';
 import { BlockIdUsage } from './block-id-usage';
-``;
 import { CaptureOnContentForBlock } from './capture-on-content-for-block';
 import { CdnPreconnect } from './cdn-preconnect';
 import { ContentForHeaderModification } from './content-for-header-modification';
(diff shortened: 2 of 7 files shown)


albchu (commented):
Code changes LGTM but I found two tophat cases that raised an eyebrow:
- I tried adding `{% content_for 'block', type: 'text', id: 'block2' %}` with your example schema and the offences persisted. This did not happen for id `block1` or `block3`
image

- If you give a content_for like `{% content_for 'block', type: 'text', id: 'blockDoesNotExist' %}`, I sort of expected to see an error message on the id value. If this one blows the scope of the check, feel free to ignore for now.

miazbikowski:
@albchu thanks for the comments!

1. `{% content_for 'block', type: 'text', id: 'blockDoesNotExist' %}` is actually legit, you don't strictly need to have a preset for a block.

2. I'll look into it - I could've sworn I'd tophatted the same case, so that weird.

PR #660, opened 2024-12-09
Fix bug where regex wouldn't match

graygilmore:

This was an oversight in a regex where we were only handling closing `}}` liquid statements but we also need to include `%}`.

If you were to try doing a filter completion right now on a liquid statement that uses `{% %}` you'd get no results.

Pull down this branch, build, and run the debugger and try doing filter completion on a line like:

```liquid
{% assign imageUrl = product | de█ %}
```

And it should be working again.

## Before you deploy


aswamy (approved):
tophatted it - works as expected 👍

jamesmengo:
🎩 LGTM

jamesmengo (approved):

PR #669, opened 2024-12-11
Add LLM instructions file suggestion

madmath:

This PR introduces support for:
- Suggesting the addition of an LLM instructions file in the repo if there isn't one
- Suggesting the update of an LLM instructions file when it looks outdated
- The "Shopify Magic" button in the top corner of the editor
- The "Shopify Magic" refactoring button in the context menu when users select a piece of Liquid
- The creation of the `.cursorrules/copilot-instructions.md` which works with multiple themes opened in the same workspace
- The `.cursorrules/copilot-instructions.md` files are generated based on https://github.com/Shopify/theme-liquid-docs 

Click [here](https://share.descript.com/view/9rxjC8lhgLy) to check a demo of this feature.

## What's next? Any follow-up issues?

After clicking on “Quick fix,” close the hover dialog automatically (so users may see the suggestions immediately) — I don't think that's possible, but I'm planning to revisit this when we get this work on main.

## Before you deploy

### DIFF
--- a/packages/vscode-extension/src/node/shopify-magic-prompts/base.ts
+++ b/packages/vscode-extension/src/node/shopify-magic-prompts/base.ts
@@ -17,23 +17,34 @@ export function basePrompt(textEditor: TextEditor): string {
         <objective>Optimal Theme Development</objective>
       </prompt_metadata>
       <focus_areas>
-        <area>Enhancing readability, conciseness, and efficiency while maintaining the same functionality</area>
-        <area>Leveraging new features in Liquid, including filters, tags, and objects</area>
-        <area>Be pragmatic and don't suggest without a really good reason</area>
-        <area>Combine multiple operations into one to improve readability and performance -- for example, use the find filter instead of where and first, if find is a valid filter</area>
-        <area>You should not suggest changes to the code that impact only HTML -- they should be focused on Liquid and Theme features.</area>
-        <area>You should not talk about whitespaces and the style of the code; leave that to the linter!</area>
-        <area>**Do not suggest to keep** something that is already valid</area>
+        <area>**Improve the conciseness and efficiency of Liquid code** while maintaining the same functionality</area>
+        <area>**Leverage the most appropriate Liquid** features (filters, tags, and objects); sometimes a filter can achieve the same result as a block of code</area>
+        <area>Avoid logic that matches the default behavior (especially for filters)</area>
+        <area>Improve readability, but **be pragmatic** and **do not suggest unless there's a strong reason**</area>
+        <area>Use filters for operations that perform the same function as a filter</area>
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/node/shopify-magic-prompts/html.ts
```diff
@@ -0,0 +1,14 @@
+export function htmlRules() {
+  return `
+<html>
+  - Use semantic HTML
+  - Use modern HTML features where appropriate, e.g. use \`<details>\` and \`<summary>\` over JS to show and hide content
+  - Use \`CamelCase\` for IDs. When appending a block or section ID, append \`-{{ block.id }}\` or \`-{{ section.id }}\` respectively
```
frandiox, line 6: Did you mean this?
```suggestion
  - Use \`camelCase\` for IDs. When appending a block or section ID, append \`-{{ block.id }}\` or \`-{{ section.id }}\` respectively
```

frandiox (commented):
Looking great, thanks a lot for all the improvements and tests! The fact that it uses up to date liquid docs is such a nice change 🙌

I'm having certificate issues to run this in my new laptop though, so I couldn't 🎩 yet :/

frandiox (approved):
Finally able to test it, great job 🎉 
I've tried to close the hover without much luck either!

While testing, I found a few minor things:

1. I got into this situation a couple of times when applying one suggestion after another. cmd+z and click again fixes the situation... no idea why it happens in the first place :/ 

image
image

2. I saw you added a comment to avoid the "Keep xyz when is already valid"... however this still happens often:

image

Should we maybe compare the suggestion with the code-to-be-replaced and skip it if it's the same? 🤔 

3. Not important but do we rename `sidefix` to something else?

image

packages/vscode-extension/src/node/shopify-magic-prompts/html.ts
karreiro: Actually, I believe this is [accurate](https://github.com/Shopify/dawn/blob/bbfd08b39f12a2193606a973eebcb2d15bc8d957/layout/theme.liquid#L312) 😌

frandiox: Oh, in that case I think we should call it [PascalCase](https://www.freecodecamp.org/news/snake-case-vs-camel-case-vs-pascal-case-vs-kebab-case-whats-the-difference/#heading-what-is-pascal-case)? Or are these things called differently in JS and Ruby worlds?

image

```suggestion
  - Use \`PascalCase\` for IDs. When appending a block or section ID, append \`-{{ block.id }}\` or \`-{{ section.id }}\` respectively

```

graygilmore:
Closing this one for now.

PR #670, opened 2024-12-12
Move loading state to title button

frandiox:
Moves the loading state from an annotation to the title UI button, and disables it while loading:

image
<br>
image


karreiro (approved):
Thank you, @frandiox!

PR #676, opened 2024-12-13
[HackDays - Copilot] Show diff view for suggestions and improve UX

frandiox:
This change uses the diff view / conflict merge UI to show the suggestions.
It also improves the timing for removing suggestions to only when the user types something manually, instead of removing them when accepting 1 single suggestion (the user might want to check the rest).

image


karreiro (approved):
Thanks a lot for this PR, @frandiox! 🔥

PR #677, opened 2024-12-13
Reference to snippet and blocks

aswamy:

Support references for snippets and blocks
- Can't merge it yet because we are reading _every_ file to check for references; should we preload them? memoize the references?

- Support references when you select files in explorer
- Support references found in schemas

## Before you deploy

### DIFF
--- a/packages/theme-language-server-common/src/server/startServer.ts
+++ b/packages/theme-language-server-common/src/server/startServer.ts
@@ -11,6 +11,7 @@ import {
   parseJSON,
   path,
   recursiveReadDirectory,
+  SourceCodeType,
 } from '@shopify/theme-check-common';
 import {
   Connection,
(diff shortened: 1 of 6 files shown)


packages/theme-language-server-common/src/server/startServer.ts
```diff
@@ -203,6 +210,26 @@ export function startServer(
...
+    for(let [uri, _] of [...blockFiles, ...sectionFiles, ...snippetFiles]) {
+      const doc = documentManager.get(uri);
+      if (!doc || doc.type !== SourceCodeType.LiquidHtml) {
+        continue;
+      }
+      sources.push(doc);
+    }
+
+    return sources;
+  }
```
aswamy, line 231: Should we keep a memoized ast for each file and invalidate it on file change? Reading every file when someone looks up references to the variable seems expensive.

aswamy:
Can revive this PR if we feel like it's something we value

PR #695, opened 2025-01-09
[Theme Check] Ensure MissingSchema theme check does not run on snippets

AribaRajput:

Surfaced bug: https://shopify.slack.com/archives/C050J35JHM2/p1736446230098879

This PR ensures that the `MissingSchema` theme check will only run on theme app extension block files and not snippets.

Added some test coverage for this as well. 
## Before you deploy

    - [x] If applicable, I've updated the `theme-app-extension.yml` config


lukeh-shopify (approved):
🚀

packages/theme-check-common/src/checks/missing-schema/index.ts
```diff
@@ -17,6 +17,15 @@ export const MissingSchema: LiquidCheckDefinition = {
...
     let foundSchema = false;
+    const file = context.file;
+    const fileUri = file.uri;
+    const relativePath = context.toRelativePath(fileUri);
+    const isBlockFile = relativePath.startsWith('blocks/');
+
+    if (!isBlockFile) {
+      // Only lint block files as this check is invalid for snippets
+      return {};
+    }
```
graygilmore, line 28: (nit) The extra variables make this a little more verbose and complicated to reason through. Could be simplified:

```suggestion
    const relativePath = context.toRelativePath(context.file.uri);

    /**
     * Theme app extension blocks are the only types of files that can have a
     * schema defined in them.
     */
    if (!relativePath.startsWith('blocks/')) {
      return {};
    }
```

packages/theme-check-common/src/checks/missing-schema/index.ts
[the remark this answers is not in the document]
AribaRajput: Ill update the docs as well in another PR

packages/theme-check-common/src/checks/app-block-missing-schema/index.spec.ts
graygilmore, line 1: For completeness in our testing we probably also want to assert that the check doesn't run when it's a file in a `snippets/` directory that doesn't have a schema.

.changeset/hot-vans-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/theme-check-common': minor
```
graygilmore, line 2: I would call this a patch but maybe renaming the check makes it a minor? We probably want to explicitly call out the rename.

packages/theme-check-common/src/checks/app-block-missing-schema/index.ts
```diff
@@ -1,13 +1,13 @@
 import { ConfigTarget, LiquidCheckDefinition, Severity, SourceCodeType } from '../../types';
 
-export const MissingSchema: LiquidCheckDefinition = {
+export const AppBlockMissingSchema: LiquidCheckDefinition = {
   meta: {
-    code: 'MissingSchema',
-    name: 'Missing schema definitions in theme app extensions should be avoided',
+    code: 'AppBlockMissingSchema',
+    name: 'Missing schema definitions in app blocks (specific to theme app extensions) should be avoided',
```
graygilmore, line 6: (nit)
```suggestion
    name: 'Missing schema definitions in theme app extensions app blocks should be avoided',
```

packages/theme-check-node/configs/all.yml
```diff
@@ -3,6 +3,9 @@
 # Do not modify manually. Your changes will be overwritten.
 ignore:
   - node_modules/**
+AppBlockMissingSchema:
+  enabled: false
+  severity: 0
 AppBlockValidTags:
```
graygilmore, line 9: Nice! We already have an `AppBlock*` config so our new name fits right in!

graygilmore (approved):
Code looks good! Would love to see one more test added and some potential tweaking to the change log.

packages/theme-check-common/src/checks/app-block-missing-schema/index.spec.ts
AribaRajput: I added this to the first test by asserting no errors when its run on a  `snippet,` but I could remove it and make it its own test?

graygilmore: Oooo I see what you did! I think that's fine. I would have structured it slightly differently:

```ts
const sourceCode = `
  // shared source
`;

describe('AppBlockMissingSchema', () => {
  it('reports an error for missing schemas', async () => {
    const offenses = await runLiquidCheck(AppBlockMissingSchema, sourceCode, 'blocks/footer.liquid');
    expect(offenses).to.have.lengthOf(1);
  });

  describe('when the file is not inside the blocks folder', () => {
    it('does not report an error', async () => {
      const offenses = await runLiquidCheck(AppBlockMissingSchema, sourceCode, 'snippets/footer.liquid');
      expect(offenses).to.have.lengthOf(1);
    });
  });

  describe('when the file already has a schema', () => {
    it('does not report an error', async () => {
      const sourceCodeWithSchema = sourceCode + "\n schema string..."

      const offenses = await runLiquidCheck(AppBlockMissingSchema, sourceCodeWithSchema, 'blocks/footer.liquid');
      expect(offenses).to.have.lengthOf(0);
    });
  });
});
```

You don't have to make any of those changes I just wanted to show how I would have approached/thought about testing this.

AribaRajput: Ohh this makes a lot of sense - I will refactor it. Thank you for the suggestion 💯

aswamy (approved):
Thanks for the quick fix. The rename seems fair. I assume [docs](https://shopify.dev/docs/storefronts/themes/tools/theme-check/checks/missing-schema) update is next.

PR #696, opened 2025-01-10
[Theme Check] Add check for duplicate settings id's 

zacariec:
Closes https://github.com/Shopify/develop-advanced-edits/issues/263

Implements a new validation check to prevent duplicate setting IDs across theme configurations. This helps theme developers identify and fix duplicate IDs early in development, preventing potential conflicts and overrides.

The implementation includes:
- New JSONCheckDefinition for unique setting ID validation
- Test coverage for valid and invalid theme configurations
- Detection logic for finding duplicate IDs in settings schema

See:
Ticket: 49020903
https://github.com/Shopify/cli/issues/4187

This PR adds a new theme check validator that detects duplicate setting IDs in theme configurations. The validator scans settings_schema.json files and reports an error when it finds multiple settings using the same ID value, preventing potential runtime conflicts.
Key changes:

- New UniqueSettingIds check that validates setting ID uniqueness
- Integration with the existing theme checker infrastructure
- Test coverage for both valid and invalid scenarios
- Error reporting with precise location information

Potential follow-ups:
- Consider extending the check to validate uniqueness across sections
- Add documentation for theme developers about ID uniqueness requirements
- Consider adding quick-fix suggestions for duplicate IDs

## What did you learn?

Working with AST traversal for JSON validation showed that we need a consistent approach to handle nested property validation. The existing node type checks provided a good pattern to follow.

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/checks/unique-settings-id/index.ts
+++ b/packages/theme-check-common/src/checks/unique-settings-id/index.ts
@@ -1,0 +1,84 @@
+import {
+  isArrayNode,
+  isLiteralNode,
+  isObjectNode,
+  isPropertyNode,
+  Severity,
+  SourceCodeType,
+} from '../../types';
+
+import type { ArrayNode, PropertyNode, JSONCheckDefinition } from '../../types';
+
+export const UniqueSettingIds: JSONCheckDefinition = {
+  meta: {
+    code: 'UniqueSettingId',
+    name: 'Prevent duplicate Ids in setting_schema',
+    docs: {
+      description: 'This check is aimed at eliminating duplicate Ids in settings_schema.json',
+      recommended: true,
+      // url: 'https://shopify.dev/docs/storefronts/themes/tools/theme-check/checks/valid-schema',
+    },
+    type: SourceCodeType.JSON,
+    severity: Severity.ERROR,
+    schema: {},
+    targets: [],
+  },
+
+  create(context) {
+    return {
+      async onCodePathEnd(file) {
+        if (isArrayNode(file.ast)) {
+          const settingIds: PropertyNode[] = [];
+
+          /* Find and loop through all of our nodes that have an id value and find their key value */
+          for (const child of file.ast.children) {
+            if (isObjectNode(child) && child.children) {
+              const settingsNode = child.children.find((node) => node.key.value === 'settings');
+
+              if (settingsNode && settingsNode.value && isArrayNode(settingsNode.value)) {
+                for (const setting of settingsNode.value.children) {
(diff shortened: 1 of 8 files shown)


packages/theme-check-common/src/checks/unique-settings-id/index.ts
```diff
@@ -0,0 +1,84 @@
...
+import type { ArrayNode, PropertyNode, JSONCheckDefinition } from '../../types';
+
+export const UniqueSettingIds: JSONCheckDefinition = {
+  meta: {
+    code: 'UniqueSettingId',
+    name: 'Prevent duplicate Ids in setting_schema',
+    docs: {
+      description: 'This check is aimed at eliminating duplicate Ids in settings_schema.json',
+      recommended: true,
+      // url: 'https://shopify.dev/docs/storefronts/themes/tools/theme-check/checks/valid-schema',
```
jamesmengo, line 19: Do we want to include this url?

packages/theme-check-common/src/checks/unique-settings-id/index.ts
```diff
@@ -0,0 +1,84 @@
...
+            if (isObjectNode(child) && child.children) {
+              const settingsNode = child.children.find((node) => node.key.value === 'settings');
+
+              if (settingsNode && settingsNode.value && isArrayNode(settingsNode.value)) {
+                for (const setting of settingsNode.value.children) {
+                  if (isObjectNode(setting) && setting.children) {
+                    const idNode = setting.children.find((node) => node.key.value === 'id');
+                    if (isPropertyNode(idNode)) {
+                      settingIds.push(idNode);
+                    }
```
aswamy, line 44: We could also move most of the "Check for dupes" code here instead of using a "settingIds" variable.

I guess you didn't want to make the code here too nested and dirty 😅

```suggestion
                    if (isPropertyNode(idNode) && isLiteralNode(idNode.value)) {
                      const id = idNode.value.value;
                      if (typeof id === 'string') {
                        if (!idMap.has(id)) {
                          idMap.set(id, []);
                        }
                        idMap.get(id)!.push(idNode);
                      }
                    }
```

aswamy (approved):
✅ Code looking good
✅ Tophatting went well

packages/theme-check-common/src/checks/unique-settings-id/index.ts
zacariec: @aswamy yep spot on, not a fan of nests and rather use guard clauses but due to the nature of this particular paradigm guard clauses don't fit. Didn't see a style guide from Shopify anywhere either.

PR #697, opened 2025-01-10
[vscode extension] Profile annotations and webview

madmath:

Command to call out to `shopify theme profile` which will

* Put profile annotations (vscode "decorations") in the relevant files 
* Show a webview with the speedscope profile, similar to if the command was run from `shopify theme profile`

Screenshot 2025-01-10 at 8 53 46 AM

Very much a v1, I think we can improve a few ways from here
* Ability to clear the annotations.
* Better resiliency regarding calling shopify CLI (open to suggestions).
* Interactivity between the webview profile and the code.
* Better UX around calling the command. 

## Before you deploy

Public API changes, new features 


charlespwd:
I'm converting this back to draft because there's these things we need to worry about:
- Am I profiling the code I'm looking at?
- How do we deal with VS Code workspaces?

Issue #625, opened 2024-11-28
Theme block JSON preset setting + default setting completion/hover


charlespwd:
Requirements:
- Key completion of `presets.[].settings` (same file)

```liquid
{% schema %}
{
  "name": "Header Announcement",
  "blocks": [{ "type": "slide" }],
  "settings": [
    {
      "type": "text",
      "id": "my-text",
      "label": "Text"
    },
    {
      "type": "text",
      "id": "my-number",
      "label": "Text"
    }
  ],
  "presets": [
    {
      "name": "Header Announcement",
      "settings": {
        "█": "<-- should complete 'my-number', 'my-text'"
      },
    }
  ],
  "default": {
    "settings": {
      "█t": "<-- should complete 'my-number', 'my-text'"
    }
  }
}
{% endschema %}
```

```[tasklist]
### Tasks
- [x] Preset setting completion
- [x] Preset setting hover
- [x] Default setting completion
- [x] Default setting hover
```

PR #700, opened 2025-01-10
Support auto-completion for presets' settings

aswamy:

Part of https://github.com/Shopify/theme-tools/issues/625
- Support presets settings auto-completion
- Support presets setting hover

Both features should work when `setting.label` contains a translation key OR literal text

## Before you deploy

### DIFF
--- a/.changeset/fifty-pugs-confess.md
+++ b/.changeset/fifty-pugs-confess.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-check-common': patch
+---
+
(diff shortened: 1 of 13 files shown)


charlespwd (approved):
This is great. Would it be a big lift to add the same behaviour in `.default.settings`? 

Forgot those exist when I wrote the issue 🙈 

Basically it's the same as a preset but it's not in an array. Those are for static sections IIRC.

https://shopify.dev/docs/storefronts/themes/architecture/sections/section-schema#default

Issue #692, opened 2025-01-08
Git operations are not updating the document manager


charlespwd:
**Describe the bug**
Git operations change files in the workspace without shooting textDocument/did{Open,Change,Save,Close} lifecycle notifications. 

This means that if you revert changes and do a rename, then the internal representation of preloaded files might be off. 

**Expected behaviour**
When files are modified "under our feet", we need to sync them to the document manager with updated values

**Actual behaviour**
Stale values are kept around

PR #703, opened 2025-01-13
[LiquidDoc] Add snippet hover support inside of {% render %} tag

jamesmengo:

https://github.com/Shopify/develop-advanced-edits/issues/496

**1) Hover support for Liquid snippets inside of a `{% render %}` tag. Users will now see:**
- The snippet name as a header
- A list of parameters with their types and descriptions (if 
documented)

If there is no `{% doc %}` present in the snippet, we will just render the snippet name

**2) Caching for fetching LiquidDoc definitions**

##### Example snippet documentation:
```liquid
{% doc %}
  @param {String} title - The title of the product
  @param {Number} border-radius - The border radius in px
{% enddoc %}
```

##### When hovering over `product-card` in `{% render 'product-card' %}`, users will see:
```markdown
### product-card

**Parameters:**
- `title`: String - The title of the product
- `border-radius`: Number - The border radius in px
```

Uploading Cursor - liquidDoc.ts — theme-tools.mp4…

- Consider adding validation for param types
- Add support for return value documentation
- Add support for example usage documentation

## What did you learn?

This is my first time working with the hover API in VS code. Pretty cool to read up on!

## Before you deploy

### DIFF
--- a/packages/theme-language-server-common/src/documents/DocumentManager.ts
+++ b/packages/theme-language-server-common/src/documents/DocumentManager.ts
@@ -12,12 +12,14 @@ import {
   IsValidSchema,
   memo,
   Mode,
+  isError,
 } from '@shopify/theme-check-common';
 import { Connection } from 'vscode-languageserver';
 import { TextDocument } from 'vscode-languageserver-textdocument';
 import { ClientCapabilities } from '../ClientCapabilities';
 import { percent, Progress } from '../progress';
 import { AugmentedSourceCode } from './types';
+import { getSnippetDefinition } from '../liquidDoc';
 
 export class DocumentManager {
   /**
@@ -170,6 +172,12 @@ export class DocumentManager {
 
             const mode = await this.getModeForUri!(uri);
             return toSchema(mode, uri, sourceCode, this.isValidSchema);
+          }),
+          /** Lazy and only computed once per file version */
+          liquidDoc: memo(async (snippetName: string) => {
+            if (isError(sourceCode.ast)) return { name: snippetName };
+
+            return getSnippetDefinition(sourceCode.ast, snippetName);
           }),
         };
       default:
--- a/packages/theme-language-server-common/src/documents/DocumentManager.ts
+++ b/packages/theme-language-server-common/src/documents/DocumentManager.ts
@@ -174,7 +174,7 @@ export class DocumentManager {
             return toSchema(mode, uri, sourceCode, this.isValidSchema);
           }),
           /** Lazy and only computed once per file version */
-          liquidDoc: memo(async (snippetName: string) => {
+          getLiquidDoc: memo(async (snippetName: string) => {
             if (isError(sourceCode.ast)) return { name: snippetName };
 
             return getSnippetDefinition(sourceCode.ast, snippetName);
--- a/packages/theme-language-server-common/src/hover/HoverProvider.ts
+++ b/packages/theme-language-server-common/src/hover/HoverProvider.ts
@@ -1,4 +1,10 @@
-import { MetafieldDefinitionMap, SourceCodeType, ThemeDocset } from '@shopify/theme-check-common';
+import {
+  GetLiquidDocDefinitionsForURI,
+  LiquidDocDefinition,
+  MetafieldDefinitionMap,
+  SourceCodeType,
+  ThemeDocset,
+} from '@shopify/theme-check-common';
 import { Hover, HoverParams } from 'vscode-languageserver';
 import { TypeSystem } from '../TypeSystem';
 import { DocumentManager } from '../documents';
@@ -12,6 +18,7 @@ import {
   LiquidObjectHoverProvider,
   LiquidTagHoverProvider,
   TranslationHoverProvider,
+  RenderSnippetHoverProvider,
 } from './providers';
 import { HtmlAttributeValueHoverProvider } from './providers/HtmlAttributeValueHoverProvider';
 import { findCurrentNode } from '@shopify/theme-check-common';
@@ -26,6 +33,12 @@ export class HoverProvider {
     readonly getMetafieldDefinitions: (rootUri: string) => Promise<MetafieldDefinitionMap>,
     readonly getTranslationsForURI: GetTranslationsForURI = async () => ({}),
     readonly getSettingsSchemaForURI: GetThemeSettingsSchemaForURI = async () => [],
+    readonly getLiquidDocDefinitionsForURI: GetLiquidDocDefinitionsForURI = async (
+      _uri,
+      snippetName,
+    ) => ({
+      name: snippetName,
+    }),
   ) {
     const typeSystem = new TypeSystem(
       themeDocset,
@@ -41,6 +54,7 @@ export class HoverProvider {
       new HtmlAttributeHoverProvider(),
       new HtmlAttributeValueHoverProvider(),
       new TranslationHoverProvider(getTranslationsForURI, documentManager),
+      new RenderSnippetHoverProvider(getLiquidDocDefinitionsForURI),
     ];
   }
 
--- a/packages/theme-language-server-common/src/hover/HoverProvider.ts
+++ b/packages/theme-language-server-common/src/hover/HoverProvider.ts
@@ -17,7 +17,7 @@ import {
 import { HtmlAttributeValueHoverProvider } from './providers/HtmlAttributeValueHoverProvider';
 import { findCurrentNode } from '@shopify/theme-check-common';
 import { GetThemeSettingsSchemaForURI } from '../settings';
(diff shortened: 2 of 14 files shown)


packages/theme-language-server-common/src/server/startServer.ts
```diff
@@ -171,6 +173,23 @@ export function startServer(
     return getDefaultSchemaTranslations();
   };
 
+  const getLiquidDocDefinitionsForURI = async (
```
jamesmengo, line 176: I want to memoize this in a follow-up PR, however we would need to invalidate the cache if the file gets updated or deleted.

I see that we do something kind of similar for metafields [here](https://github.com/Shopify/theme-tools/blob/3b20751ddae8466e59e84f1fbfa0c631a981e6cc/packages/theme-language-server-common/src/server/startServer.ts#L510-L513)

I would want to add this logic in a follow-up outside of this PR, anyhow.

@aswamy before I dive into this, is the `memoize` function a good fit here?

packages/theme-language-server-common/src/hover/providers/SnippetHoverProvider.ts
```diff
@@ -0,0 +1,66 @@
...
+    const parentNode = ancestors.at(-1);
+    if (
+      currentNode.type !== NodeTypes.String ||
+      !parentNode ||
+      parentNode.type !== NodeTypes.RenderMarkup
+    ) {
+      return null;
+    }
+
+    const snippetName = currentNode.value.replace(/['"]/g, '');
```
jamesmengo, line 32: to fix - there may be also be params wrapped in ''

packages/theme-language-server-common/src/documents/DocumentManager.ts
```diff
@@ -171,6 +173,12 @@ export class DocumentManager {
             const mode = await this.getModeForUri!(uri);
             return toSchema(mode, uri, sourceCode, this.isValidSchema);
           }),
+          /** Lazy and only computed once per file version */
+          liquidDoc: memo(async (snippetName: string) => {
+            if (isError(sourceCode.ast)) return { name: snippetName };
```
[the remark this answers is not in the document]
jamesmengo: Just to make sure we're on the same page - I'm currently returning an object that looks like this as a fallback: `{name: snippetName}`.

I wanted the `RenderSnippetHoverProvider` to receive a valid `SnippetDefinition` and not have to handle `undefined`, in which it would have to build its own `{name: snippetName}`.

Does my line of reasoning make sense? (e.g. choosing NOT to return undefined if file is unparseable, since I technically _could_ (but don't want to) add handling downstream)

aswamy (commented):
Did basic tophatting seems good
image


Minor comments

packages/theme-language-server-common/src/hover/HoverProvider.ts
```diff
@@ -26,6 +28,12 @@ export class HoverProvider {
     readonly getMetafieldDefinitions: (rootUri: string) => Promise<MetafieldDefinitionMap>,
     readonly getTranslationsForURI: GetTranslationsForURI = async () => ({}),
     readonly getSettingsSchemaForURI: GetThemeSettingsSchemaForURI = async () => [],
+    readonly getSnippetDefinitionForURI: GetSnippetDefinitionForURI = async (
+      _uri,
```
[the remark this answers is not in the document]
jamesmengo: it could also be for a file that doesn't liquidDoc defined inside

packages/theme-language-server-common/src/documents/DocumentManager.ts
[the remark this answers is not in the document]
jamesmengo: I had the same thoughts and actually moved some of that logic into this [https://app.graphite.dev/github/pr/Shopify/theme-tools/729/LiquidDoc-Prevent-rendering-of-hover-tooltip-for-snippets-with-no-renderable-content] just above it in the stack. Let's chat about this there!


https://app.graphite.dev/github/pr/Shopify/theme-tools/729/LiquidDoc-Prevent-rendering-of-hover-tooltip-for-snippets-with-no-renderable-content

packages/theme-language-server-common/src/hover/providers/RenderSnippetHoverProvider.ts
```diff
@@ -0,0 +1,66 @@
...
+      };
+    }
+
+    const parts = [`### ${snippetDefinition.name}`];
+
+    if (liquidDoc.parameters?.length) {
+      const parameters = liquidDoc.parameters
+        ?.map(
+          ({ name, type, description }: LiquidDocParameter) =>
+            `- \`${name}\`${type ? `: ${type}` : ''} ${description ? `- ${description}` : ''}`,
```
jamesmengo, line 52: Decision: choosing to render the type if present, which is different from JSDOC which doesn't show the type in hover

packages/theme-language-server-common/src/documents/DocumentManager.ts
```diff
@@ -171,6 +173,12 @@ export class DocumentManager {
             const mode = await this.getModeForUri!(uri);
             return toSchema(mode, uri, sourceCode, this.isValidSchema);
           }),
+          /** Lazy and only computed once per file version */
+          getLiquidDoc: memo(async (snippetName: string) => {
```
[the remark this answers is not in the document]
charlespwd: Yes that's what #692 is about lol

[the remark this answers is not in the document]
jamesmengo: sweet! In that case, @karreiro wdyt about keeping the memoization as-is?

packages/theme-language-server-common/src/hover/providers/RenderSnippetHoverProvider.ts
karreiro: Personally, I believe we could show `{Object}`, because ultimately, it's the type the language servers will understand. With that, we also tie this feature with typing, in the sense of clarifying that types are not optional with the doc tag. What do you think?

packages/theme-language-server-common/src/liquidDoc.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export type GetSnippetDefinitionForURI = (
+  uri: string,
+  snippetName: string,
+) => Promise<SnippetDefinition>;
+
+export type LiquidDocParameter = {
+  name: string;
+  description: string | null;
+  type: string | null;
```
aswamy, line 15:
```suggestion
  description?: string;
  type?: string;
```

packages/theme-language-server-common/src/liquidDoc.ts
```diff
@@ -0,0 +1,50 @@
...
+): SnippetDefinition {
+  const liquidDocParameters: LiquidDocParameter[] = visit<
+    SourceCodeType.LiquidHtml,
+    LiquidDocParameter
+  >(snippet, {
+    LiquidDocParamNode(node: LiquidDocParamNode) {
+      return {
+        name: node.paramName.value,
+        description: node.paramDescription?.value ?? null,
+        type: node.paramType?.value ?? null,
```
aswamy, line 39:
```suggestion
        description: node.paramDescription?.value,
        type: node.paramType?.value,
```

packages/theme-language-server-common/src/liquidDoc.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    });
+  });
+
+  it('should extract name, description and type from param annotations', async () => {
+    const ast = toAST(`
+        {% doc %}
+          @param {String} firstParam - The first param
+          @param {Number} secondParam - The second param
+          @param paramWithNoType - param with no type
+          @param paramWithOnlyName
```
aswamy, line 35: You should also have:
- 1 param with type but no description
- 1 `@unsupport-param`
- 1 free-form text
image

packages/theme-language-server-common/src/hover/providers/RenderSnippetHoverProvider.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+          name: 'title',
+          description: 'The title of the product',
+          type: 'string',
+        },
+        {
+          name: 'border-radius',
+          description: 'The border radius in px',
+          type: 'number',
+        },
+      ],
```
aswamy, line 24: You should add more iterations of parameters (e.g. without description, type, etc) since you have logic behind the templating around them.

See review comment above for the ones I've done during tophatting.

aswamy (approved):
Looking good James 😎 Small comments, but none blocking merge

image

image

- I put a breakpoint in the memoized `getLiquidDoc` and looks like it's working well.

karreiro (approved):
Thank you, @jamesmengo! LGTM and works as expected as well 🎩 :)

Issue #571, opened 2024-11-07
Support `content_for "block"` type completion + document link


charlespwd:
We want block type completion

```
{% content_for "block", type: "█" %}
{% content_for "block", type: "foo█" %}
```

Steps:
- Create a `ContentForBlockTypeCompletionProvider` in theme-language-server-common/src/completions/providers
- Import it and add an instance to the list of providers in [theme-language-server-common/src/completions/CompletionProvider](https://github.com/smintfy/theme-tools/blob/main/packages/theme-language-server-common/src/completions/CompletionsProvider.ts#L1)
- Logic should be something like
  - Early return unless 
    - the `node.type === NodeTypes.String`
    - the `parentNode.type === NodeTypes.NamedArgument` && its name is `=== 'type'`
    - the `grandParentNode.type === NodeTypes.ContentForMarkup` && its `contentForType` is `=== 'block'`
  - Return all the options from the block schema registry (depends on #572)

PR #709, opened 2025-01-14
Support `content_for` block type completion + document link

aswamy:

Fixes https://github.com/Shopify/theme-tools/issues/571

Support `content_for` block type completion + document link

## Before you deploy

### DIFF
--- a/packages/theme-language-server-common/src/completions/providers/ContentForBlockTypeCompletionProvider.ts
+++ b/packages/theme-language-server-common/src/completions/providers/ContentForBlockTypeCompletionProvider.ts
@@ -1,0 +1,42 @@
+import { NodeTypes } from '@shopify/liquid-html-parser';
+import { CompletionItem, CompletionItemKind } from 'vscode-languageserver';
+import { LiquidCompletionParams } from '../params';
+import { Provider } from './common';
+
+export class ContentForBlockTypeCompletionProvider implements Provider {
+  constructor(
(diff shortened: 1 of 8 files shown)


packages/theme-language-server-common/src/completions/providers/ContentForBlockTypeCompletionProvider.ts
```diff
@@ -0,0 +1,42 @@
...
+      parentNode.name !== 'type' ||
+      grandParentNode.type !== NodeTypes.ContentForMarkup ||
+      grandParentNode.contentForType.value !== 'block'
+    ) {
+      return [];
+    }
+
+    return (await this.getThemeBlockNames(doc.uri, false)).map((blockName) => ({
+      label: blockName,
+      kind: CompletionItemKind.Keyword,
```
karreiro, line 38: It doesn't seem like we're completing a keyword, so I personally believe we should use `CompletionItemKind.EnumMember`, as the completion items are part of a well-defined set of options (standardized and formally grouped).

Alternatively, we could use `CompletionItemKind.Value` as well, but it doesn't seem so ideal since it refers to literal values not tied to a certain group.

(Just mentioning this because this impacts the icons of the completion items, so it also gives extra clues for users) :)

karreiro (commented):
Thank you for this PR, @aswamy! Excellent stuff 🎩 It works as expected, and I've left only one minor comment about the completion items :)

PR #711, opened 2025-01-16
Add allowedDomains check for RemoteAsset theme checker

### LINKED ISSUE #410: Add config exceptions to Remote Asset check

There are valid exceptions for the single CDN domain that might repeat a lot, and it would be nice to document those on the config file, my personal example now is with `polyfill.io` that does some magic not available on shopify CDN and is not beauty to add exceptions comments on each place. Other example would be a shop that choose to use Cloudinary across the theme, it is clearly better if the config file could make clear what remotes are "valid" for the store.

``` yaml
RemoteAsset:
  additional_remotes:
    - polyfill.io
```

**Describe alternatives you've considered**

Using disable comments solves the issue when most violations can be made on the same file, but it adds too much noise when violations are expected on many files.

**Additional context**

I've been reading [the implementation of the check](https://github.com/Shopify/theme-check/blob/main/lib/theme_check/checks/remote_asset.rb), and it seems that adding this feature seems to be pretty straightforward.

zacariec:
I'll update the PR template based on the revised commit message that focuses on developer experience and performance:

Fixes #410

This PR adds domain validation to the RemoteAsset theme checker to ensure remote assets (scripts, stylesheets) are loaded from Shopify CDNs and approved domains. When a remote asset is referenced from an unlisted domain, the checker will provide guidance to help developers optimize performance and follow best practices.

Key changes:
- New `allowedDomains` configuration in RemoteAsset check schema
- Domain validation logic that checks resources against approved domains
- Performance-focused warning messages guiding developers to use CDNs
- Test coverage for both allowed and non-allowed domain scenarios

Potential follow-ups:
- Add documentation about approved domains and CDN usage
- Consider updating existing warnings when not using Shopify CDN to also flag they can add as an allowed domain
- Consider creating a guide for determining which domains should be allowlisted

## What did you learn?
The variety of ways developers include remote assets highlighted the need for clear guidance on performance best practices. The test cases helped identify common patterns in how external resources are referenced in themes.

## Before you deploy
  - [ ] If applicable, I've updated the `theme-app-extension.yml` config (N/A - performance check applies to all themes)


packages/theme-check-common/src/checks/remote-asset/index.ts
```diff
@@ -96,7 +97,11 @@ function valueIsShopifyHosted(attr: ValuedHtmlAttribute): boolean {
   });
 }
 
-export const RemoteAsset: LiquidCheckDefinition = {
+const schema = {
+  allowedDomains: SchemaProp.array(SchemaProp.string()),
```
charlespwd, line 101: From the errors, looks like you're missing the way to make those optional.

PR #714, opened 2025-01-16
Support auto-completion on every preset setting

aswamy:

Fixes bug where we only support preset setting auto-completion for the first preset object

## Before you deploy


charlespwd:
Looks like we have an unrelated thing not being bocked? Why are we calling `which shopify` in the tests?

charlespwd (approved):

PR #729, opened 2025-01-23
[LiquidDoc] Return undefined for invalid snippet definitions

jamesmengo:
## Before you deploy


jamesmengo:
### Snippet exists, but no liquidDoc
image


### Snippet does not exist
image
- Renders a red squiggly + warning via theme check

karreiro (approved):
Thank you for this PR, @jamesmengo.

I don't feel strongly about this, but I believe this is the ideal presentation for an undocumented snippet:
image

As a user, it feels like the language server has identified the snippet successfully, it just doesn't have documentation to present, so it feels like it's working well.

Here, we're just showing a link, and it feels like the language server has not identified that snippet:
image

I personally believe the [first one](https://github.com/user-attachments/assets/06225c3f-2ce2-4b52-93ad-c3ead374eab8) is more satisfying to use, but that's a personal opinion so I'm happy with both directions.

jamesmengo:
@karreiro I'm on board with going in that direction! I like your line of reasoning

PR #733, opened 2025-01-24
[LiquidDoc] Parser support for optional parameters

jamesmengo:

Part of https://github.com/Shopify/develop-advanced-edits/issues/525

Adds parser support for `[` and `]` delimiters around the `stringName`, which indicate that a param is **optional**

Within the code, I'm storing this as 'required' rather than 'optional'. The default will be `required: true`

`@param [paramName]`

- https://github.com/Shopify/theme-tools/pull/735

## Before you deploy

### DIFF
--- a/.changeset/smart-onions-pump.md
+++ b/.changeset/smart-onions-pump.md
@@ -1,0 +1,5 @@
+---
+'@shopify/liquid-html-parser': minor
+---
+
+Add parser support for optional liquiddoc parameters
--- a/packages/liquid-html-parser/grammar/liquid-html.ohm
+++ b/packages/liquid-html-parser/grammar/liquid-html.ohm
@@ -400,11 +400,16 @@ LiquidDoc <: Helpers {
   openControl:= "@" | end
 
   fallbackNode = "@" anyExceptStar<endOfParam>
-  paramNode = "@param" strictSpace* paramType? strictSpace* paramName (strictSpace* "-")? strictSpace* paramDescription
+  paramNode = "@param" strictSpace* paramType? strictSpace* (optionalParamName | paramName) (strictSpace* "-")? strictSpace* paramDescription
   paramType = "{" strictSpace* paramTypeContent strictSpace* "}"
   paramTypeContent = anyExceptStar<("}"| strictSpace)>
-  paramName = identifierCharacter+
-  paramDescription = anyExceptStar<endOfParam>
+
+  paramName = (~endOfParamName identifierCharacter)+
+  optionalParamName = "[" strictSpace* optionalParamNameContent strictSpace* "]"
+  optionalParamNameContent = anyExceptStar<endOfParamName>
+  endOfParamName = strictSpace* ("]" | "@")
+
+  paramDescription = (~"]" anyExceptStar<endOfParam>)
   endOfParam = strictSpace* (newline | end)
 }
 
--- a/packages/liquid-html-parser/src/stage-1-cst.spec.ts
+++ b/packages/liquid-html-parser/src/stage-1-cst.spec.ts
@@ -1011,11 +1011,12 @@ describe('Unit: Stage 1 (CST)', () => {
           expectPath(cst, '0.children.0.value').to.equal('@param');
         });
 
-        it('should parse @param with name', () => {
+        it('should parse required @param with name', () => {
           const testStr = `{% doc %} @param paramWithNoDescription {% enddoc %}`;
           cst = toCST(testStr);
 
           expectPath(cst, '0.children.0.type').to.equal('LiquidDocParamNode');
+          expectPath(cst, '0.children.0.required').to.equal(true);
           expectPath(cst, '0.children.0.paramName.type').to.equal('TextNode');
           expectPath(cst, '0.children.0.paramName.value').to.equal('paramWithNoDescription');
           expectPath(cst, '0.children.0.paramName.locStart').to.equal(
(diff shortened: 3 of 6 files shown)


jamesmengo:
* **#733** <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/733?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/733?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/liquid-html-parser/grammar/liquid-html.ohm
```diff
@@ -403,7 +403,8 @@ LiquidDoc <: Helpers {
   paramNode = "@param" strictSpace* paramType? strictSpace* paramName (strictSpace* "-")? strictSpace* paramDescription
   paramType = "{" strictSpace* paramTypeContent strictSpace* "}"
   paramTypeContent = anyExceptStar<("}"| strictSpace)>
-  paramName = identifierCharacter+
+  paramName =  "["? strictSpace* paramNameContent strictSpace* "]"?
```
graygilmore, line 406: Since only having a single square bracket isn't valid I wonder if we should change to something like this:

```ohm
paramNode = "@param" strictSpace* paramType? strictSpace* (optionalParam | paramName) (strictSpace* "-")? strictSpace* paramDescription
optionalParam = "[" strictSpace* paramName strictSpace* "]"
```

Then we either match on `optionalParam` _or_ `paramName` and there's no ambiguity.

packages/liquid-html-parser/src/stage-1-cst.spec.ts
```diff
@@ -1028,6 +1029,77 @@ describe('Unit: Stage 1 (CST)', () => {
...
+          expectPath(cst, '0.children.1.paramName.locEnd').to.equal(
+            testStr.indexOf('paramWithWhitespace') + 'paramWithWhitespace'.length,
+          );
+
+          expectPath(cst, '0.children.2.type').to.equal('LiquidDocParamNode');
+          expectPath(cst, '0.children.2.required').to.equal(false);
+          expectPath(cst, '0.children.2.paramType.value').to.equal('String');
+        });
+
+        it('should parse @param with missing optional delimiters as required', () => {
```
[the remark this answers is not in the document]
graygilmore: My previous comment speaks to this a bit: I'm not a huge fan of supporting invalid syntax. My preference would be for this to fail to be parsed.

packages/liquid-html-parser/src/stage-1-cst.ts
```diff
@@ -1341,10 +1342,17 @@ function toLiquidDocAST(source: string, matchingSource: string, offset: number)
       paramType: 2,
       paramName: 4,
       paramDescription: 8,
+      required: function (nodes: Node[]) {
+        const paramName = nodes[4];
+        return (
+          paramName.children[0].sourceString === '' || paramName.children[4].sourceString === ''
```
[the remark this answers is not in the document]
graygilmore: Instead of checking for empty strings could we use a regex here with `test()`? I wonder if it'd easier to understand in the future what this is doing. Something like:

```ts
const regex = /^\[\s*\w*\s*\]$/;
const name = nodes.map((n) => n.sourceString).join('');
return regex.test(name)
```

graygilmore (dismissed):

packages/liquid-html-parser/grammar/liquid-html.ohm
jamesmengo: Modified the approach to go in this direction. This helps resolve some other downstream issues I was having deciding on behaviour, so 💯 agree

.changeset/smart-onions-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/liquid-html-parser': minor
+---
+
+Add parser support for optional liquiddoc parameters
```
graygilmore, line 5: I know that this isn't the most important thing in the world but might be worth getting in the habit of beefing these up a bit. I know it might not feel like the most important thing now (just like how writing robust commit messages doesn't feel worth it in the moment) but it might be the thing that saves somebody later!

```suggestion
Add support for parsing optional parameters in LiquidDoc. Optional parameters are now designated by wrapping the name in square brackets (e.g. `@param [parameterName]`).
```

packages/liquid-html-parser/grammar/liquid-html.ohm
```diff
@@ -400,11 +400,16 @@ LiquidDoc <: Helpers {
...
+  paramNode = "@param" strictSpace* paramType? strictSpace* (optionalParamName | paramName) (strictSpace* "-")? strictSpace* paramDescription
   paramType = "{" strictSpace* paramTypeContent strictSpace* "}"
   paramTypeContent = anyExceptStar<("}"| strictSpace)>
-  paramName = identifierCharacter+
-  paramDescription = anyExceptStar<endOfParam>
+
+  paramName = (~endOfParamName identifierCharacter)+
+  optionalParamName = "[" strictSpace* optionalParamNameContent strictSpace* "]"
+  optionalParamNameContent = anyExceptStar<endOfParamName>
+  endOfParamName = strictSpace* ("]" | "@")
```
graygilmore, line 410: A param name can end with a `@`? Or is "end" here more like "book ends"?

graygilmore (approved):
Did not get a 🎩 in yet but since this is just parser changes maybe the tests are all we need here!

packages/liquid-html-parser/grammar/liquid-html.ohm
jamesmengo: `anyExceptStar<endOfParamName>` translates to `(~endOfParamName any)*`, which will prevent 

`@param [border_width @asdf ]` from matching as a `optionalParamName`. 

Instead it will match `@param [border_width ` and be a `fallbackNode`

Though I'm questioning the direction here 🤔  let's pair!

jamesmengo:
Closing in favour of https://github.com/Shopify/theme-tools/pull/753

packages/liquid-html-parser/src/stage-1-cst.ts
```diff
@@ -111,11 +112,18 @@ export interface ConcreteBasicNode<T> {
 export interface ConcreteLiquidDocParamNode
   extends ConcreteBasicNode<ConcreteNodeTypes.LiquidDocParamNode> {
   name: 'param';
-  paramName: ConcreteTextNode;
+  paramName: ConcreteLiquidDocParamNameNode;
   paramDescription: ConcreteTextNode | null;
   paramType: ConcreteTextNode | null;
 }
 
+export interface ConcreteLiquidDocParamNameNode
```
jamesmengo, line 120: It may look kind of weird to have `required` live in this node, however this is consistent with how the parsing logic works, and allows us to use the `mappings` in `stage-1`.

We do this because it aligns with the parsing grammar, and allows us to leverage the `Mappings` mechanism. We map this to a more sensible location in `stage-2` [here](https://github.com/Shopify/theme-tools/pull/733/files#diff-d4d621a0e25c584008f71bf2ca34d635d50e0a6a1af987addb4ecf76ce649300R1302) (`node.paramName.required -> node.required`)

aswamy (approved):
Looks great! Can't wait for prettier PR!

PR #735, opened 2025-01-25
Wrap optional parameters name in `[]` in hover tooltip ----- Behaviour diverges slightly from JSDoc here. JSDoc does not indicate that it's optional, rather it implicitly adds `| undefined` to the types

jamesmengo:
## Before you deploy


packages/theme-language-server-common/src/hover/providers/RenderSnippetHoverProvider.spec.ts
```diff
@@ -16,57 +16,42 @@ describe('Module: RenderSnippetHoverProvider', async () => {
...
-        systemTranslations: async () => ({}),
-      },
-      async (_rootUri: string) => ({} as MetafieldDefinitionMap),
-      async () => ({}),
-      async () => [],
-      getSnippetDefinition,
-    );
-  };
-
-  beforeEach(async () => {
```
jamesmengo, line 60: I want to make extending custom cases easier, rather than having a massive test case that's hard to read / understand

packages/theme-language-server-common/src/hover/providers/RenderSnippetHoverProvider.ts
```diff
@@ -67,4 +61,15 @@ export class RenderSnippetHoverProvider implements BaseHoverProvider {
       },
     };
   }
+
+  private buildParameters(parameters: LiquidDocParameter[]) {
+    return parameters
+      .map(({ name, type, description, required }: LiquidDocParameter) => {
+        const nameStr = required ? `\`${name}\`` : `\`[${name}]\``;
```
aswamy, line 68: I wrote JS doc for a typescript function for optional params. I even had square bracket for optional params, but when you hover over the function, it doesn't show it with square brackets in the hover text.


- Should we do something like for ours as well?
- Or should we have an explicit "optional" after the paramName?

image

image


Suggestion:
image

aswamy (approved):
Looks great! One question for discussion, but not required.

packages/theme-language-server-common/src/hover/providers/RenderSnippetHoverProvider.ts
jamesmengo: I like this idea! It's more expressive. Let's go in this direction :)

PR #737, opened 2025-01-27
[LiquidDoc] Modify paramDescription to be null if empty

jamesmengo:

Returns `null` in the `liquid-html-parser` if `paramDescription` value is `''`. 

Also added some test coverage in the language server for this case.

- [paramDescription](https://github.com/Shopify/theme-tools/blob/main/packages/liquid-html-parser/grammar/liquid-html.ohm#L403) is always matched in the grammar. I experimented with changing that matching logic but ultimately felt it was simpler this logic in instead `stage-2`, though we could revisit this

### DIFF
--- a/packages/liquid-html-parser/src/stage-2-ast.ts
+++ b/packages/liquid-html-parser/src/stage-2-ast.ts
@@ -1974,7 +1974,8 @@ function toHtmlSelfClosingElement(
 }
 
 function toNullableTextNode(node: ConcreteTextNode | null): TextNode | null {
-  if (!node) return null;
+  if (!node || node.value === '') return null;
+
   return toTextNode(node);
 }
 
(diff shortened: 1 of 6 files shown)


packages/liquid-html-parser/src/stage-2-ast.ts
```diff
@@ -1974,7 +1974,8 @@ function toHtmlSelfClosingElement(
 }
 
 function toNullableTextNode(node: ConcreteTextNode | null): TextNode | null {
-  if (!node) return null;
+  if (!node || node.value === '') return null;
```
jamesmengo, line 1977: @charlespwd I would prefer to update the `ohm` logic, but it wasn't so straightforward since `paramDescription` is always matched.

Wanted to get a sanity check on this decision, or some help editing `ohm`!

packages/theme-language-server-common/src/liquidDoc.spec.ts
```diff
@@ -33,6 +33,7 @@ describe('Unit: makeGetLiquidDocDefinitions', () => {
           @param {Number} secondParam - The second param
           @param paramWithNoType - param with no type
           @param paramWithOnlyName
+          @param {Number} paramWithNoDescription
```
aswamy, line 36: you should also include a test for free-form text to see if it gets picked up by `getSnippetDefinition`

aswamy (approved):
The updated tests look good. Thanks for the follow-up PR. 👍

Issue #624, opened 2024-11-28
Theme block JSON preset validation


charlespwd:
Maybe duplicate of what @navdeep5 is doing? 

Should we add a tasklist in here?

Or a duplicate of https://github.com/Shopify/theme-tools/issues/602

Top of mind:
```[tasklist]
### Tasks
- [x] Make sure setting id/value pairs in presets match the block
- [x] Make sure block type is accepted by parent
- [x] Preset names can be a translation key. Show an error when this translation key does not resolve to a value. (can share overlap to the work done for verifying schema name value). This validation should work for section presets as well.
- [x] ... ?
```

bakura10:
I think this is somehow related. But since theme blocks are introduced (I think) we can now connect a block value as a default. For instance, inside a section, I have this block:

```
"blocks": [
    {
      "type": "collection",
      "name": "t:sections.featured_collections.blocks.collection.name",
      "settings": [
        {
          "type": "collection",
          "id": "collection",
          "label": "t:sections.featured_collections.blocks.collection.collection"
        },
        {
          "type": "text",
          "id": "title",
          "label": "t:global.text.title",
          "default": "{{ block.settings.collection.title }}"
        }
  }
]
```

This is convenient and allow to automatically connect the title with the collection's title of the block.

Theme check report does not currently allow that though:

<img width="589" alt="Image" src="https://github.com/user-attachments/assets/0476b47d-9dbc-415e-a480-7201e0984d5e" />

I am not sure if this is expected @charlespwd

charlespwd:
Hmm that's def a bug. Looks like Liquid in `{% schema %}` is being visited but shouldn't be.

PR #742, opened 2025-01-29
Theme check for settings keys inside presets and default

aswamy:

Fixes https://github.com/Shopify/theme-tools/issues/624
Fixes https://github.com/Shopify/theme-tools/issues/602
- Talked to @charlespwd and we won't be doing setting value check here

Shoutout to @AribaRajput for creating the initial PR to get this done quickly
- Closing this previous PR: https://github.com/Shopify/theme-tools/pull/620/files

- Found a few missing pieces of completion / hover / checks for schema that ill be tackling

## How to test

NOTE: This check DOES NOT check if the theme block exists or not. That is the responsibility of another theme check. I have noticed a bug where if you have local blocks, and there are references to local blocks that dont exist within `presets` and `default`, no theme check error is thrown. I will fix that in a following PR.

### Testing settings

1. Create preset/default settings
  ```
  "settings": [{
      "type": "text",
      "id": "top-level-setting",
      "label": "TEXT"
    }
  ],
  ```
2. Create `presets.[].settings` and `default.settings` that has keys that match the `settings.id` defined above. Create some that don't. See error.

### Testing preset/default block settings defined in local block

1. Create a local block with settings
  ```
  "blocks": [
    {
      "type": "local-block",
      "name": "local block",
      "settings": [
        {
          "id": "local-setting",
          "type": "text",
          "label": "local setting"
        }
      ]
    }
  ]
  ```
3. Create `presets.[].blocks.[].settings` and `default.blocks.[].settings` that has keys that match the `settings.id` defined above. Create some that don't. See error.

### Testing preset/default block settings defined in theme block

1. Create blocks that allows all theme blocks
  ```
  "blocks": [
    {
      "type": "@theme"
    }
  ]
  ```
2. Create `presets.[].blocks.[].settings` and `default.blocks.[].settings` that has keys that match the `settings.id` defined in the block file being referenced. Create some that don't. See error.

## Before you deploy

### DIFF
--- a/.changeset/forty-moose-pay.md
+++ b/.changeset/forty-moose-pay.md
@@ -1,0 +1,9 @@
+---
+'@shopify/theme-check-common': minor
+'@shopify/theme-check-node': minor
+---
+
+Theme check verifies if setting key exists within block schemas and section schemas
+
+- Check if the keys inside `presets.[].settings` and `default.settings` exist as `settings.[].id` in the same file
+- Check if the keys inside `presets.[](recursive .blocks.[]).settings` and `default.blocks.[].settings` exist as `settings.[].id` inside the referenced block's file
--- a/packages/theme-check-common/src/checks/valid-settings-key/index.spec.ts
+++ b/packages/theme-check-common/src/checks/valid-settings-key/index.spec.ts
@@ -1,0 +1,321 @@
+import { expect, describe, it } from 'vitest';
+import { ValidSettingsKey } from './index';
+import { check } from '../../test';
+
+describe('Moduel: ValidSettingsKey', () => {
+  const schemaTemplate = {
+    name: 'Example',
+    settings: [
+      {
+        id: 'existent-setting',
+        type: 'text',
+        label: 'Example Text Setting',
+      },
+    ],
+  };
+
+  describe('default settings', () => {
+    it('does not report an error when default setting exists', async () => {
+      const theme = {
+        'sections/example.liquid': toLiquidFile({
+          ...schemaTemplate,
+          default: {
+            settings: {
+              'existent-setting': 'value',
+            },
+          },
+        }),
+      };
+
+      const offenses = await check(theme, [ValidSettingsKey]);
+      expect(offenses).to.have.length(0);
+    });
+
+    it('reports an error when default setting does not exist', async () => {
+      const theme = {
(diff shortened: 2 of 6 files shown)


.changeset/forty-moose-pay.md
```diff
@@ -0,0 +1,9 @@
+---
+'@shopify/theme-check-common': minor
+'@shopify/theme-check-node': minor
+---
+
+Theme check verifies if setting key exists in schema
```
jamesmengo, line 6: Should we specify that this is for theme blocks? Or does this apply for other asset types as well?

packages/theme-check-common/src/checks/valid-settings-key/index.spec.ts
```diff
@@ -0,0 +1,317 @@
...
+      };
+
+      const offenses = await check(theme, [ValidSettingsKey]);
+      expect(offenses).to.have.length(1);
+      expect(offenses[0].message).to.equal(
+        `Setting 'non-existent-setting' does not exist in schema.`,
+      );
+    });
+
+    it('does not report an error when default settings are defined in a block schema', async () => {
```
jamesmengo, line 53: Took me a second to see what we were doing here, maybe we could clarify this


```suggestion
    it('does not report an error when missing default settings is referenced from a block schema', async () => {
```

jamesmengo (approved):
Review went well, and the only edge case I could find is when referencing a block with theme check errors (this doesn't render anything, which is expected - in this screenshot, `menu` has an invalid schema)

I wonder if we could report invalid schemas 🤔

Either way, this changeset LGTM!

packages/theme-check-common/src/checks/valid-settings-key/index.spec.ts
```diff
@@ -0,0 +1,317 @@
...
+          },
+        }),
+      };
+
+      const offenses = await check(theme, [ValidSettingsKey]);
+      expect(offenses).to.have.length(1);
+      expect(offenses[0].message).to.equal(
+        `Setting 'non-existent-setting' does not exist in schema.`,
+      );
+    });
```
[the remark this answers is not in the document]
aswamy: I can definitely do 2 missing settings. But checking across multiple rules wouldn't be unit testing _this_ rule.

Issue #570, opened 2024-11-07
Support `content_for "block"` named argument completion


charlespwd:

mbarak:
"context" attributes are also supported in `{% content_for "blocks" %}` (blocks plural as opposed to block singular in the description). This variant doesn't accept `id` and `type` arguments but does support "context" attributes/arguments (like the singular block variant).

The supported context arguments are:
- `context.<core_resource>` ex: `context.product`
- `context.metaobject.<metaobject type>` ex: `context.metaobject.123-metaobject_type`

See here for more details:

https://github.com/Shopify/shopify/blob/main/areas/core/shopify/components/online_store/app/liquid/tags/content_for/block_strategy.rb#L9-L17

PR #745, opened 2025-01-30
Add completion support for content_for parameters

graygilmore:

Part of https://github.com/Shopify/theme-tools/issues/570

- Update shopify.dev so that `content_for` tag has `content.*` param

## Tophatting

- Clone code
- Open VSCode and press F5
- Open a Theme in the new window
- Have a block in the `blocks` folder (e.g. `blocks/example.liquid`
- In a section file, create a content_for 'block' tag
  - `{% content_for 'block' %}` - if you used auto-completion, the `id` and `type` should be prepopulated
- Try adding a `context.` param

NOTE: You can always force trigger auto-completion from VSCode using ctrl + space 

## Before you deploy

### DIFF
--- a/.changeset/fresh-suns-provide.md
+++ b/.changeset/fresh-suns-provide.md
@@ -1,0 +1,5 @@
+---
+'@shopify/liquid-html-parser': minor
+---
+
+Support parsing incomplete content_for tags for completions
--- a/.changeset/fresh-suns-provide.md
+++ b/.changeset/fresh-suns-provide.md
@@ -1,5 +1,7 @@
 ---
 '@shopify/liquid-html-parser': minor
+'@shopify/theme-language-server-common': minor
 ---
 
-Support parsing incomplete content_for tags for completions
+- Support parsing incomplete content_for tags in completion context
+- Support content_for param completion
--- a/packages/liquid-html-parser/grammar/liquid-html.ohm
+++ b/packages/liquid-html-parser/grammar/liquid-html.ohm
@@ -133,6 +133,7 @@ Liquid <: Helpers {
     contentForType (argumentSeparatorOptionalComma contentForTagArgument) (space* ",")? space*
 
   contentForTagArgument = listOf<contentForNamedArgument<delimTag>, argumentSeparatorOptionalComma>
+  completionModeContentForTagArgument = listOf<contentForNamedArgument<delimTag>, argumentSeparatorOptionalComma> (argumentSeparator? (liquidVariableLookup<delimTag>))?
   contentForNamedArgument<delim> = (variableSegment ("." variableSegment)*) space* ":" space* (liquidExpression<delim>)
 
   contentForType = liquidString<delimTag>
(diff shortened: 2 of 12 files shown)


aswamy:
Force update:
- rebase with main
- simply stage-2's toContentForMarkup logic
- added a change to ObjectCompletionProvider so we don't have extra completion items
image

packages/theme-language-server-common/src/completions/providers/ObjectCompletionProvider.ts
```diff
@@ -20,6 +20,11 @@ export class ObjectCompletionProvider implements Provider {
       return [];
     }
 
+    // ContentFor uses VariableLookup to support completion of NamedParams.
+    if (ancestors.at(-1)?.type === NodeTypes.ContentForMarkup) {
```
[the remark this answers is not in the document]
aswamy: Hmm not sure what you mean. If you mean the values of `closest` params, (e.g. `content.product: █`), it wouldn't have a parent of `NodeTypes.ContentForMarkup` (it would be `NodeTypes.NamedParam`).

If you are referring to `closest.█` lookups, we are still fine because `ObjectCompletionProvider` has a quick return (right above this code) that returns if the variable is composed of multiple parts (like `closest` + `product`).

.changeset/fresh-suns-provide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/liquid-html-parser': minor
+---
+
+Support parsing incomplete content_for tags for completions
```
jamesmengo, line 5: Lets update this / add one for completions

packages/liquid-html-parser/src/stage-1-cst.spec.ts
```diff
@@ -1776,6 +1776,25 @@ describe('Unit: Stage 1 (CST)', () => {
...
+      const toCST = (source: string) => toLiquidHtmlCST(source, { mode: 'completion' });
+
+      cst = toCST(`{% content_for "blocks", id: 1, cl█ %}`);
+
+      expectPath(cst, '0.markup.type').to.equal('ContentForMarkup');
+      expectPath(cst, '0.markup.args.0.type').to.equal('NamedArgument');
+      expectPath(cst, '0.markup.args.1.type').to.equal('VariableLookup');
+    });
+
+    it('blah blah', () => {
```
charlespwd, line 1790: :D

.changeset/fresh-suns-provide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/liquid-html-parser': minor
```
charlespwd, line 2: missing update in language server

packages/liquid-html-parser/grammar/liquid-html.ohm
```diff
@@ -551,18 +552,24 @@ StrictLiquidHTML <: LiquidHTML {
 
 WithPlaceholderLiquid <: Liquid {
   liquidFilter<delim> := space* "|" space* identifier (space* ":" space* filterArguments<delim> (space* ",")?)?
+  liquidTagContentForMarkup :=
+    contentForType (argumentSeparatorOptionalComma completionModeContentForTagArgument) (space* ",")? space*
```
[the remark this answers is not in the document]
graygilmore: I *almost* went down this path but ending up bailing out after hitting an awkward blocker that would require too much of a lift for this one feature. I think we *should* eventually add this as a standard piece for all param completions, though!

PR #734, opened 2025-01-25
[LiquidDoc] Formatting support for optional `@param` delimiters `[]`

jamesmengo:
## What are you adding in this PR?
Part of https://github.com/Shopify/developer-tools-team/issues/525

Adds prettier support for LiquidDoc parameters with optional params, indicated by wrapping the param name in `[]`

Cursor -  Extension Development Host  product-variant-picker liquid — jm-theme

## What's next? Any followup issues?
https://github.com/Shopify/theme-tools/pull/735

## Before you deploy

- [x] I included a minor bump `changeset`
- [x] My feature is backward compatible

PR #753, opened 2025-02-04
[LiquidDoc] Parsing support for optional parameters and default values

jamesmengo:

Closes [533](https://github.com/Shopify/developer-tools-team/issues/533)
Closes [534](https://github.com/Shopify/developer-tools-team/issues/534)

Adds `LiquidDoc` parsing support to allow users to define optional parameters and default values.

**Optional Parameters**
- By default, parameters are now required unless they have `[]` around the name
- Parameters with incomplete delimiters `e.g. ([missingTail)` will map to a `TextNode`

**Default Values**
- Default values are supported for optional parameters - `[param=default]`
- Default values will match anything between `=` and `]`. Leading and Trailing spaces are trimmed.

https://github.com/Shopify/theme-tools/pull/734

## Before you deploy

### DIFF
--- a/packages/liquid-html-parser/src/stage-1-cst.spec.ts
+++ b/packages/liquid-html-parser/src/stage-1-cst.spec.ts
@@ -1011,21 +1011,340 @@ describe('Unit: Stage 1 (CST)', () => {
           expectPath(cst, '0.children.0.value').to.equal('@param');
         });
 
-        it('should parse @param with name', () => {
-          const testStr = `{% doc %} @param paramWithNoDescription {% enddoc %}`;
+        it('should parse required @param', () => {
+          const testStr = `{% doc %}
+            @param paramWithNameOnly
+            @param paramWithNoType - this has no type
+            @param {String} paramWithNoDescription
+            @param {String} paramWithDescriptionAndType - param with description and type
+          {% enddoc %}`;
           cst = toCST(testStr);
 
           expectPath(cst, '0.children.0.type').to.equal('LiquidDocParamNode');
-          expectPath(cst, '0.children.0.paramName.type').to.equal('TextNode');
-          expectPath(cst, '0.children.0.paramName.value').to.equal('paramWithNoDescription');
+          expectPath(cst, '0.children.0.paramName.type').to.equal('LiquidDocParamNameNode');
+          expectPath(cst, '0.children.0.paramName.paramNameContent.value').to.equal(
+            'paramWithNameOnly',
+          );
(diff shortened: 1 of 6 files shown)


packages/liquid-html-parser/src/stage-1-cst.spec.ts
```diff
@@ -1011,30 +1011,352 @@ describe('Unit: Stage 1 (CST)', () => {
           expectPath(cst, '0.children.0.value').to.equal('@param');
         });
 
-        it('should parse @param with name', () => {
-          const testStr = `{% doc %} @param paramWithNoDescription {% enddoc %}`;
+        it('should parse required @param', () => {
```
jamesmengo, line 1014: In these tests, I typically test permutations, and include `location` assertions for the first permutation only (aside from the whitespace case).

To avoid being overly verbose / repetitive, I'm just checking the value of the `description` / `type` to ensure that nothing has completely broken in the matching logic.

packages/liquid-html-parser/src/stage-1-cst.ts
```diff
@@ -110,11 +111,18 @@ export interface ConcreteBasicNode<T> {
 export interface ConcreteLiquidDocParamNode
   extends ConcreteBasicNode<ConcreteNodeTypes.LiquidDocParamNode> {
   name: 'param';
-  paramName: ConcreteTextNode;
+  paramName: ConcreteLiquidDocParamNameNode;
   paramDescription: ConcreteTextNode | null;
   paramType: ConcreteTextNode | null;
 }
 
+export interface ConcreteLiquidDocParamNameNode
```
jamesmengo, line 119: The syntax for default value and required properties are related to how the `ParamName` is defined, so I'm choosing to include that information in this `ConcreteLiquidDocParamNameNode` 

This makes it a bit funky to access, but this aligns with how the parsing logic works, and we map this to a more sensible location in `stage-2`

packages/liquid-html-parser/src/stage-2-ast.ts
```diff
@@ -1285,14 +1289,11 @@ function buildAst(
           name: node.name,
           position: position(node),
           source: node.source,
-          paramName: {
-            type: NodeTypes.TextNode,
-            value: node.paramName.value,
-            position: position(node.paramName),
-            source: node.paramName.source,
-          },
+          paramName: toTextNode(node.paramName.paramNameContent),
```
jamesmengo, line 1292: `@param {String} [param_name=123]`

This allows us to refer to the `node.paramName`, which will be the node referring to `param_name` with the associated location info

PR #725, opened 2025-01-20
Add parsing and prettier support for @example tags in liquid doc

EvilGenius13:
## What are you adding in this PR?

References: https://github.com/Shopify/develop-advanced-edits/issues/471

Added support for parsing and formatting `@example` tags within `LiquidDoc` blocks.

## Feedback
Right now, we aren't controlling whitespace for example tags. If someone wants to write `check   out.     my great    example`, it would stay in that format. Should trim whitespace or allow people writing examples to have full control?

## What's next? Any followup issues?

I want to add prettier support to automatically move `@example` tags and content below `@param` as it is the proper flow.

## What did you learn?

AST's and CST's are quite puzzle to figure out.

## Test
https://github.com/user-attachments/assets/70c8452b-820d-4091-acbb-dd6a720821cc

## Before you deploy

- [ ] I included a minor bump `changeset`
- [ ] My feature is backward compatible

PR #759, opened 2025-02-04
stage 1/2 & prettier refactor along with extra test additions

EvilGenius13:

Follow up to https://github.com/Shopify/theme-tools/pull/725

There were some comments on refactoring and adding additional test coverage around `exampleNodes`. This also adds a changeset as the other PR did not have one.

I will be making another pass at prettier support when someone has a `liquiddoc` node with multiple `@example` tags as well as spacing between `@param` and `@example` tags.

## Before you deploy

### DIFF
--- a/.changeset/many-seahorses-smoke.md
+++ b/.changeset/many-seahorses-smoke.md
@@ -1,0 +1,14 @@
+---
+'@shopify/prettier-plugin-liquid': minor
+'@shopify/liquid-html-parser': minor
+---
+
+Add parsing and prettier support for example node in liquiddoc
+Example:
+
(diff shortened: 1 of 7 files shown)


.changeset/many-seahorses-smoke.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/prettier-plugin-liquid': minor
+'@shopify/liquid-html-parser': minor
+---
+
+Add parsing and prettier support for example node in liquiddoc
```
jamesmengo, line 6: Let's beef this up slightly by adding an example :)

PR #763, opened 2025-02-05
Example tag hover support

EvilGenius13:

Closes: https://github.com/Shopify/developer-tools-team/issues/529

Add hover support for `@example` in liquid `{% doc %}` tags.

Along with `parameter` hover introduced in https://github.com/Shopify/theme-tools/pull/703, we are adding support for examples.

Example product card snippet:
```liquid
{% doc %}
  @param {String} title - The title of the product
  @param {Number} border-radius - The border radius in px
  @example
  {% render 'product-card', title: 'Product Title', border-radius: 10 %}
  @example
  {% render 'product-cards', title: 'Product Title', border-radius: 20 %}
{% enddoc %}

<div class="product-card" style="border-radius: {{ border-radius }}px">
  <h2>{{ title }}</h2>
</div>
```

Hovering over the render snippet will look like:
image

## Before you deploy


packages/theme-language-server-common/src/liquidDoc.ts
```diff
@@ -16,35 +16,57 @@ export type SnippetDefinition = {
...
+        name: node.paramName.value,
+        description: node.paramDescription?.value ?? null,
+        type: node.paramType?.value ?? null,
+        nodeType: 'param',
+      };
     },
-  );
+    LiquidDocExampleNode(node: LiquidDocExampleNode) {
+      return {
+        content: node.exampleContent.value,
```
aswamy, line 56: Actually, we might have to be a bit smarter than just trimming the string. When we have an example written like the following, trimming the leading+trailing spaces will end up looking like the 2nd picture.
image
image

Instead, we should see how many space characters are leading the first line, then remove that many spaces for each line in the example. This also means that you need to update the prettier rule so that we allow format on save. What do you think?

cc: @jamesmengo

aswamy (commented):
Tophat went great Josh!

image

packages/theme-language-server-common/src/liquidDoc.ts
aswamy: If you want, you can separate multi-line example support and multi-line param support into another ticket. I assume they would encounter the same formatting issues.

jamesmengo (approved):

PR #775, opened 2025-02-10
Update the document manager on git operations

### LINKED ISSUE #692: Git operations are not updating the document manager

charlespwd:

Tricky part was to not invalidate the cache twice for the same change.

Lots of LSP lifecycle messages (onDidCreateFiles, onDidRenameFiles,
onDidDeleteFiles, onDidSaveTextDocument) were fired before the
onDidChangeWatchedFile events.

Had to do a little trick to avoid invalidating the cache twice for
those.

Fixes #692

## Before you deploy

### DIFF
--- a/packages/theme-language-server-common/src/server/startServer.ts
+++ b/packages/theme-language-server-common/src/server/startServer.ts
@@ -16,7 +16,6 @@ import {
 import {
   Connection,
   FileChangeType,
-  FileEvent,
   FileOperationRegistrationOptions,
   InitializeResult,
   ShowDocumentRequest,
@@ -36,6 +35,7 @@ import { OnTypeFormattingProvider } from '../formatting';
 import { HoverProvider } from '../hover';
 import { JSONLanguageService } from '../json/JSONLanguageService';
 import { LinkedEditingRangesProvider } from '../linkedEditingRanges/LinkedEditingRangesProvider';
+import { SnippetDefinition } from '../liquidDoc';
 import { RenameProvider } from '../rename/RenameProvider';
 import { RenameHandler } from '../renamed/RenameHandler';
 import { GetTranslationsForURI } from '../translations';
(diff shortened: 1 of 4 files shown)


packages/theme-language-server-common/src/server/startServer.ts
```diff
@@ -568,12 +571,79 @@ export function startServer(
...
+      fs.readDirectory.invalidate(path.dirname(change.uri));
+      fs.readFile.invalidate(change.uri);
+      fs.stat.invalidate(change.uri);
+
+      // Update the document manager with the new contents of the file
+      switch (change.type) {
+        case FileChangeType.Changed:
+        case FileChangeType.Created:
+          // If a file is created (or changed) under out feet, we update its contents.
+          promises.push(documentManager.changeFromDisk(change.uri));
```
jamesmengo, line 616:
```suggestion
          // If a file is created (or changed) under our feet, we update its contents.
```

jamesmengo (approved):
Works for updating `LiquidDoc` definitions when running `git checkout / git reset` 👍

packages/theme-language-server-common/src/server/startServer.ts
```diff
@@ -568,12 +571,79 @@ export function startServer(
...
+   *   - onDidRenameFiles
+   *   - onDidDeleteFiles
+   *   - onDidSaveTextDocument
+   *
+   * Not redundant for operations that happen outside of the editor
+   *   - git pull, checkout, reset, stash pop, etc.
+   *   - shopify theme metafields pull
+   *   - etc.
+   */
+  connection.onDidChangeWatchedFiles(async (params) => {
```
aswamy, line 598: Why can't this be the central logic for invalidating cache instead of `onDidX` methods? Are there any situations where `onDidCreateFiles` would trigger, but not this?

charlespwd: You are 100% right. Corrected in https://github.com/Shopify/theme-tools/pull/775/commits/326f02dc3c5a689614abc0c9afdba7305448fe9d

PR #789, opened 2025-02-13
Support liquid doc inner tags completion + hover

aswamy:

Fixes https://github.com/Shopify/developer-tools-team/issues/506
- Supports code completion of `@param` `@example` and `@description` tags inside of `doc` liquid tag
- Can hover over each of the tags to show help doc

## Tophat

- Create a doc tag
- When you type `@` and trigger code-completion, it should provide you with `@param` `@example` and `@description` options
- Once you complete the code and fill out the data, you can hover over `@X` to show its help doc

## Before you deploy


EvilGenius13 (approved):
Thank you for this PR @aswamy! 🎩 went well. One quick question. I was able to get each tag to show up when I typed `@e` `@p` and `@d`. However, when I tried just typing `@`, it would not pop up with the 3 options as referenced in `packages/theme-language-server-common/src/completions/providers/LiquidDocTagCompletionProvider.spec.ts`. Not sure if this is just my local dev not working as expected.

aswamy:
> it would not pop up with the 3 options

@EvilGenius13 Thanks for the investigation. That's because code completion does not trigger when you only have 1 character like `@`. If you want to force-trigger code-completion, just do `ctrl` + `space`.

EvilGenius13:
> > it would not pop up with the 3 options
> 
> @EvilGenius13 Thanks for the investigation. That's because code completion does not trigger when you only have 1 character like `@`. If you want to force-trigger code-completion, just do `ctrl` + `space`.

That worked! Thank you 🚀

packages/theme-language-server-common/src/utils/liquidDoc.ts
```diff
@@ -0,0 +1,49 @@
...
+    return `### ${nameStr}${typeStr}${descStr}`;
+  }
+
+  const descStr = description ? ` - ${description}` : '';
+  return `- ${nameStr}${typeStr}${descStr}`;
+}
+
+export function formatLiquidDocTagHandle(label: string, description: string, example: string) {
+  return `### @${label}\n\n${description}\n\n` + `**Example**\n\n\`\`\`liquid\n${example}\n\`\`\``;
+}
```
aswamy, line 21: Ignore this - it is part of [this PR](https://github.com/Shopify/theme-tools/pull/774#).

It will go away once it's merged and i rebase against main.

jamesmengo: can we set that as the base branch for this PR, or would that screw up the diff?

PR #794, opened 2025-02-14
Add doc tag auto completion

EvilGenius13:
References: https://github.com/Shopify/developer-tools-team/issues/530
Add autocompletion for the doc tag.
Ex: If you start typing `{% do` an auto completion option will come up for `doc`
If you accept it, it will look like:
```liquid
{% doc %}

{% enddoc %}
```

The actual implementation for the doc tag lives [here](https://github.com/Shopify/theme-liquid-docs/blob/main/data/tags.json) in the `theme-liquid-docs` repo. We don't want to ship usage of the tag yet. For now, the tag has been hardcoded and will be subsequently removed when we officially ship it in the `theme-liquid-docs` repo.

## Before you deploy

### DIFF
--- a/packages/theme-language-server-common/src/completions/providers/LiquidTagsCompletionProvider.ts
+++ b/packages/theme-language-server-common/src/completions/providers/LiquidTagsCompletionProvider.ts
@@ -40,6 +40,17 @@ export class LiquidTagsCompletionProvider implements Provider {
     const partial = node.name.replace(CURSOR, '');
     const blockParent = findParentNode(partial, ancestors);
     const tags = await this.themeDocset.tags();
+    if (!tags.some(tag => tag.name === 'doc')) {
+      tags.push({
+        name: 'doc',
+        category: 'theme',
+        description: 'Creates a documentation block in your theme.',
+        syntax: '{% doc %}\n content\n{% enddoc %}',
+        syntax_keywords: [{ keyword: 'content', description: 'The content of the doc' }],
+        deprecated: false,
+        parameters: [],
+      });
+    }
     return tags
       .filter(({ name }) => name.startsWith(partial))
       .sort(sortByName)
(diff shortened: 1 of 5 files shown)


packages/theme-language-server-common/src/completions/providers/LiquidTagsCompletionProvider.ts
```diff
@@ -40,6 +40,17 @@ export class LiquidTagsCompletionProvider implements Provider {
     const partial = node.name.replace(CURSOR, '');
     const blockParent = findParentNode(partial, ancestors);
     const tags = await this.themeDocset.tags();
+    if (!tags.some((tag) => tag.name === 'doc')) {
```
EvilGenius13, line 43: This is where we hardcoded the tag for now. The reason for the additional check is that this `completions` is run multiple times and will push the doc tag each time without it.

packages/theme-language-server-common/src/completions/providers/LiquidTagsCompletionProvider.ts
```diff
@@ -40,6 +40,17 @@ export class LiquidTagsCompletionProvider implements Provider {
...
+      tags.push({
+        name: 'doc',
+        category: 'theme',
+        description: 'Creates a documentation block in your theme.',
+        syntax: '{% doc %}\n content\n{% enddoc %}',
+        syntax_keywords: [{ keyword: 'content', description: 'The content of the doc' }],
+        deprecated: false,
+        parameters: [],
+      });
```
aswamy, line 52: > We don't want to ship usage of the tag yet. For now, the tag has been hardcoded and will be subsequently removed when we officially ship it in the theme-liquid-docs repo.

Doesn't this temporary solution and the real solution actively do the same thing? Or does pushing to `theme-liquid-docs` do more?

Like does pushing to that repo automatically generate a docs page that we don't want to expose yet?

PR #791, opened 2025-02-14
Add ValidRenderSnippetParams check for snippet render validation

jamesmengo:
## Before you deploy

- [ ] This PR includes a new checks or changes the configuration of a check
  - [ ] I included a minor bump `changeset`
  - [ ] It's in the `allChecks` array in `src/checks/index.ts`
  - [ ] I ran `yarn build` and committed the updated configuration files
    
    
    
    - [ ] If applicable, I've updated the `theme-app-extension.yml` config
  - [ ] I've made a PR to update the [shopify.dev theme check docs](https://github.com/Shopify/shopify-dev/tree/main/content/storefronts/themes/tools/theme-check/checks) if applicable (link PR here).

- [ ] I included a minor bump `changeset`
- [ ] My feature is backward compatible

- [ ] I included a patch bump `changeset`

PR #795, opened 2025-02-14
Support Liquid doc param type completion

aswamy:

https://github.com/Shopify/developer-tools-team/issues/508

- Adds type completion for liquid docs
- NOTE: packages/theme-check-common/src/liquid-doc/utils.ts is actually from @jamesmengo [PR](https://github.com/Shopify/theme-tools/pull/791) that isn't merged yet

## Tophat 1

- Tophat
- Create a doc tag
- When you type `@param {` and trigger code-completion, it should provide you with types

## Tophat 2

- Render a snippet with liquid doc params
- If the param is `string` or unknown, the completion will place surrounding quotes and put your cursor in the middle of it
- If the param is `boolean`, `object`, or `number`, it will place a default value
  - Press tab to keep the default value
  - Start typing to override the default value

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/liquid-doc/utils.ts
+++ b/packages/theme-check-common/src/liquid-doc/utils.ts
@@ -1,0 +1,30 @@
+export enum SupportedParamTypes {
+  String = 'string',
+  Number = 'number',
+  Boolean = 'boolean',
+  Object = 'object',
+}
+
+export enum SupportedDocTagTypes {
+  Param = 'param',
+  Example = 'example',
+  Description = 'description',
+}
+
+/**
+ * Provides a default completion value for an argument / parameter of a given type.
+ */
+export function getDefaultValueForType(type: string | null) {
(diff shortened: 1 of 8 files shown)


packages/theme-check-common/src/liquid-doc/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+  Number = 'number',
+  Boolean = 'boolean',
+  Object = 'object',
+}
+
+export enum SupportedDocTagTypes {
+  Param = 'param',
+  Example = 'example',
+  Description = 'description',
+}
```
aswamy, line 12: I only added this part - i will rebase the branch once [this PR](https://github.com/Shopify/theme-tools/pull/791) is merged

jamesmengo (approved):
LGTM! I created a separate  issue on the board to only return the suggestions when in a snippet file (this would apply to both the tags and the types).

Issue #788, opened 2025-02-13
[Feature] Theme check - Disable checks using Liquid comments `next-line`


madsenmm:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
Would be nice to expand this check to work as eslint, with the option to only disable check for next-line.
Prevents the need to enable check after the code block again.

**Current:**

```liquid
# theme-check-disable UnknownFilter
assign has_layout = ids | has: 'layout'
# theme-check-enable UnknownFilter
```

**Feature request**
```liquid
# theme-check-disable-next-line UnknownFilter
assign has_layout = ids | has: 'layout'
```

**Additional context**
https://shopify.dev/docs/storefronts/themes/tools/theme-check/configuration#disable-checks-using-liquid-comments

PR #796, opened 2025-02-15
Add ability to disable theme check for next line

graygilmore:

- Closes https://github.com/Shopify/theme-tools/issues/788

Adds the ability to disable a theme check rule only on the next line with `theme-check-disable-next-line`:

```liquid
{% # theme-check-disable-next-line UnknownFilter %}
{{ product | foo }}
```

Just like `theme-check-disable` you can pass no specific rule to disable all rules:

```liquid
{% # theme-check-disable-next-line %}
{% assign x = product | foo %}
```

## Before you deploy

- [ ] (will do this once we have approval of this approach) I've made a PR to update the [shopify.dev theme check docs](https://github.com/Shopify/shopify-dev/tree/main/content/storefronts/themes/tools/theme-check/checks) if applicable (link PR here).


packages/theme-check-common/src/disabled-checks/index.ts
```diff
@@ -25,6 +27,18 @@ export function createDisabledChecksModule() {
...
 
+      if (command === 'disable-next-line') {
+        if (!disabledRanges.has(check)) {
+          disabledRanges.set(check, []);
+        }
+        disabledRanges.get(check)!.push({
+          from: position.end,
+          // Find the index of the second newline which signifies the end of the
+          // "next line" that we're disabling the rule for.
+          to: position.end + source.slice(position.end).indexOf('\n', 1),
```
graygilmore, line 38: This is the most fragile piece of this approach. When we're in this part of the code the `source` is just a string. What I'm doing here is getting the index of the *second* new line by passing `1` as the second argument to `indexOf`. The very first character after the slice should be a new line which should return an index of `0` so we add 1 to that number to change the starting point of the search.

```
                                              slice start                                slice end
                                                   v                                          v
{% # theme-check-disable-next-line UnknownFilter %}(\n  {{ product | broken }}\n{{ variable }})
                                                    ^^                        ^^
                                               first new line            next new line (end)
```

packages/theme-check-common/src/disabled-checks/index.ts
```diff
@@ -25,6 +27,26 @@ export function createDisabledChecksModule() {
         disabledRanges.get(check)!.push({ from: position.end });
       }
 
+      if (command === 'disable-next-line') {
+        if (!disabledRanges.has(check)) {
+          disabledRanges.set(check, []);
+        }
+
+        const nextNewLineIndex = source.slice(position.end).indexOf('\n', 1);
```
graygilmore, line 35: This is the most fragile piece of this approach. When we're in this part of the code the `source` is just a string. What I'm doing here is getting the index of the *second* new line by passing `1` as the second argument to `indexOf`. The very first character after the slice should be a new line which should return an index of `0` so we add 1 to that number to change the starting point of the search.

```
                                              slice start                                slice end
                                                   v                                          v
{% # theme-check-disable-next-line UnknownFilter %}(\n  {{ product | broken }}\n{{ variable }})
                                                    ^^                        ^^
                                               first new line            next new line (end)
```

packages/theme-check-common/src/disabled-checks/index.spec.ts
```diff
@@ -125,6 +125,85 @@ ${buildComment('theme-check-enable LiquidFilter')}
...
+        for (const buildComment of commentTypes) {
+          const file = `{% render 'something-1' %}
+            {{ 'asset-1' | random_filter }}
+            ${buildComment('theme-check-disable-next-line LiquidFilter')}
+
+            {{ 'asset-2' | random_filter }}`;
+
+          const offenses = await check({ 'code.liquid': file }, checks);
+
+          expect(offenses).toHaveLength(3);
```
jamesmengo, line 202: This is kind of self documenting, but I had to go up to verify that the expected # of offenses is 3, and that the file value was similar to that of the previous tests. Maybe we could add a comment for that?

packages/theme-check-common/src/disabled-checks/index.spec.ts
```diff
@@ -125,6 +125,85 @@ ${buildComment('theme-check-enable LiquidFilter')}
...
+            }),
+          ];
+
+          expect(offenses).toEqual(expect.arrayContaining(expectedOffenses));
+          expect(offenses).toHaveLength(expectedOffenses.length);
+        }
+      });
+    });
+
+    describe('when next line is used on an empty line', () => {
```
jamesmengo, line 191: Is this something that could blow up if the comment is placed in the last line of the file? If so, is it worth adding a test for that?

(i.e. there's no other empty lines after it)

packages/theme-check-common/src/disabled-checks/index.ts
```diff
@@ -25,6 +27,28 @@ export function createDisabledChecksModule() {
...
+
+        console.log({ nextNewLineIndex });
```
jamesmengo, line 37: 🔥

jamesmengo: lol

jamesmengo (approved):

PR #797, opened 2025-02-18
Validate visible_if fields

six-shopify:

Fixes https://github.com/Shopify/shopify/issues/570056. In the end there were a few different cases I felt merited their own error messages, so the scope of this check is a little bit broader than the above issue.

I didn't see any other rules that operated on both Liquid and JSON files; in the end the check exports two rule definitions (which enforce the same rule). Maybe there's an easier way to do this that I missed!

- Update shopify.dev theme check docs

## Before you deploy

      - unsure if this is applicable, i'm not familiar with theme app extensions
    - to do

### DIFF
--- a/.changeset/sixty-clouds-confess.md
+++ b/.changeset/sixty-clouds-confess.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-check-common': minor
+---
+
+New check: ValidVisibleIf
--- a/.changeset/sixty-clouds-confess.md
+++ b/.changeset/sixty-clouds-confess.md
@@ -2,4 +2,4 @@
 '@shopify/theme-check-common': minor
 ---
 
-Add `ValidVisibleIf` check
+Add `ValidVisibleIf` check to ensure `visible_if` expressions are well-formed and only reference defined settings keys
--- a/.changeset/sixty-clouds-confess.md
+++ b/.changeset/sixty-clouds-confess.md
@@ -2,4 +2,4 @@
 '@shopify/theme-check-common': minor
 ---
 
-New check: ValidVisibleIf
+Add `ValidVisibleIf` check
--- a/packages/theme-check-common/src/checks/valid-visible-if/index.ts
+++ b/packages/theme-check-common/src/checks/valid-visible-if/index.ts
@@ -1,12 +1,10 @@
 import {
   toLiquidHtmlAST,
-  toLiquidAST,
   type LiquidVariableLookup,
-  NodeTypes,
   type LiquidString,
 } from '@shopify/liquid-html-parser';
 import { LiquidCheckDefinition, Severity, SourceCodeType, type Context } from '../../types';
-import { nodeAtPath, parseJSON } from '../../json';
+import { getLocStart, nodeAtPath, parseJSON } from '../../json';
 import { getSchema, isBlockSchema, isSectionSchema } from '../../to-schema';
 import { reportWarning } from '../../utils';
 import { visit } from '../../visitor';
@@ -14,6 +12,9 @@ import { findRoot } from '../../find-root';
 import { join } from '../../path';
 
 const variableExpressionMatcher = /{{(.+?)}}/;
+const adjustedPrefix = '{% if ';
+const adjustedSuffix = ' %}{% endif %}';
+const offsetAdjust = '{{'.length - adjustedPrefix.length;
 
 // TODO: also perform analysis for json files?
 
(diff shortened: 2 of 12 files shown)


packages/theme-check-common/src/checks/valid-visible-if/index.ts
```diff
@@ -0,0 +1,296 @@
...
+async function getGlobalSettings(context: Context<SourceCodeType>) {
+  const globalSettings: string[] = [];
+
+  try {
+    const path = join(
+      await findRoot(context.file.uri, context.fileExists),
+      'config/settings_schema.json',
+    );
+    const settingsFile = await context.fs.readFile(path);
+    const settings = parseJSON(settingsFile);
```
six-shopify, line 191: Unsure if there's a better way to do this

.changeset/sixty-clouds-confess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/theme-check-common': minor
+---
+
+New check: ValidVisibleIf
```
albchu, line 5: Nit: Theres a preexisting convention for changelog message structure when introducing new checks: https://github.com/Shopify/theme-tools/blob/main/packages/theme-check-common/CHANGELOG.md#minor-changes-6

packages/theme-check-common/src/checks/valid-visible-if/index.ts
albchu: I think this is good, but to be extra safe, @charlespwd WYT?

albchu (commented):
Added some comments after initial code review. Really fantastic stuff Six! Just some minor nits.

I will do a more thorough tophatting shortly.

packages/theme-check-common/src/checks/valid-visible-if/index.ts
```diff
@@ -0,0 +1,296 @@
...
+            report(`Invalid visible_if: can't refer to "block" when not in a block file.`, lookup);
+          } else {
+            report(validateLookup(lookup, vars), lookup);
+          }
+        }
+      },
+    };
+  },
+};
+
```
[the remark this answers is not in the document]
six-shopify: done!

albchu:
This was an unexpected error message based on the way I broke the liquid expression's boolean operator. Any ideas why this is the case and is there a reasonably straightforward way to improve the error messaging in this kind of case?
image

Also I noticed that the validation doesnt really care about whitespace so something like this `"{{block.settings.image_ratio=='custom'}}"` is considered valid from theme check's perspective. Is this simply down to quirks with our theme-tools liquid parsing?

albchu (approved):
I have some clarifying questions based on my tophatting but otherwise I am very happy with the implementation here. See what CP says but Im otherwise completely satisfied with whats here. Great job Six!

packages/theme-check-common/src/checks/valid-visible-if/visible-if-utils.ts
```diff
@@ -0,0 +1,139 @@
...
+        warning: `visible_if expression contains no references to any settings. This is likely an error.`,
+      };
+    }
+
+    return vars;
+  } catch (error) {
+    if (error instanceof SyntaxError) {
+      // Because of our hackish approach, the underlying error is likely to
+      // include an incorrect range and/or mention {% if %} tags. Squelch the
+      // details and just report that it's a simple syntax error.
```
albchu, line 88: [Context and alignment with CP](https://shopify.slack.com/archives/C087FDXMAUB/p1739455202244939)

packages/theme-check-common/src/checks/valid-visible-if/visible-if-utils.ts
```diff
@@ -0,0 +1,139 @@
...
+    // ignore absent or malformed settings schema
+  }
+
+  return globalSettings;
+}
+
+export function getVariableLookupsInExpression(
+  expression: string,
+): LiquidVariableLookup[] | { warning: string } {
+  // parsers other than liquidjs don't yet support expressions in {{ variable }}
```
graygilmore, line 45: We have a few references in here to hacky/gnarly approach but it doesn't look like we explicitly say what that approach is. Considering what we're doing to make this work I think we should have a pretty robust comment explaining the approach because the next person to poke at this without context won't have much of an idea of what's happening.

packages/theme-check-common/src/checks/valid-visible-if/visible-if-utils.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  // note this is the reverse of the above similar-looking case
+  return `Invalid variable: "${poppedSegments.join(
+    '.',
+  )}" refers to a namespace, but is being used here as a variable.`;
+}
+
+function getNormalizedLookups(lookup: LiquidVariableLookup) {
+  const nestedLookups = lookup.lookups.map((l) => {
```
graygilmore, line 131: (nit) prefer named variables here instead of `1`

graygilmore (approved):
No 🎩.

I think everything in here makes sense but it might be good to get one more set of eyes.

packages/theme-check-common/src/checks/valid-visible-if/visible-if-utils.ts
graygilmore: perfect!

packages/theme-check-node/configs/recommended.yml
```diff
@@ -126,6 +126,9 @@ ValidSettingsKey:
 ValidStaticBlockType:
   enabled: true
   severity: 0
+ValidVisibleIf:
```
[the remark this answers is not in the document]
aswamy: As long as we document that in shopify-dev that disabling this disables both, we should be good 👍

aswamy (approved):

PR #801, opened 2025-02-19
Add hover support for the description tag

EvilGenius13:

Closes https://github.com/Shopify/developer-tools-team/issues/505

Add hover support for the `@description` tag when hovering over a snippet & supporting tests.

Example Output:
image

## Before you deploy

### DIFF
--- a/packages/theme-language-server-common/src/hover/providers/RenderSnippetHoverProvider.ts
+++ b/packages/theme-language-server-common/src/hover/providers/RenderSnippetHoverProvider.ts
@@ -54,6 +54,11 @@ export class RenderSnippetHoverProvider implements BaseHoverProvider {
       parts.push('', '**Parameters:**', parameters);
     }
 
+    if (liquidDoc.description) {
+      const description = liquidDoc.description.content.trim();
+      parts.push('', '**Description:**', description);
+    }
+
     if (liquidDoc.examples?.length) {
       const examples = liquidDoc.examples
         ?.map(({ content }) => `\`\`\`liquid${content}\`\`\``)
--- a/packages/theme-language-server-common/src/hover/providers/RenderSnippetHoverProvider.ts
+++ b/packages/theme-language-server-common/src/hover/providers/RenderSnippetHoverProvider.ts
@@ -51,7 +51,7 @@ export class RenderSnippetHoverProvider implements BaseHoverProvider {
 
     if (liquidDoc.description) {
       const description = liquidDoc.description.content;
-      parts.push('', '**Description:**', description);
+      parts.push('', '**Description:**', '\n', description);
     }
 
     if (liquidDoc.parameters?.length) {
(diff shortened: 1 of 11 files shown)


packages/theme-language-server-common/src/hover/providers/RenderSnippetHoverProvider.ts
```diff
@@ -54,6 +54,11 @@ export class RenderSnippetHoverProvider implements BaseHoverProvider {
       parts.push('', '**Parameters:**', parameters);
     }
 
+    if (liquidDoc.description) {
+      const description = liquidDoc.description.content.trim();
+      parts.push('', '**Description:**', description);
+    }
```
aswamy, line 60: Do you or @jamesmengo have opinions on whether A) description should be after parameters, B) before, or C) in the order it's written?

I am soft-leaning towards A (since liquid doc code-completion in typescript usually likes description, params, and return).

At the same time, they also _allow_ C if you want to reorder things around 🤷. Not a strong opinion on the choices, but just wondering what everyone is thinking here.

image

packages/theme-language-server-common/src/hover/providers/RenderSnippetHoverProvider.ts
```diff
@@ -54,6 +54,11 @@ export class RenderSnippetHoverProvider implements BaseHoverProvider {
       parts.push('', '**Parameters:**', parameters);
     }
 
+    if (liquidDoc.description) {
+      const description = liquidDoc.description.content.trim();
```
aswamy, line 58: I don't like trimming it here. The content should come trimmed at an earlier stage.

EvilGenius13: You are absolutely right. I will change that.

packages/theme-language-server-common/src/hover/providers/RenderSnippetHoverProvider.ts
```diff
@@ -49,6 +49,11 @@ export class RenderSnippetHoverProvider implements BaseHoverProvider {
 
     const parts = [`### ${snippetDefinition.name}`];
 
+    if (liquidDoc.description) {
+      const description = liquidDoc.description.content;
+      parts.push('', '**Description:**', description);
```
aswamy, line 54:
```suggestion
      parts.push('', '**Description:**', '\n', description);
```

aswamy (approved):
Looks good! 1 optional suggestion.

packages/theme-language-server-common/src/hover/providers/RenderSnippetHoverProvider.ts
EvilGenius13: Agreed, that looks much better. I'll make that change

PR #802, opened 2025-02-20
Add padding for liquid doc content

aswamy:

Add padding between liquid doc nodes
- We will NOT rearrange the nodes
- We will only add padding between nodes that aren't similar
- fixed a typo in stage-1

## Tophat
- Have a liquid doc tag in a snippet
- Have a few params, examples, and a description
- Place a few params before the examples, and a few after
- Save the file to trigger prettier

Edge cases:
- No content in liquid doc
- 1 node inside liquid doc 

## Before you deploy

### DIFF
--- a/.changeset/hungry-socks-report.md
+++ b/.changeset/hungry-socks-report.md
@@ -1,0 +1,5 @@
+---
+'@shopify/prettier-plugin-liquid': patch
+---
+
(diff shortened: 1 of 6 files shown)


EvilGenius13 (approved):
Thanks for the changes @aswamy! 🎩 went well.

PR #806, opened 2025-02-20
Fix local block validations

miazbikowski:
We've introduced the ability to add a `name` on the preset blocks.

There exists a theme check that ensures local blocks are valid. It was considering ANY block with a `name` to be a local block (except for static blocks, see below), when it should not be taking preset blocks into account for this.

The bug before the fix:
image

It looks like we had some erroneous tests around this erroneous behaviour, so I cleaned those up.

SECONDARILY, this check considered static blocks (which are only in presets) with names to be erroneous, when really we do support static blocks in the preset to support names now, so I've removed the checking around that.

## Follow-up

Update the erroneous theme check docs that show preset blocks with names as being wrong, as well as static blocks with names: https://github.com/Shopify/shopify-dev/pull/53848

## Before you deploy

### DIFF
--- a/.changeset/twenty-maps-drum.md
+++ b/.changeset/twenty-maps-drum.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-check-common': patch
+---
+
+Fixed local blocks validation to not consider preset blocks as containing local blocks when a name is present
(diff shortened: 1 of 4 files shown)


albchu (approved):
LGTM

graygilmore (approved):
I don't have too much context on this particular feature but the code makes sense to me!

PR #808, opened 2025-02-21
New `UnusedDocParam` theme-check rule

aswamy:

Closes https://github.com/Shopify/theme-tools/issues/803
- Add check to ensure `@param` variable is actually used in the snippet

## Tophat
- Have a snippet with doc
- Ensure the params defined are actually used
- We will consider it "unused" if the variable is actually a for-loop scoped variable
- You can hover over the warning, and press quick fix to remove the param

e.g.
```
{% doc %}
  @param {string} name - This is a name.
  @param {object} opts - options // Unused
  @param {unknown} extra - extra data // Unused
{% enddoc %}

{{ name }}

{% tablerow extra in product.variants %}
  {{ extra }}
{% endtablerow %}

{% for extra in product.variants %}
  {{ extra }}
{% endfor %}
```

## Before you deploy

### DIFF
--- a/.changeset/dull-badgers-sing.md
+++ b/.changeset/dull-badgers-sing.md
@@ -1,0 +1,8 @@
+---
+'@shopify/theme-check-common': minor
+'@shopify/theme-check-node': minor
+---
+
+New `UnusedDocParam` theme-check rule
+
+- Theme check will verify if variable defined in `@param` tag within a snippet's `doc` is used
--- a/packages/theme-check-common/src/checks/unused-doc-param/index.ts
+++ b/packages/theme-check-common/src/checks/unused-doc-param/index.ts
@@ -1,0 +1,59 @@
+import { LiquidDocParamNode, NodeTypes } from '@shopify/liquid-html-parser';
+import { LiquidCheckDefinition, Severity, SourceCodeType } from '../../types';
+import { isLoopScopedVariable } from '../utils';
+
+export const UnusedDocParam: LiquidCheckDefinition = {
+  meta: {
+    code: 'UnusedDocParam',
+    name: 'Prevent unused doc parameters',
+    docs: {
+      description:
+        'This check exists to ensure any parameters defined in the `doc` tag are used within the snippet.',
+      recommended: true,
+      url: 'https://shopify.dev/docs/storefronts/themes/tools/theme-check/checks/unused-doc-param',
+    },
+    type: SourceCodeType.LiquidHtml,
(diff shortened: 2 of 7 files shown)


.changeset/dull-badgers-sing.md
```diff
@@ -0,0 +1,8 @@
+---
+'@shopify/theme-check-common': minor
+'@shopify/theme-check-node': minor
+---
+
+New `UnusedDocParam` theme-check rule
+
+- Theme check will verify if variable defined in `@param` tag within a snippet's `doc` is used
```
jamesmengo, line 8:
```suggestion
- Theme check will verify that parameters defined within a snippet's `doc` header are used within the body
```

packages/theme-check-common/src/checks/utils.ts
```diff
@@ -85,3 +86,13 @@ export function hasAttributeValueOf(attr: ValuedHtmlAttribute, value: string) {
 export function isLiquidString(node: LiquidHtmlNode): node is NodeOfType<NodeTypes.String> {
   return node.type === NodeTypes.String;
 }
+
+export function isLoopScopedVariable(variableName: string, ancestors: LiquidHtmlNode[]) {
```
jamesmengo, line 90:
```suggestion
export function isLocallyScopedVariable(variableName: string, ancestors: LiquidHtmlNode[]) {
```

packages/theme-check-common/src/checks/unused-doc-param/index.ts
```diff
@@ -0,0 +1,59 @@
...
+      async onCodePathEnd() {
+        for (const [variable, node] of definedLiquidDocParams.entries()) {
+          if (!usedVariables.has(variable)) {
+            context.report({
+              message: `The parameter '${variable}' is defined but not used in this file.`,
+              startIndex: node.position.start,
+              endIndex: node.position.end,
+              suggest: [
+                {
+                  message: `Remove the unused doc parameter '${variable}'`,
```
jamesmengo, line 49: WDYT about something like: 
```suggestion
                  message: `Remove unused parameter '${variable}'`,
```
or 

`Remove unused parameter '${variable}' from `doc` header`

packages/theme-check-common/src/checks/utils.ts
aswamy: I didn't choose "locally scoped" because i don't check other scopes outside loop like `capture` tags

jamesmengo (approved):

PR #809, opened 2025-02-21
Doc tag completion fix and new theme check for doc tag usage

aswamy:

Closes https://github.com/Shopify/developer-tools-team/issues/572
commit 1: prevents `doc` tag completions outside snippet file
commit 2: theme check to prevent `doc` tag usage outside snippet files

## Tophat
- Add a `doc` tag inside a snippet file
- Make sure code completion still works for `@param` tag and `@param` tag types
- Add a `doc` tag a section/block file
- You will get a theme check error
- Follow the quick suggestion to remove the tag

## Before you deploy
commit 2:

commit 1:

### DIFF
--- a/packages/theme-check-common/src/checks/unsupported-doc-tag/index.ts
+++ b/packages/theme-check-common/src/checks/unsupported-doc-tag/index.ts
@@ -1,0 +1,45 @@
+import { LiquidCheckDefinition, Severity, SourceCodeType } from '../../types';
+import { dirname } from '../../path';
+
+export const UnsupportedDocTag: LiquidCheckDefinition = {
+  meta: {
+    code: 'UnsupportedDocTag',
+    name: 'Prevent unsupported doc tag usage',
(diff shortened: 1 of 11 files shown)


packages/theme-check-common/src/checks/unsupported-doc-tag/index.ts
```diff
@@ -0,0 +1,45 @@
...
+      return {};
+    }
+
+    return {
+      async LiquidRawTag(node) {
+        if (node.name !== docTagName) {
+          return;
+        }
+        context.report({
+          message: `The \`${docTagName}\` tag should only be used within a snippet.`,
```
jamesmengo, line 32:
```suggestion
          message: `The \`${docTagName}\` tag can only be used within a snippet.`,
```

jamesmengo (approved):
Left some non-blocking comments! Let's make sure to discuss them :)

PR #820, opened 2025-02-25
[LiquidDoc] Modify LiquidDoc parsing to capture leading whitespace and add inline flag

jamesmengo:
Part of https://github.com/Shopify/developer-tools-team/issues/573

**[‼️] The tests on this branch are broken, but I fix them in the [PR above](https://app.graphite.dev/github/pr/Shopify/theme-tools/824/%5BLiquidDoc%5D-Update-Prettier-%2B-Formatting-behaviour-for-descriptions-and-examples). I'm splitting these up for the sake of review, but I will ship them together in a stack.**

Before this change, examples and description nodes had similar types of content, but were being treated differently in stage 1.

This change makes the parsing consistent between examples and descriptions by:
-  **removing the trimming we do for the description content** in `stage-1`.
- assigning an `isInline` property to examples and descriptions. I consume this in the next PR in the stack.

```
@description inline means it's on the same line
@description 
otherwise it can be like this
```

- Update hover, prettier to work with this new parsing behaviour

## Before you deploy

### DIFF
--- a/packages/liquid-html-parser/grammar/liquid-html.ohm
+++ b/packages/liquid-html-parser/grammar/liquid-html.ohm
@@ -409,7 +409,7 @@ LiquidDoc <: Helpers {
   // We use this as an escape hatch to stop matching TextNode and try again when one of these characters is encountered
   openControl:= "@" | end
 
-  descriptionNode = "@description" strictSpace* descriptionContent
+  descriptionNode = "@description" space* descriptionContent
 
   paramNode = "@param" strictSpace* paramType? strictSpace* (optionalParamName | paramName) (strictSpace* "-")? strictSpace* paramDescription
   paramType = "{" strictSpace* paramTypeContent strictSpace* "}"
@@ -422,7 +422,7 @@ LiquidDoc <: Helpers {
   paramDescription = (~"]" anyExceptStar<endOfParam>)
   endOfParam = strictSpace* (newline | end)
 
-  exampleNode = "@example" strictSpace* exampleContent
+  exampleNode = "@example" space* exampleContent
   exampleContent = anyExceptStar<endOfExample>
   endOfExample =  strictSpace* ("@" | end)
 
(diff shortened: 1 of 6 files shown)


jamesmengo:
* **#821** <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/821?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#820** <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/820?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/820?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/liquid-html-parser/src/stage-1-cst.ts
```diff
@@ -1417,16 +1420,11 @@ function toLiquidDocAST(source: string, matchingSource: string, offset: number)
       source,
       content: 2,
       isImplicit: false,
-    },
-    descriptionContent: {
-      type: ConcreteNodeTypes.TextNode,
-      value: function (this: Node) {
-        return this.sourceString.trim();
```
jamesmengo, line 1424: We're no longer trimming here in `ast-1`.

This means that `content.value` will contain leading + trailing whitespace, which we use to determine whether an example / description were provided inline or start on their own line.

```
@description inline means it's on the same line
@description 
otherwise it can be like this
````

File: packages/liquid-html-parser/grammar/liquid-html.ohm [hunk 2]
```diff
@@ -409,7 +409,7 @@ LiquidDoc <: Helpers {
   // We use this as an escape hatch to stop matching TextNode and try again when one of these characters is encountered
   openControl:= "@" | end
 
-  descriptionNode = "@description" strictSpace* descriptionContent
+  descriptionNode = "@description" space* descriptionContent
```
jamesmengo, line 412: Now, we are matching `newlines` as well. This allows us to check for a newline character in index `[1]`.

If it is **not** present, then we know that the description content was provided in line.

aswamy (approved):
> [‼️] The tests on this branch are broken, but I fix them in the [PR above](https://app.graphite.dev/github/pr/Shopify/theme-tools/824/%5BLiquidDoc%5D-Update-Prettier-%2B-Formatting-behaviour-for-descriptions-and-examples). I'm splitting these up for the sake of review, but I will ship them together in a stack.

Why is it in a separate PR instead of a separate commit 😅

PR #824, opened 2025-02-25
Move whitespace trimming for LiquidDoc description and example nodes from `liquid-html-parser` to `theme-check-common`

jamesmengo:
## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/liquid-doc/liquidDoc.spec.ts
+++ b/packages/theme-check-common/src/liquid-doc/liquidDoc.spec.ts
@@ -106,7 +106,7 @@ describe('Unit: getSnippetDefinition', () => {
       liquidDoc: {
         examples: [
           {
-            content: '\n          {{ product }}\n',
+            content: '{{ product }}',
             nodeType: 'example',
           },
         ],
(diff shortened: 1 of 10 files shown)


packages/theme-language-server-common/src/hover/providers/RenderSnippetHoverProvider.ts
```diff
@@ -61,7 +61,7 @@ export class RenderSnippetHoverProvider implements BaseHoverProvider {
 
     if (liquidDoc.examples?.length) {
       const examples = liquidDoc.examples
-        ?.map(({ content }) => `\`\`\`liquid${content}\`\`\``)
+        ?.map(({ content }) => `\`\`\`liquid\n${content}\n\`\`\``)
```
jamesmengo, line 64: This was coming for 'free', but was more as a side effect of:
1) forcing examples to be on their own line.
2) including leading whitespaces in `content.value`

This made it really hard to have consistent behavior between descriptions and examples.

packages/theme-check-common/src/liquid-doc/liquidDoc.spec.ts
```diff
@@ -106,7 +106,7 @@ describe('Unit: getSnippetDefinition', () => {
       liquidDoc: {
         examples: [
           {
-            content: '\n          {{ product }}\n',
```
jamesmengo, line 109: We intentionally were capturing this in the past to defer this decision until now since we reckoned that it would be easier to remove this than to add it back.

aswamy (approved):

PR #834, opened 2025-02-28
[LiquidDoc][Theme Check] Check for duplicated params passed to `{% render %}` tag

jamesmengo:

Closes https://github.com/Shopify/developer-tools-team/issues/539

Add a theme check to check for duplicated param passed into a `{% render %}` tag

**Includes:**
- **Reporting** duplicated params
- **Suggesting** fixes to remove -> We will always keep the first occurrence and suggest removal for subsequent occurrences
- Support for **aliases** `{% render 'snippet-name' with 'test' as param, param: 'test' %}`

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/checks/duplicate-render-snippet-params/index.ts
+++ b/packages/theme-check-common/src/checks/duplicate-render-snippet-params/index.ts
@@ -1,0 +1,87 @@
+import { LiquidCheckDefinition, Severity, SourceCodeType } from '../../types';
+import { LiquidNamedArgument, RenderMarkup } from '@shopify/liquid-html-parser';
+import { isLiquidString } from '../utils';
+
+export const DuplicateRenderSnippetParams: LiquidCheckDefinition = {
+  meta: {
+    code: 'DuplicateRenderSnippetParams',
+    name: 'Duplicate Render Snippet Parameters',
+    docs: {
+      description:
+        'This check ensures that no duplicate parameter names are provided when rendering a snippet.',
+      recommended: true,
+      url: 'https://shopify.dev/docs/storefronts/themes/tools/theme-check/checks/duplicate-render-snippet-params',
+    },
+    type: SourceCodeType.LiquidHtml,
+    severity: Severity.WARNING,
+    schema: {},
+    targets: [],
+  },
+
+  create(context) {
+    return {
+      async RenderMarkup(node: RenderMarkup) {
+        if (!isLiquidString(node.snippet)) {
+          return;
+        }
+
+        const snippetName = node.snippet.value;
+
+        const encounteredParams = new Set<string>();
+        for (const param of node.args) {
+          const paramName = param.name;
+          if (encounteredParams.has(paramName) || paramName === node.alias) {
+            context.report({
+              message: `Duplicate parameter '${paramName}' in render tag for snippet '${snippetName}'.`,
+              startIndex: param.position.start,
+              endIndex: param.position.end,
+              suggest: [
+                {
+                  message: `Remove duplicate parameter '${paramName}'`,
+                  fix: (fixer) => {
+                    // This parameter removal logic is duplicated in UnrecognizedRenderSnippetParams
+                    // Consider extracting to a shared utility or simplifying the removal approach in the parsing steps.
+                    // I chose not to do so here as I would like more examples to see how this should be done.
+                    const sourceBeforeArg = node.source.slice(0, param.position.start);
+                    const matches = sourceBeforeArg.match(/,\s*/g);
+                    const lastWhitespaceMatch = matches ? matches[matches.length - 1] : null;
+                    let startPos = lastWhitespaceMatch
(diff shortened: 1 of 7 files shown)


jamesmengo:
* **#834** <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/834?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/834?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/theme-check-common/src/checks/duplicate-render-snippet-params/index.ts
```diff
@@ -0,0 +1,87 @@
...
+          const paramName = param.name;
+          if (encounteredParams.has(paramName) || paramName === node.alias) {
+            context.report({
+              message: `Duplicate parameter '${paramName}' in render tag for snippet '${snippetName}'.`,
+              startIndex: param.position.start,
+              endIndex: param.position.end,
+              suggest: [
+                {
+                  message: `Remove duplicate parameter '${paramName}'`,
+                  fix: (fixer) => {
```
jamesmengo, line 41: This part has me feeling icky, I haven't landed on an approach yet. I'm considering many.
I want to be pragmatic here. I don't think this code will change soon, and if it does I'm hoping that we can simplify this logic by potentially getting more information during the parse step instead.

Not sure how much time I want to invest as a result.

**Ideas**
- Leaving this comment here to indicate duplication
- Extracting to a shared utility but testing in the tests for individual checks
- Extracting a utility that just provides indexes

Factors
- The code is 1-1 duplicated, so I do want to call out duplication
- This code IS tested through the individual theme checks.
- I tried extracting this to a utility and didn't like the way I had to write tests. Probably a skill issue. Will take time either way.
- I feel like the better approach is to do this programatically via modifying parsing logic. 



@aswamy WDYT?

packages/theme-check-common/src/checks/duplicate-render-snippet-params/index.ts
```diff
@@ -0,0 +1,87 @@
...
+                    // This parameter removal logic is duplicated in UnrecognizedRenderSnippetParams
+                    // Consider extracting to a shared utility or simplifying the removal approach in the parsing steps.
+                    // I chose not to do so here as I would like more examples to see how this should be done.
+                    const sourceBeforeArg = node.source.slice(0, param.position.start);
+                    const matches = sourceBeforeArg.match(/,\s*/g);
+                    const lastWhitespaceMatch = matches ? matches[matches.length - 1] : null;
```
aswamy, line 47:
```suggestion
                    const lastCommaMatch = matches?.at(-1);
```

packages/theme-check-common/src/checks/duplicate-render-snippet-params/index.ts
```diff
@@ -0,0 +1,87 @@
...
+              startIndex: param.position.start,
+              endIndex: param.position.end,
+              suggest: [
+                {
+                  message: `Remove duplicate parameter '${paramName}'`,
+                  fix: (fixer) => {
+                    // This parameter removal logic is duplicated in UnrecognizedRenderSnippetParams
+                    // Consider extracting to a shared utility or simplifying the removal approach in the parsing steps.
+                    // I chose not to do so here as I would like more examples to see how this should be done.
+                    const sourceBeforeArg = node.source.slice(0, param.position.start);
```
aswamy, line 45:
```suggestion
                    const sourceBeforeArg = node.source.slice(node.position.start, param.position.start);
```
We should probably not use source since it contains the ENTIRE file content before the node.

packages/theme-check-common/src/checks/duplicate-render-snippet-params/index.ts
```diff
@@ -0,0 +1,87 @@
...
+                    const matches = sourceBeforeArg.match(/,\s*/g);
+                    const lastWhitespaceMatch = matches ? matches[matches.length - 1] : null;
+                    let startPos = lastWhitespaceMatch
+                      ? param.position.start - (lastWhitespaceMatch.length - 1)
+                      : param.position.start;
+
+                    if (isLastParam(param)) {
+                      startPos -= 1;
+                    }
```
aswamy, line 54: you should leave a comment on why we're doing this

aswamy (approved):
minor changes

packages/theme-check-common/src/checks/duplicate-render-snippet-params/index.ts
jamesmengo: Paired on this. Not ideal, but we're going to leave this as is as it's the more pragmatic approach.

We should reconsider the next time we touch this code

jamesmengo:
### Merge activity

* **Feb 28, 10:12 PM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/theme-tools/834).

PR #835, opened 2025-02-28
Fix file URI handling for Windows platforms

EvilGenius13:
Part of https://github.com/Shopify/cli/issues/3497

After a long debugging session and some async work with @aswamy, we found where the issue was coming from.

On windows platforms, filepaths look something like this `file:\\\C:\path\to\file`.
- After normalization, it looks like /C:/path/to/file
- When this glob is provided for `asyncGlob` on Windows, it gets confused because of the leading `/`
- We strip that leading `/` if it exists before a path

This implementation checks if the OS is windows before changing the way we change the URI to ensure we don't affect unix based systems. 

## Before you deploy

### DIFF
--- a/packages/theme-check-node/src/index.ts
+++ b/packages/theme-check-node/src/index.ts
@@ -26,6 +26,7 @@ import glob = require('glob');
 import { autofix } from './autofix';
 import { findConfigPath, loadConfig as resolveConfig } from './config';
 import { NodeFileSystem } from './NodeFileSystem';
+import { fileURLToPath } from 'node:url';
 
 const asyncGlob = promisify(glob);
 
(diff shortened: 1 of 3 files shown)


packages/theme-check-node/src/index.ts
```diff
@@ -153,9 +153,19 @@ export async function getTheme(config: Config): Promise<Theme> {
...
-  const normalizedGlob = path
+  let normalizedGlob = path
     .normalize(path.join(config.rootUri.replace(/^file:/, ''), '**/*.{liquid,json}'))
     .replace(/\\/g, '/');
+
+  // The glob is always an absoluate path, so windows paths should always
+  // start with a drive letter (with few exceptions we can ignore).
+  // More info: https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats
+  //
+  // If an absoluate path starts with a slash on Windows, we strip it
```
graygilmore, line 164: Could we add a test for this? Is there a reasonable way to accomplish that?

aswamy: > Could we add a test for this?

Will do!

> Is there a reasonable way to accomplish that?

We can't tophat this in a reasonable way because we have to release theme-tools FIRST then update CLI. To verify this works i installed CLI on my windows machine, then updated its source code to contain this change.

aswamy:
### Force push
- Ended up using a function built into node to convert path to url
- Still added a test even though we are testing a built-in function so i know it works on Windows as well when it is run in our CI

Built in function:
image

PR #845, opened 2025-03-10
Improve alias parsing in the cst & ast

EvilGenius13:
Closes: https://github.com/Shopify/developer-tools-team/issues/607

- Added proper position tracking for aliases by introducing a `RenderAliasExpression` node type.
- Updated the parser to capture positional information for the `as` and `with` part of render tags
- Modified any consumers of render markup to use the new structure

## Before you deploy

### DIFF
--- a/packages/liquid-html-parser/src/stage-1-cst.ts
+++ b/packages/liquid-html-parser/src/stage-1-cst.ts
@@ -379,7 +379,7 @@ export interface ConcreteLiquidTagRenderMarkup
   extends ConcreteBasicNode<ConcreteNodeTypes.RenderMarkup> {
   snippet: ConcreteStringLiteral | ConcreteLiquidVariableLookup;
   variable: ConcreteRenderVariableExpression | null;
-  aliasExpression: ConcreteRenderAliasExpression | null;
+  alias: ConcreteRenderAliasExpression | null;
   renderArguments: ConcreteLiquidNamedArgument[];
 }
 
@@ -391,7 +391,7 @@ export interface ConcreteRenderVariableExpression
 
 export interface ConcreteRenderAliasExpression
   extends ConcreteBasicNode<ConcreteNodeTypes.RenderAliasExpression> {
-  alias: string;
+  value: string;
 }
 
 export interface ConcreteLiquidVariableOutput
@@ -873,7 +873,7 @@ function toCST<T>(
       type: ConcreteNodeTypes.RenderMarkup,
       snippet: 0,
       variable: 1,
-      aliasExpression: 2,
+      alias: 2,
       renderArguments: 3,
       locStart,
       locEnd,
@@ -908,7 +908,7 @@ function toCST<T>(
     },
     renderAliasExpression: {
       type: ConcreteNodeTypes.RenderAliasExpression,
-      alias: 3,
+      value: 3,
       locStart,
       locEnd,
       source,
--- a/packages/liquid-html-parser/src/stage-1-cst.ts
+++ b/packages/liquid-html-parser/src/stage-1-cst.ts
@@ -82,6 +82,7 @@ export enum ConcreteNodeTypes {
   RenderMarkup = 'RenderMarkup',
   PaginateMarkup = 'PaginateMarkup',
   RenderVariableExpression = 'RenderVariableExpression',
+  RenderAliasExpression = 'RenderAliasExpression',
   ContentForNamedArgument = 'ContentForNamedArgument',
 
   LiquidDocParamNode = 'LiquidDocParamNode',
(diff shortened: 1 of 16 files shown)


packages/liquid-html-parser/src/stage-1-cst.ts
```diff
@@ -377,8 +378,8 @@ export interface ConcreteLiquidTagContentForMarkup
 export interface ConcreteLiquidTagRenderMarkup
   extends ConcreteBasicNode<ConcreteNodeTypes.RenderMarkup> {
   snippet: ConcreteStringLiteral | ConcreteLiquidVariableLookup;
-  alias: string | null;
   variable: ConcreteRenderVariableExpression | null;
+  aliasExpression: ConcreteRenderAliasExpression | null;
```
jamesmengo, line 382: can we keep this as-is?

packages/prettier-plugin-liquid/src/printer/print/liquid.ts
```diff
@@ -183,7 +183,7 @@ function printNamedLiquidBlockStart(
     case NamedTags.render: {
       const markup = node.markup;
       const trailingWhitespace =
-        markup.args.length > 0 || (markup.variable && markup.alias) ? line : ' ';
+        markup.args.length > 0 || (markup.variable && markup.aliasExpression?.alias) ? line : ' ';
```
jamesmengo, line 186: Does this condition need to be updated?

When would `markup.aliasExpression` be `truthy` but  `markup.aliasExpression?.alias` be `falsy`?

jamesmengo:
This is going in the right direction! Left some questions / comments :)

packages/prettier-plugin-liquid/src/printer/print/liquid.ts
EvilGenius13: Yeah, you're right. I was digging deeper than we needed.

packages/theme-check-common/src/checks/valid-render-snippet-param-types/index.ts
```diff
@@ -126,11 +126,13 @@ export const ValidRenderSnippetParamTypes: LiquidCheckDefinition = {
       liquidDocParameters: Map<string, LiquidDocParameter>,
     ) {
       if (
-        node.alias &&
+        node.alias?.value &&
```
jamesmengo, line 129: Can we just check if `node.alias` is not null?

packages/theme-check-common/src/checks/unrecognized-render-snippet-params/index.ts
```diff
@@ -73,36 +73,33 @@ export const UnrecognizedRenderSnippetParams: LiquidCheckDefinition = {
...
-            },
-          });
+        if (
+          source.slice(startIndex, node.alias.position.start).includes(' with ') ||
+          source.slice(startIndex, node.alias.position.start).includes(' for ')
+        ) {
+          startIndex = node.variable.position.start + 1;
+        } else {
+          startIndex = node.alias.position.start;
         }
```
jamesmengo, line 87: What does this case do? Can you maybe add a test to illustrate that?

packages/theme-check-common/src/checks/unrecognized-render-snippet-params/index.ts
```diff
@@ -73,36 +73,33 @@ export const UnrecognizedRenderSnippetParams: LiquidCheckDefinition = {
       liquidDocParameters: Map<string, LiquidDocParameter>,
       snippetName: string,
     ) {
-      if (node.alias && !liquidDocParameters.has(node.alias) && node.variable) {
-        const asAliasMatch = node.source
-          .slice(node.variable.position.end, node.position.end)
-          .match(new RegExp(`\\s+as\\s+${node.alias}`));
+      if (node.alias?.value && !liquidDocParameters.has(node.alias.value) && node.variable) {
```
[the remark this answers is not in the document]
EvilGenius13: I like that, much more clean.

jamesmengo (approved):

packages/liquid-html-parser/src/stage-2-ast.ts
```diff
@@ -427,6 +429,12 @@ export interface RenderVariableExpression extends ASTNode<NodeTypes.RenderVariab
   name: LiquidExpression;
 }
 
+/** Represents the `as name` expressions in render nodes */
+export interface RenderAliasExpression extends ASTNode<NodeTypes.RenderAliasExpression> {
+  /** {% render 'snippet' as name %} */
```
aswamy, line 434: This is supposed to be `{% render 'snippet' for array as name %}` or `{% render 'snippet' with object as name %}` right?

https://shopify.dev/docs/api/liquid/tags/render

aswamy (approved):
Amazing work! HUGE improvement! Tophat went well!

- Tested out param renaming and it affected alias params
- Tested out duplicate params (where one was alias) and works
- Tested out unknown alias param
- Tested out incorrect param type with alias param

packages/liquid-html-parser/src/stage-2-ast.ts
EvilGenius13: Yeah you're right. I'll edit that!

PR #851, opened 2025-03-12
Add CSS LSP for `stylesheet` tags within liquid

JoviDeCroock:

This adds support for generating auto-complete information in `{% stylesheet %}` tags. Currently this only focusses on the current document and doesn't really leverage a way to "learn" from existing `.css` files. For that we'd have to watch `.css` files and add them to our CSS LSP. This is entirely possible but didn't want to add too many changes at once. First gauging interest for this particular feature.

This approach basically adds an additional data-point to our liquid document about the stylesheet tag we found, when we have a stylesheet and the cursor is in range of the stylesheet we choose to parse it. We then retrieve hover/auto-complete information to return it to the user.

I have tested this out locally and it works great!

- Learn from `.css` files and other `.liquid` stylesheet templates
  - You can look at `doComplete2` and how they leverage a `filesystem` abstraction
- We could even add support for `{% javascript %}` which would have to learn from `.ts` and `.js` files.

## Before you deploy

    - [x] If applicable, I've updated the `theme-app-extension.yml` config
  - [x] I've made a PR to update the [shopify.dev theme check docs](https://github.com/Shopify/shopify-dev/tree/main/content/storefronts/themes/tools/theme-check/checks) if applicable (link PR here).

### DIFF
--- a/packages/theme-language-server-common/src/css/CSSLanguageService.ts
+++ b/packages/theme-language-server-common/src/css/CSSLanguageService.ts
@@ -95,10 +95,9 @@ export class CSSLanguageService {
         // Which means that the completions will be at the same line number in the Liquid document
         const stylesheetString =
           Array(schemaLineNumber).fill('\n').join('') +
-          stylesheetTag.source.slice(
-            stylesheetTag.blockStartPosition.end,
-            stylesheetTag.blockEndPosition.start,
-          );
+          stylesheetTag.source
+            .slice(stylesheetTag.blockStartPosition.end, stylesheetTag.blockEndPosition.start)
+            .replace(/\n$/, ''); // Remove trailing newline so parsing errors don't show up on `{% endstylesheet %}`
         const stylesheetTextDocument = TextDocument.create(
           textDocument.uri,
           'json',
--- a/packages/theme-language-server-common/src/diagnostics/runChecks.ts
+++ b/packages/theme-language-server-common/src/diagnostics/runChecks.ts
@@ -1,19 +1,19 @@
 import {
-  ThemeBlockSchema,
   check,
   findRoot,
   makeFileExists,
+  Offense,
   path,
   SectionSchema,
-  SourceCodeType,
-  Offense,
   Severity,
+  SourceCodeType,
+  ThemeBlockSchema,
 } from '@shopify/theme-check-common';
 
-import { DocumentManager } from '../documents';
+import { CSSLanguageService } from '../css/CSSLanguageService';
+import { AugmentedSourceCode, DocumentManager } from '../documents';
 import { Dependencies } from '../types';
 import { DiagnosticsManager } from './DiagnosticsManager';
-import { CSSLanguageService } from '../css/CSSLanguageService';
 import { offenseSeverity } from './offenseToDiagnostic';
 
 export function makeRunChecks(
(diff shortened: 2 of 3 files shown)


charlespwd (commented):
Looks a whole lot like what we do for JSON inside `{% schema %}`, if you look at [startServer](https://github.com/Shopify/theme-tools/blob/main/packages/theme-language-server-common/src/server/startServer.ts#L472-L478), you can see we do something like jsonLanguageService.complete(params) or something. Maybe that'd be simpler? 

Note that either approach only works if `{% stylesheet %}` doesn't accept liquid though. If it does, we're kind of screwed since any kind of Liquid would break the CSS parser.

packages/theme-language-server-common/src/documents/DocumentManager.ts
```diff
@@ -181,10 +181,26 @@ export class DocumentManager {
           textDocument,
         };
       case SourceCodeType.LiquidHtml:
+        const stylesheet = sourceCode.source.match(
+          /{% stylesheet %}([\s\S]*?){% endstylesheet %}/,
```
charlespwd, line 185: This feels a bit brittle? Maybe the AST visitor route would be better? Folks forget their spaces all the time.

packages/theme-language-server-common/src/css/CSSLanguageService.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+            }
+          {% endstylesheet %}
+          <div>hello world</div>
+
+        `,
+      );
+
+      const completions = await cssLanguageService.completions(params);
+      assert(isCompletionList(completions));
+      expect(completions.items).to.have.lengthOf(357);
```
JoviDeCroock, line 58: This looks to be normal when I compare it to other implementations, I asserted that the top one is the relevant item here 😅

packages/theme-language-server-common/src/css/CSSLanguageService.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  async setup(clientCapabilities: LSPClientCapabilities) {
+    await Promise.all(
+      Modes.map(async (mode) => {
+        const service = getCSSLanguageService({
+          clientCapabilities,
+        });
+
+        this.services[mode] = service;
```
charlespwd, line 34: It's pretty unclear, ngl. The mode thing is for switching between different JSON schemas between theme app extensions JSON schemas for the `{% schema %}` tag inside block files. I don't think we need a mode switch for theme app extensions? 

The TL;DR is that theme app extensions look a bit like themes
```
assets/
blocks/
snippets/
```

But, unfortunately, the `{% schema %}` JSON code that they accept isn't 100% the same as the ones we use for themes. And since we wanted to have matching of file patterns -> JSON schemas, we needed the mode switch to target `blocks/*.liquid` files with different JSON schemas. 

I don't think we need that for CSS or JS.

packages/theme-language-server-common/src/css/CSSLanguageService.ts
```diff
@@ -0,0 +1,129 @@
...
+        const textDocument = document.textDocument;
+        let offset = 0;
+        let isDiagnostics = false;
+        if ('position' in params && params.position.line !== 0 && params.position.character !== 0) {
+          offset = textDocument.offsetAt(params.position);
+        } else {
+          const stylesheetIndex = document.source.indexOf('{% stylesheet %}');
+          console.log(stylesheetIndex);
+          offset = stylesheetIndex;
+          isDiagnostics = true;
```
charlespwd, line 92: Why not use the visitor option?
```ts
const stylesheet = ancestors.find(
  (node): node is LiquidRawTag =>
    node.type === NodeTypes.LiquidRawTag && node.name === 'stylesheet',
);
```

I _believe_ the stylesheet tag is also of NodeType.LiquidRawTag so that the prettier plugin can switch to the css embed mode for those.

charlespwd: oh derp two lines down. I should read lol

packages/theme-language-server-common/src/css/CSSLanguageService.ts
JoviDeCroock: Removed it

yarn.lock
```diff
@@ -6717,12 +6727,17 @@ vscode-languageserver-textdocument@^1.0.11, vscode-languageserver-textdocument@^
   resolved "https://registry.npmjs.org/vscode-languageserver-textdocument/-/vscode-languageserver-textdocument-1.0.11.tgz#0822a000e7d4dc083312580d7575fe9e3ba2e2bf"
   integrity sha512-X+8T3GoiwTVlJbicx/sIAF+yuJAqz8VvwJyoMVhwEMoEKE/fkDmrqUgDMyBECcM2A2frVZIUj5HI/ErRXCfOeA==
 
+vscode-languageserver-textdocument@^1.0.12:
```
charlespwd, line 6730: Think we can yarn-deduplicate this one? We do use ls-texdocument

.changeset/gentle-beds-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/theme-language-server-common': minor
```
charlespwd, line 2: We could add theme-check-vscode: minor here so that this changeset appears in the VS Code extension marketplace changelog

yarn.lock
JoviDeCroock: I tried deduping but... there was a lot more getting deduped 😅

mxstbr:
Yess!! Let's :shipit:

benjaminsehl:
`stylesheet` doesn't accept Liquid — I actually would love if it could, but we can for sure ignore that for a long time!

hyped for this!

packages/theme-language-server-common/src/diagnostics/runChecks.ts
```diff
@@ -82,3 +125,18 @@ export function makeRunChecks(
...
+  // Sum the lengths of all previous lines (plus newline characters)
+  let offset = 0;
+  for (let i = 0; i < position.line; i++) {
+    offset += lines[i].length + 1; // +1 for the newline character
+  }
+
+  // Add the character offset in the current line
+  offset += position.character;
+
+  return offset;
```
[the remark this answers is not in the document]
charlespwd: Also IIRC worries about \r\n for us

adamchipperfield:
Are there plans to add support for the `{% style %}` tag with this?

graygilmore:
> Are there plans to add support for the `{% style %}` tag with this?

Not at this time. The problem is that when you're inside a `style` tag you're writing Liquid and CSS parsing fails under that scenario. We've talked about it a bit but have no plans to start building a whole new parser for it at this time.

PR #859, opened 2025-03-18
Introduce fixer for invalid conditions

karreiro:
🚧


packages/theme-check-common/src/checks/invalid-comparison-syntax/index.ts
```diff
@@ -0,0 +1,64 @@
...
+            });
+          }
+        }
+      },
+    };
+  },
+};
+
+function isValidComparisonConnector(token: string): boolean {
+  const validConnectors = ['and', 'or', '%}', 'contains', 'startswith', 'endswith'];
```
ianks, line 62: `startswith` and `endswith` are not operators in liquid AFAIK

karreiro: Thank you, @ianks!

PR #861, opened 2025-03-19
Fix bug where theme check would fail for theme app extensions

graygilmore:

- Fixes https://github.com/Shopify/cli/issues/5269#issuecomment-2720241078

This bug only surfaces when a developer would manually run the theme check against a theme app extension from the CLI. Inside the VSCode extension we were already doing the right thing so you wouldn't notice it there and the Shopify CLI hardcodes the check when you build your app so you wouldn't notice it there either.

This adds some logic to where we load the config to check if we're inside of an app by looking for two app-like config files. If so, we use `theme-check:theme-app-extension` as our default checks rather than the recommended. This ensures that our context is appropriately set as 'app' instead of 'theme' and removes false positives when running the check.

### DIFF
--- a/.changeset/slow-bananas-sing.md
+++ b/.changeset/slow-bananas-sing.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-check-node': patch
+---
+
+Fix false positives when running theme check on theme app extensions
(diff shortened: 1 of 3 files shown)


jamesmengo (approved):

charlespwd (approved):
TY :D

PR #874, opened 2025-04-02
Aswamy/support liquid doc complex types

aswamy:

Support complex types for liquid doc param

## Before you deploy


packages/theme-check-docs-updater/src/themeLiquidDocsManager.ts
```diff
@@ -36,6 +36,10 @@ export class ThemeLiquidDocsManager implements ThemeDocset, JsonValidationSet {
     return findSuitableResource(this.loaders('objects'), JSON.parse, [], this.log);
   });
 
+  liquidDrops = memo(async (): Promise<ObjectEntry[]> => {
+    return findSuitableResource(this.loaders('objects'), JSON.parse, [], this.log);
+  });
```
karreiro, line 41: I've noticed that `liquidDrops` shares the same implementation as `objects`. May you share extra context on this?

I believe the purpose of `theme-check-docs-updater` is to get up-to-date resources for us, so we may process and use those resources in other packages.

I'm wondering if we could reuse existing resources at the `theme-check-common` and `theme-language-server-common` levels. Please, let me know if I'm missing something something here tho 🙇

karreiro (commented):
Thank you, @aswamy! This is going to bring so much power to LiquidDoc ✨

karreiro (approved):
Thank you, @aswamy!

PR #879, opened 2025-04-08
Do not report errors for non-global LiquidDoc @param variable names

charlespwd:


karreiro (approved):
Thank you, @charlespwd!

Issue #880, opened 2025-04-08
Major performance regression on 3.77.1 vs 3.76.2


charlespwd:
**Describe the bug**
There's a major performance regression introduced to theme check's performance between CLI 3.76.2 and CLI 3.77.1

**Source**

```liquid
# on 3.76.2
shopify theme check  7.48s user 0.45s system 161% cpu 4.901 total

# on 3.77.1
shopify theme check  22.47s user 0.87s system 150% cpu 15.512 total
```

**Expected behaviour**
3.77.1 shouldn't be this slow.

jamesmengo:
We can also disable / enable specific checks via the `theme-check.yml` to see if it's related to the LiquidDoc checks. I wonder if it's all the cross-file stuff that we added 🤔

aswamy:
My thoughts exactly @jamesmengo . I will start disabling all LiquidDoc checks first, then see if any specific ones are causing it. Im betting there is something we should be memoizing that we aren't.

aswamy:
Ok most LiquidDoc theme checks seem fine, but as expected, it only jumps up in completion time if i enable any of these three (all of which do cross-file checks):

```
MissingRenderSnippetParams
UnrecognizedRenderSnippetParams
ValidRenderSnippetParamTypes
```

EvilGenius13:
Would this potentially be because of the `readFile` we're doing on these checks?

PR #885, opened 2025-04-09
Add `getDocDefinition` to fetch memoized docs

aswamy:

Fixes https://github.com/Shopify/theme-tools/issues/880
- A few of our LiquidDoc related theme checks were slowing down theme checks because they directly access the file system rather than access it using our memoized system
- CLI theme check does not have access to the DocumentManager which provides proper caching
  - to follow existing patterns we use an injected method (that should also be flexible when we want to support fetching LiquidDoc from blocks)

image

### 📊 RESULT
| **Before**                                                                 | **After**                                                                 |
|----------------------------------------------------------------------------|---------------------------------------------------------------------------|
| 19.04s real             27.24s user             1.83s sys                  | 5.52s real              6.74s user              0.33s sys                 |

## Tophatting

- Checkout repo
- Run `yarn install && yarn build`
- Go to a theme repo
- Create a doc tag in one of its snippets
- Go to the file that renders that snippet, and ensure at least one warning/error exists on the render tag from liquiddoc
  - Ensure one of the errors that appears is `MissingRenderSnippetParams`, `UnrecognizedRenderSnippetParams`, or `ValidRenderSnippetParamTypes` (since that's what got changed in this PR)
- In the theme repo, run the the following

```
  node <path-to-theme-tools>/Shopify/theme-tools/packages/theme-check-node/dist/cli.js .
```

- three items should be consoled (offences, config, and theme files)
  - check to see if your render tag offence exists in the logs

#### BONUS TOPHATTING
- If you prepend the command above with `/usr/bin/time -l -h `, you will get stats on its performance

## What did you learn?

- Just new areas of the code i haven't explored before (specifically how we inject things into theme-checks)

## Before you deploy

### DIFF
--- a/packages/theme-check-node/src/index.ts
+++ b/packages/theme-check-node/src/index.ts
@@ -2,7 +2,6 @@ import {
   Config,
   JSONSourceCode,
   JSONValidator,
-  LiquidHtmlNode,
   LiquidSourceCode,
   Offense,
   SectionSchema,
@@ -124,7 +123,7 @@ export async function themeCheckRun(
     theme
       .filter((file) => isSnippet(file.uri))
       .map((file) => [
-        file.uri,
+        path.relative(URI.file(root).toString(), file.uri),
         memo(async (): Promise<SnippetDefinition | undefined> => {
           const ast = file.ast;
           if (!isLiquidHtmlNode(ast)) {
@@ -146,7 +145,7 @@ export async function themeCheckRun(
     getSectionSchema: async (name) => sectionSchemas.get(name)?.(),
     getBlockSchema: async (name) => blockSchemas.get(name)?.(),
     getAppBlockSchema: async (name) => blockSchemas.get(name)?.() as any, // cheating... but TODO
-    getDocDefinition: async (uri) => docDefinitions.get(uri)?.(),
+    getDocDefinition: async (relativePath) => docDefinitions.get(relativePath)?.(),
   });
 
   return {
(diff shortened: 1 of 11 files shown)


packages/theme-check-node/src/index.ts
```diff
@@ -115,6 +120,20 @@ export async function themeCheckRun(
       toSchema(config.context, source.uri, source, isValidSchema) as Promise<SectionSchema | undefined>
     )
   ]));
+  const docDefinitions = new Map(
+    theme
+      .filter((file) => isSnippet(file.uri))
```
EvilGenius13, line 125: Should we be grabbing blocks too?

aswamy: I have a separate PR for block support - for _now_ im only fixing performance issues

packages/theme-check-node/src/index.ts
```diff
@@ -127,6 +146,7 @@ export async function themeCheckRun(
     getSectionSchema: async (name) => sectionSchemas.get(name)?.(),
     getBlockSchema: async (name) => blockSchemas.get(name)?.(),
     getAppBlockSchema: async (name) => blockSchemas.get(name)?.() as any, // cheating... but TODO
+    getDocDefinition: async (uri) => docDefinitions.get(uri)?.(),
```
charlespwd, line 149: Last nit is probably that this should be either a relative path or a name instead?

EvilGenius13 (approved):
Huge performance improvement. 🎩  locally using `theme check` and running the server. Awesome @aswamy! 🚀

charlespwd (approved):
Bah I'm not sure I'm really that concerned about it. context.toUri is a thing so should be ok!

jamesmengo (approved):

PR #888, opened 2025-04-10
Amend UndefinedObject theme check to recognize @param variables

EvilGenius13:
Closes https://github.com/Shopify/developer-tools-team/issues/667

This change updates the `UndefinedObject` check to correctly identify variables declared using the `@param` tag within liquid doc blocks.

This is resolved by:

1. Adding a visitor for `LiquidDocParamNode` to the check
2. If we find a valid parameter name we add it to the `variableScopes` map with an empty array which treats it as a global variable within the file.

## Follow Up
I will open a separate issue/PR to allow snippets to use this check.

## Testing

- Pull down the branch
- Build the branch
- Hit F5 to start the development server
- Try adding a variable in a block file such as `{{ mage }}` and make sure it's caught as `UndefinedObject`
- Add a doc block (mini template here)
```javascript
{% doc %}
@param {string} mage
{% enddoc %}
```
- Watch the check disappear

I also added tests which you can run.

## Before you deploy

- [X] This PR includes a new checks or changes the configuration of a check
  - [X] I ran `yarn build` and committed the updated configuration files

### DIFF
--- a/.changeset/wild-seals-agree.md
+++ b/.changeset/wild-seals-agree.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-check-common': patch
+---
+
+Updated UndefinedObject check to recognize variables declared via `@param` within doc tags in blocks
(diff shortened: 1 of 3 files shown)


jamesmengo (commented):
This is pretty different from the snippets approach, is it time to update that one as well?

PR #889, opened 2025-04-11
Support `content_for` tag parameter completion

aswamy:

Closes https://github.com/Shopify/developer-tools-team/issues/664
- If a block contains `doc` tag, we read its params and offer it as completion when you try to do `content_for` tag param completion

image

15-33-4oi27-14o3i

## Tophat
- Checkout the code
- Create doc on a block
- Find a `content_for` tag that renders that block
- Trigger code completion within the tag and get offered param completion for params you don't already have

## Before you deploy

### DIFF
--- a/packages/theme-language-server-common/src/completions/providers/ContentForParameterCompletionProvider.ts
+++ b/packages/theme-language-server-common/src/completions/providers/ContentForParameterCompletionProvider.ts
@@ -9,6 +9,8 @@ import { CURSOR, LiquidCompletionParams } from '../params';
 import { Provider } from './common';
 import { AugmentedLiquidSourceCode } from '../../documents';
 import { DEFAULT_COMPLETION_OPTIONS } from './data/contentForParameterCompletionOptions';
+import { DocDefinition, GetDocDefinitionForURI } from '@shopify/theme-check-common';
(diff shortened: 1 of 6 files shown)


packages/theme-language-server-common/src/completions/providers/ContentForParameterCompletionProvider.ts
```diff
@@ -124,4 +130,62 @@ export class ContentForParameterCompletionProvider implements Provider {
...
+  ) {
+    let options = DEFAULT_COMPLETION_OPTIONS;
+
+    const partial = node.name!.replace(CURSOR, '');
+
+    if (isTypeBlocks) {
+      options = {
+        closest: DEFAULT_COMPLETION_OPTIONS.closest,
+      };
+    }
```
jamesmengo, line 147: Are there plans to use this at a later date?

jamesmengo (approved):
Left a couple of nits, but tophat went great! Thanks for the cleanup too :)

packages/theme-language-server-common/src/completions/providers/ContentForParameterCompletionProvider.ts
jamesmengo: Misread this: `parentNode.contentForType.value == 'blocks'`

PR #891, opened 2025-04-11
allow UndefinedObject check to work on snippets

EvilGenius13:

Closes: https://github.com/Shopify/developer-tools-team/issues/668

The `UndefinedObject` check would not look at snippet files. Now that we have liquiddoc support for them, we can amend the check to work on snippets.

## Testing
- Pull down the branch
- Build the branch
- Hit F5 to start the development server
- Try adding a variable in a snippet file such as `{{ mage }}` and make sure it's caught as `UndefinedObject`
- Add a doc block (mini template here)
```javascript
{% doc %}
@param {string} mage
{% enddoc %}
```
- Watch the check disappear

## Before you deploy

- [X] This PR includes a new checks or changes the configuration of a check
  - [X] I ran `yarn build` and committed the updated configuration files


packages/theme-check-common/src/liquid-doc/liquidDoc.ts
```diff
@@ -45,6 +45,16 @@ export interface LiquidDocDescription extends LiquidDocNode {
...
+export function hasLiquidDoc(snippet: LiquidHtmlNode): boolean {
+  let foundDocTag = false;
+  visit<SourceCodeType.LiquidHtml, void>(snippet, {
+    LiquidRawTag(node) {
+      if (node.name === 'doc') foundDocTag = true;
+    },
+  });
+  return foundDocTag;
+}
+
```
EvilGenius13, line 57: `hasLiquidDoc` is meant as a helper to detect the presence of doc tags but without building the arrays of what's inside the nodes. This should be a bit more performant if we're simply trying to check if `{% doc %} {% enddoc %} are in a file

EvilGenius13:
Thanks for everyone talking about the options we have! For now we have added an extra check in the `UndefinedObject` to look for `doc` tags if it's in a snippet. If there are none, we will just skip the check. I will open a separate issue about adding another check and laying out another area to discuss options.

packages/theme-check-common/src/liquid-doc/liquidDoc.ts
aswamy: Im okay with you using `extractDocDefinition` as well to check if liquidDoc is defined; but i can see the benefit of keeping this light as possible. 👍

aswamy (approved):

PR #894, opened 2025-04-15
Building blocks for `docs` on `blocks`

aswamy:

closes https://github.com/Shopify/developer-tools-team/issues/670

- Cleans up variables and methods so that we can support blocks in all of our LiquidDoc utils and memoized methods
- Most important pieces are in:
  - `packages/theme-check-common/src/liquid-doc/liquidDoc.ts`
  - `packages/theme-check-node/src/index.ts`
  - `packages/theme-language-server-common/src/diagnostics/runChecks.ts`
  - `packages/theme-language-server-common/src/documents/DocumentManager.ts`
- The rest are just refactors caused by the 4 changes above

- nothing specific

## Tophat

- Ensure the changed theme checks still work in VSCode AND via terminal

  - MissingRenderSnippetParams
  - UnrecognizedRenderSnippetParams
  - ValidRenderSnippetParamTypes

- VSCode tophatting
  - Open branch in VSCode
  - Press F5
  - Open Theme
  - Create doc in snippet
  - Go to render tag in a different file that renders the changed snippet
  - See theme check error

- terminal tophatting
  - Checkout this branch
  - build the branch `yarn build`
  - cd to a theme with errors (you can create some in the VScode tophatting step)
  - Run `node <path/to-theme-tools>/packages/theme-check-node/dist/cli.js .`

## Before you deploy

### DIFF
--- a/packages/theme-check-node/src/index.ts
+++ b/packages/theme-check-node/src/index.ts
@@ -1,20 +1,20 @@
 import {
   Config,
+  DocDefinition,
   JSONSourceCode,
   JSONValidator,
   LiquidSourceCode,
   Offense,
   SectionSchema,
-  SnippetDefinition,
   Theme,
   ThemeBlockSchema,
   toSourceCode as commonToSourceCode,
   check as coreCheck,
-  getSnippetDefinition,
+  extractDocDefinition,
+  filePathSupportsLiquidDoc,
   isBlock,
   isIgnored,
   isSection,
-  isSnippet,
   memo,
   path as pathUtils,
   toSchema,
@@ -120,18 +120,19 @@ export async function themeCheckRun(
     )
   ]));
   const docDefinitions = new Map(
-    theme
-      .filter((file) => isSnippet(file.uri))
-      .map((file) => [
-        path.relative(URI.file(root).toString(), file.uri),
-        memo(async (): Promise<SnippetDefinition | undefined> => {
-          const ast = file.ast;
-          if (!isLiquidHtmlNode(ast)) {
-            return undefined;
-          }
-          return getSnippetDefinition(ast, path.basename(file.uri, '.liquid'));
-        }),
-      ]),
+    theme.map((file) => [
+      path.relative(URI.file(root).toString(), file.uri),
+      memo(async (): Promise<DocDefinition | undefined> => {
+        const ast = file.ast;
+        if (!isLiquidHtmlNode(ast)) {
+          return undefined;
+        }
+        if (!filePathSupportsLiquidDoc(file.uri)) {
+          return undefined;
+        }
+        return extractDocDefinition(file.uri, ast);
+      }),
+    ]),
   );
 
   const offenses = await coreCheck(theme, config, {
--- a/packages/theme-language-server-common/src/completions/CompletionsProvider.ts
+++ b/packages/theme-language-server-common/src/completions/CompletionsProvider.ts
@@ -1,5 +1,5 @@
 import {
-  GetSnippetDefinitionForURI,
+  GetDocDefinitionForURI,
   MetafieldDefinitionMap,
   SourceCodeType,
   ThemeDocset,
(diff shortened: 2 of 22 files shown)


aswamy:
> I know the logic should be the same in `GetDocDefinitionForURI` regardless of kind of `DocSupportedFile` we have (block vs snippet) but would there be any benefit to having one of the many tests using `DocSupportedFile.BLOCK` and switch it to be a `SNIPPET` type instead?

Good catch! Actually, i think I _SHOULD_ be using `DocSupportedFile.SNIPPET` in most of the existing test cases instead of `BLOCK` since existing theme checks are against snippets. It doesn't really make functional difference in the test cases since they are stubbed, but for logical consistency, ill fix that. I don't think there is any benefit from switching between both though 🤷.

packages/theme-language-server-common/src/documents/DocumentManager.ts
```diff
@@ -195,10 +196,10 @@ export class DocumentManager {
             return toSchema(mode, uri, sourceCode, this.isValidSchema, false);
           }),
           /** Lazy and only computed once per file version */
-          getLiquidDoc: memo(async (snippetName: string) => {
+          getLiquidDoc: memo(async (type: DocSupportedFile, name: string) => {
             if (isError(sourceCode.ast)) return undefined;
 
-            return getSnippetDefinition(sourceCode.ast, snippetName);
+            return extractDocDefinition(sourceCode.ast, type, name);
```
charlespwd, line 202: Isn't it weird for a document to not know its type? Wouldn't a snippet always be a snippet and a block always a block, why memo with arguments? Shouldn't this take no args at all?

packages/theme-check-node/src/index.ts
```diff
@@ -145,7 +157,8 @@ export async function themeCheckRun(
     getSectionSchema: async (name) => sectionSchemas.get(name)?.(),
     getBlockSchema: async (name) => blockSchemas.get(name)?.(),
     getAppBlockSchema: async (name) => blockSchemas.get(name)?.() as any, // cheating... but TODO
-    getDocDefinition: async (relativePath) => docDefinitions.get(relativePath)?.(),
+    getDocDefinition: async (type: DocSupportedFile, name: string) =>
+      docDefinitions.get(path.join(type, `${name}.liquid`))?.(),
```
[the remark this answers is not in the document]
aswamy: yeah... i feel like i had a reason to use a more structured approach (having explicit types being passed in). Let me see what happens if i keep relativePath.

packages/theme-language-server-common/src/documents/DocumentManager.ts
aswamy: This isn't fetching the CURRENT file's LiquidDoc. It is fetching the LiquidDoc for _another_ block or snippet's LiquidDoc

packages/theme-language-server-common/src/completions/CompletionsProvider.ts
```diff
@@ -13,23 +13,23 @@ import { createLiquidCompletionParams } from './params';
...
   LiquidTagsCompletionProvider,
   ObjectAttributeCompletionProvider,
   ObjectCompletionProvider,
   Provider,
   RenderSnippetCompletionProvider,
+  RenderSnippetParameterCompletionProvider,
   TranslationCompletionProvider,
-  FilterNamedParameterCompletionProvider,
-  ContentForParameterCompletionProvider,
 } from './providers';
```
jamesmengo, line 32: ❤️

packages/theme-language-server-common/src/hover/providers/RenderSnippetHoverProvider.ts
```diff
@@ -42,12 +43,12 @@ export class RenderSnippetHoverProvider implements BaseHoverProvider {
       return {
         contents: {
           kind: 'markdown',
-          value: `### ${snippetDefinition.name}`,
+          value: `### ${path.basename(snippetName)}`,
```
jamesmengo, line 46: When would path.basename be diff than `snippetName`?

aswamy: hmm no clue why i did it this way - i think it should _just_ be `snippetName` not `path.basename(snippetName)`. Good catch. Ill push a change.

PR #895, opened 2025-04-15
Fix ValidBlockTarget behaviour when it comes to static blocks

charlespwd:
- Store an array of StaticBlockDefinitions to the ThemeBlockSchema & SectionSchema objects that we cache
  - `schema.staticBlockDefs: { type, id }[]`;
- Validate `Preset.Block` nodes that are `static` against the `StaticBlockDef[]` of the parent's,
  - Report a warning unless the parent has a StaticBlockDef entry with the same id & type
- Two new error messages:
  - `Could not find a static block of type "static" with id "mismatching_id" in this file.` (top level preset)
  - `Could not find a static block of type "static" with id "mismatching_id" in "blocks/slideshow.liquid".` (nested presets)

No

## What did you learn?

This check did not consider static blocks and was throwing false warnings

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/utils/block.ts
+++ b/packages/theme-check-common/src/utils/block.ts
@@ -6,23 +6,29 @@ import {
   SourceCodeType,
   ThemeBlock,
   Context,
+  StaticBlockDef,
 } from '../types';
 import { getLocEnd, getLocStart, nodeAtPath } from '../json';
 import { doesFileExist } from './file-utils';
 
-export type BlockNodeWithPath = {
-  node: Section.Block | ThemeBlock.Block | Preset.Block;
+export type BlockDefNodeWithPath = {
+  node: Section.Block | ThemeBlock.Block;
+  path: string[];
+};
+
+export type PresetBlockNodeWithPath = {
+  node: Preset.Block;
   path: string[];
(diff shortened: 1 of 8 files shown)


packages/theme-check-common/src/utils/block.ts
```diff
@@ -112,9 +118,19 @@ export function getBlocks(validSchema: ThemeBlock.Schema | Section.Schema): {
...
 
 export function isInvalidPresetBlock(
-  blockNode: Section.Block | ThemeBlock.Block | Preset.Block,
-  rootLevelThemeBlocks: BlockNodeWithPath[],
+  blockId: string,
+  blockNode: Preset.Block,
+  rootLevelThemeBlocks: BlockDefNodeWithPath[],
+  staticBlockDefs: StaticBlockDef[],
 ): boolean {
+  const isStaticBlock = 'static' in blockNode && blockNode.static;
```
clauderic, line 126: Should we add `static?: boolean` to the `Preset.Block` type?

clauderic (approved):
Looks great 💯

karreiro (approved):
Thank you for this PR, @charlespwd!

PR #900, opened 2025-04-17
Restrict block `doc` param names

aswamy:

Closes https://github.com/Shopify/developer-tools-team/issues/671
- Restrict `content_for` param names inside of `doc` tag to avoid collision
- Unlike our `ValidDocParamNames`, this one is an error because the collision could be fatal rather than just ambiguous

## Tophat

- Checkout this branch
- `yarn build`
- F5
- Open a theme
- Create a block file
- Add `doc` tag to it and create a `param` with the name `id`, `type`, or `closest` - you should see an error

17-32-jq5wk-fg2kd

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/checks/valid-doc-param-names/index.ts
+++ b/packages/theme-check-common/src/checks/valid-doc-param-names/index.ts
@@ -7,7 +7,7 @@ export const ValidDocParamNames: LiquidCheckDefinition = {
     name: 'Valid doc parameter names',
     docs: {
       description:
-        'This check exists to ensure any parameter names defined in LiquidDoc do not collide with liquid tags and global variables.',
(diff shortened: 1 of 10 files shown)


packages/theme-check-common/src/checks/valid-doc-param-names/index.ts
```diff
@@ -7,7 +7,7 @@ export const ValidDocParamNames: LiquidCheckDefinition = {
     name: 'Valid doc parameter names',
     docs: {
       description:
-        'This check exists to ensure any parameter names defined in LiquidDoc do not collide with reserved words.',
+        'This check exists to ensure any parameter names defined in LiquidDoc do not collide with liquid tags and global variables.',
```
jamesmengo, line 10: I thiiink this reads better as `OR global variables`

PR #902, opened 2025-04-22
Allow use of 'section.' in block visible_if

miazbikowski:
Addresses https://github.com/Shopify/online-store-web/issues/22606

This fix removes the requirement to be in a section file in order to use the `{{ section }}` when using `invisible_if`, since it's valid that we would do this.

There's one particularity about this case, though. Previously, if a section was in use in a section file, or a block was in use in a block file, or the other liquid variables, we'd hit this line and [validate that the setting we're using actually exists](https://github.com/Shopify/theme-tools/blob/main/packages/theme-check-common/src/checks/valid-visible-if/index.ts#L115). But in this new case, we don't know which section the block might be used in, so we can't add that additional validation. Hence, I added a case where if we're using the `section` and we're in a block, we don't do the lookup check. 

Nothing, but hope to make it for release scheduled in a few days.

## What did you learn?
My brain wrinkled extra hard for a minute, readjusting to Typescript and how Theme Check works. But it was nice to flex that disused muscle.

## Before you deploy


packages/vscode-extension/syntaxes
graygilmore, line 1: I have a feeling this change shouldn't have come along for the ride?

graygilmore (commented):
Makes sense! Just a note about the `syntaxes` submodule changes.

.changeset/smart-panthers-collect.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/theme-check-common': patch
+---
+
+invisible_if should not warn on use of 'section' in a block file
```
mmorissette, line 5: Not sure what this file is but this should be `visible_if` as opposed to `invisible_if`

```suggestion
visible_if should not warn on use of 'section' in a block file
```

mmorissette (approved):
changes look good

PR #915, opened 2025-04-28
Show a hover window for things that are knowingly untyped

### LINKED ISSUE #913: Hover info missing when type unknown
**Describe the bug**
When I hover over a variable that is untyped, there's no hover info shown. 

https://github.com/user-attachments/assets/ad921395-0bf1-45d9-bbee-a8b901f2f313

**Source**

```liquid
{% # this shows hover info %}
{% assign variants = product.variants %}

{{ variants }}

{% # this is does not? %}
{% assign results = search.results %}

{{ results }}
```

**Expected behaviour**
- both variants and results have a hover info window

**Actual behaviour**
- when we don't know the type of something, we don't show the hover info window

**Debugging information**
 - OS [e.g. Windows, Mac, Linux] - Mac
 - Theme Check Version : `main` branch

charlespwd:
Fixes #913

Making it so that when you hover over variables that have a unknown type, you still show a window with just its name. 

Reasoning: _not_ having anything made me feel like hover was bugged. No information = information. 

But we should only do that when we know that the variable is defined (assigned). Variable that were never assigned should still show nothing. We'll do that by differentiating between "untyped" and "unknown". Unknown being possible only for identifiers that don't have a TypeRange associated with them. 

Instead, we'll show the window with just the variable name. Seems silly, but it's valuable. 

### DIFF
--- a/packages/theme-language-server-common/src/TypeSystem.ts
+++ b/packages/theme-language-server-common/src/TypeSystem.ts
@@ -374,12 +374,12 @@ type ObjectEntryName = ObjectEntry['name'];
 type FilterEntryName = FilterEntry['name'];
 
 /** Untyped is for declared variables with unknown type */
-const Untyped = 'untyped' as const;
-type Untyped = typeof Untyped;
+export const Untyped = 'untyped' as const;
+export type Untyped = typeof Untyped;
 
 /** Unknown is for variables that don't exist, type would come from context (e.g. snippet var without LiquidDoc) */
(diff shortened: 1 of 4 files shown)


EvilGenius13:
I'm probably being picky so feel free to ignore. Should we add anything more to the hover information when it's untyped or unknown. 
This is what we see now:
image
Could we instead get something like below where we add a small blurb like "no information available" or "unknown variable" or something? I'm just thinking when we only return the name it appears like it didn't load properly.
image

PR #916, opened 2025-04-29
For aliased filters, show the doc of the base filter

charlespwd:
Noticed that the `t` filter had a silly hover window.

Now the `t` filter will show the docs of the `translate` filter.

image

- [x] I have a patch changeset


karreiro (approved):
Thank you, @charlespwd!

PR #919, opened 2025-04-29
Support LiquidDoc param array types

aswamy:

Closes https://github.com/Shopify/developer-tools-team/issues/682

- Add support for array types in LiquidDoc

## Tophat

- Open the branch in VSCode
- `yarn install && yarn build`
- Press F5
- Open a theme and create a snippet/block with LiquidDoc
- Try out different types (`string`, `string[]`, `product`, `product[]`, `object`, `object[]`, etc.)
- NOTE: `object` is not a real type. It is treated as "untyped", so `object[]` is treated an array of untyped objects

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/checks/valid-doc-param-types/index.ts
+++ b/packages/theme-check-common/src/checks/valid-doc-param-types/index.ts
@@ -1,5 +1,5 @@
 import { LiquidCheckDefinition, Severity, SourceCodeType } from '../../types';
-import { getValidParamTypes, BasicParamTypes } from '../../liquid-doc/utils';
+import { getValidParamTypes, parseParamType } from '../../liquid-doc/utils';
 
 export const ValidDocParamTypes: LiquidCheckDefinition = {
   meta: {
@@ -22,22 +22,29 @@ export const ValidDocParamTypes: LiquidCheckDefinition = {
       return {};
     }
 
+    // To avoid recalculating valid param types during theme-check, constructing
+    // the promise beforehand.
+    const validParamTypesPromise = context
+      .themeDocset!.liquidDrops()
+      .then((entries) => new Set(getValidParamTypes(entries).keys()));
+
     return {
       async LiquidDocParamNode(node) {
         if (!node.paramType) {
           return;
         }
 
-        const objectEntries = await context.themeDocset!.objects();
+        const parsedParamType = parseParamType(await validParamTypesPromise, node.paramType.value);
 
-        if (getValidParamTypes(objectEntries).has(node.paramType.value)) {
+        if (parsedParamType) {
           return;
         }
 
         context.report({
           message: `The parameter type '${node.paramType.value}' is not supported.`,
-          startIndex: node.paramType.position.start,
-          endIndex: node.paramType.position.end,
+          // Index is offset to include the curly brackets around the param type
+          startIndex: node.paramType.position.start - 1,
+          endIndex: node.paramType.position.end + 1,
           suggest: [
             {
               message: 'Remove invalid parameter type',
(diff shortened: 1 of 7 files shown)


packages/theme-check-common/src/checks/valid-doc-param-types/index.ts
```diff
@@ -28,16 +28,23 @@ export const ValidDocParamTypes: LiquidCheckDefinition = {
           return;
         }
 
-        const objectEntries = await context.themeDocset!.objects();
+        const objectEntries = await context.themeDocset!.liquidDrops();
 
-        if (getValidParamTypes(objectEntries).has(node.paramType.value)) {
+        const validParamTypes = getValidParamTypes(objectEntries);
```
jamesmengo, line 33: Outside the scope of this PR:
Does `objectEntries` change in different areas of the system? If not, I wonder if we can avoid computing this multiple times, wdyt?

packages/theme-check-common/src/liquid-doc/utils.ts
```diff
@@ -107,3 +107,19 @@ export function getValidParamTypes(objectEntries: ObjectEntry[]): Map<string, st
 
   return paramTypes;
 }
+
+export function parseParamType(
+  validParamTypes: Set<string>,
+  value: string,
+): [string, boolean] | undefined {
```
jamesmengo, line 114: Are you open to returning a structured type here?

packages/theme-language-server-common/src/TypeSystem.ts
```diff
@@ -659,20 +660,26 @@ function inferType(
...
+  const parsedParamType = parseParamType(new Set(validParamTypes.keys()), paramTypeValue);
 
-  if (getValidParamTypes(liquidDrops).has(node.paramType.value)) {
-    return node.paramType.value;
+  if (!parsedParamType) return Untyped;
+
+  const [type, isArray] = parsedParamType;
+
+  // BasicParamTypes.Object does not map to any specific type in the type system.
+  const transformedParamType = type === BasicParamTypes.Object ? Untyped : type;
```
jamesmengo, line 676: mega-nit: slightly confusing to read but I don't feel strongly about this

jamesmengo (approved):
Excited about this change! Just 1 comment re: return types that I'd like to align on.

Side note: how much effort would it take to implement a hover on the array accessors so that `asdf.first` in the screenshot below would display a `product`?

image

packages/theme-language-server-common/src/TypeSystem.spec.ts
```diff
@@ -493,6 +493,27 @@ describe('Module: TypeSystem', () => {
...
+      assert(isLiquidVariableOutput(variableOutput));
+      const inferredType = await typeSystem.inferType(
+        variableOutput.markup,
+        ast,
+        'file:///snippets/example.liquid',
+      );
+      expect(inferredType).to.eql({
+        kind: 'array',
+        valueType: 'product',
+      });
```
charlespwd, line 515: :chef-kiss:

packages/theme-check-common/src/liquid-doc/utils.ts
charlespwd: We could also name the tuple to make it more obvious that the second boolean means `isArray`

```
): [pseudoType: PseudoType, isArray: boolean] | undefined
```

charlespwd:
> Side note: how much effort would it take to implement a hover on the array accessors so that asdf.first in the screenshot below would display a product?

I thought we did do that? 🤔 Is that bugged?

https://github.com/Shopify/theme-tools/blob/971acf27d8e58839aa28cfd49eb829c41a0ad458/packages/theme-language-server-common/src/TypeSystem.ts#L791-L792

EvilGenius13 (approved):

packages/theme-check-common/src/checks/valid-doc-param-types/index.ts
aswamy: Hmm yes - i can't do this memoization across code completion AND theme checks because they both have different `themeDocset` injected into it, but maybe i can memoize this within the theme check. Good catch!

aswamy:
>> Side note: how much effort would it take to implement a hover on the array accessors so that asdf.first in the screenshot below would display a product?

> I thought we did do that? 🤔 Is that bugged?

@charlespwd @jamesmengo is right! Something is bugged out about this. Ill do a quick follow-up after i do the PR comments + merge this PR.

charlespwd:
image

lol. hi! that's me again! 🙈

PR #923, opened 2025-04-30
Fixup hover of builtin properties of arrays and literals

### LINKED ISSUE #921: LiquidObjectAttributeHoverProvider does not support array attributes
Did a bit of investigation, and see like we do support **some** attributes, but not all. Hover works for attributes that exist IN the docset's object map.

E.g. Imagine you have an array of colors

```
{{ color_array.first.rgba.size }}
```

Hover would work on `color_array` (it would recognize correctly as `color[]`), and `rgba` (it would recognize correctly as `string`).

Hover on `first` and `size` does not work for two different reasons:
- `size` does not work because `string` type does not exist in objectMap ([here](https://github.com/Shopify/theme-tools/blob/main/packages/theme-language-server-common/src/hover/providers/LiquidObjectAttributeHoverProvider.ts#L37))
- `first` does not work because we explicitly don't support `array` support ([here](https://github.com/Shopify/theme-tools/blob/main/packages/theme-language-server-common/src/hover/providers/LiquidObjectAttributeHoverProvider.ts#L33))

_Originally posted by @aswamy in https://github.com/Shopify/theme-tools/issues/919#issuecomment-2842208885_

What to fix:
- Make it work for arrays
- Make it work for basic types that aren't in `objectMap` - i.e. `string`, `number`, boolean`, etc.

charlespwd:
Add hover support of builtin properties

```liquid
{% # string builtins %}
{{ product.title.size }} -> number
{{ product.title.first }} -> string
{{ product.title.last }} -> string
{{ product.title.last.last }} -> string
{{ product.title.foo }} -> null

{% # array lookups %}
{{ product.images.first }} -> image
{{ product.images.last }} -> image
{{ product.images.size }} -> number
{{ product.images.foo }} -> null
```

Fixes #921

## Before you deploy

### DIFF
--- a/packages/theme-language-server-common/src/TypeSystem.ts
+++ b/packages/theme-language-server-common/src/TypeSystem.ts
@@ -827,6 +827,7 @@ function inferPseudoTypePropertyType(
 ) {
   const parentEntry: ObjectEntry | undefined = objectMap[curr];
 
+  // When doing a non string lookup, we don't really know the type. e.g.
   // products[0]
   // products[true]
   // products[(0..10)]
@@ -834,7 +835,10 @@ function inferPseudoTypePropertyType(
     return Untyped;
   }
 
+  // When we don't have docs for the parent entry
   if (!parentEntry) {
+    // It might be that the parent entry is a string.
+    // We do support a couple of properties for those
     if (curr === 'string') {
       switch (lookup.value) {
         // some_string.first
@@ -858,23 +862,26 @@ function inferPseudoTypePropertyType(
       }
     }
 
-    // something.foo could be anything
+    // Or it might be that the parent entry is untyped, so its subproperty
+    // could also be untyped (kind of like if `foo` is `any`, then `foo.bar` is `any`)
     return Untyped;
   }
 
   const propertyName = lookup.value;
   const property = parentEntry.properties?.find((property) => property.name === propertyName);
 
-  // product.unknown
+  // When the propety is not known, return Untyped. e.g.
+  // product.foo
+  // product.bar
   if (!property) {
     // Debating between returning Untyped or Unknown here
-    // Might be that we have outdated docs. Prob better to return untyped here.
+    // Might be that we have outdated docs. Prob better to return Untyped.
     return Untyped;
   }
 
+  // When the property is known & we have docs for it, return its type. e.g.
   // product.image
   // product.images
-  // return the type of the property
   return objectEntryType(property);
 }
 
--- a/packages/theme-language-server-common/src/TypeSystem.ts
+++ b/packages/theme-language-server-common/src/TypeSystem.ts
@@ -374,7 +374,7 @@ type Identifier = string;
 type ObjectEntryName = ObjectEntry['name'];
 type FilterEntryName = FilterEntry['name'];
 
-/** Untyped is for declared variables with unknown type */
+/** Untyped is for declared variables without a type (like `any`) */
 export const Untyped = 'untyped' as const;
 export type Untyped = typeof Untyped;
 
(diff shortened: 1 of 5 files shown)


packages/theme-language-server-common/src/TypeSystem.ts
```diff
@@ -809,7 +809,7 @@ function inferArrayTypeLookupType(curr: ArrayType, lookup: LiquidExpression) {
       }
 
       default: {
-        return Untyped;
+        return Unknown;
```
charlespwd, line 812: `{{ product.images.foo }}` should not have a hover window since we know that `product.images` is an `image[]` and `foo` is not a platform supported feature.

packages/theme-language-server-common/src/hover/providers/LiquidObjectAttributeHoverProvider.ts
```diff
@@ -29,14 +30,28 @@ export class LiquidObjectAttributeHoverProvider implements BaseHoverProvider {
...
+          kind: 'markdown',
+          value: render(entry, nodeType),
+        },
+      };
     }
 
-    const objectMap = await this.typeSystem.objectMap(params.textDocument.uri, ancestors[0]);
     const parentEntry = objectMap[parentType];
-    if (!parentEntry) {
+    if (!parentEntry && parentType !== Unknown && parentType !== 'untyped') {
```
charlespwd, line 54: am not sure about this condition anymore, should revisit. 

If I have something that is `any` and I hover over one of its properties, then its property should also be `any` and I should show a hover window.

packages/theme-language-server-common/src/TypeSystem.ts
graygilmore, line 1: Maybe this file should have some unit tests?

graygilmore (approved):
Some of the comments aren't completely clear (e.g. L868 `// product.unknown`) and could use a little bit of context but otherwise I think I follow what's happening here.

packages/theme-language-server-common/src/TypeSystem.ts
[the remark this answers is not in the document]
charlespwd: No?!
image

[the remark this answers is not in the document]
charlespwd: Good catch! Forgot that the fixer included the markup of the current node. Fixed!

packages/theme-language-server-common/src/TypeSystem.ts
```diff
@@ -830,18 +830,51 @@ function inferPseudoTypePropertyType(
...
+        default: {
+          // For the string type, any property access other than first/last/size
+          // is unknown. This is different from an untyped/any object where any
+          // property access would return untyped.
+          // String is a known type with specific properties, so accessing
+          // undefined properties returns an unknown.
+          return Unknown;
+        }
+      }
+    }
```
aswamy, line 859: Im surprised this doesn't already exist in the code

aswamy (approved):
Tophat went great ✅ One minor issue, but i think it existed way before this PR:

E.g. `{{ product.images.first.height }}`

If your cursor is right after any `.`, you'll get the hover text for `product`, not the attribute you're actually hovering on.

image

packages/theme-language-server-common/src/TypeSystem.ts
charlespwd: 3b634d34 !

packages/theme-language-server-common/src/TypeSystem.ts
charlespwd: I think it does for array but not for strings? 🤔

PR #926, opened 2025-05-01
Fix attribute hover when directly after the dot

### LINKED ISSUE #925: Hover bug if your cursor is right after any `.`
**Describe the bug**
If your cursor is right after any ., you'll get the hover text for product, not the attribute you're actually hovering on.

**Source**

Image

```liquid
{{ product.title.█size }}
```

Returns hover of the wrong thing. 

**Expected behaviour**
Should show docs for size

**Actual behaviour**
Not that

charlespwd:
and in other places... by adding a bunch of tests to findCurrentNode
and fixing a couple of edge cases

Fixes #925

## Before you deploy

### DIFF
--- a/.changeset/smooth-carrots-prove.md
+++ b/.changeset/smooth-carrots-prove.md
@@ -1,0 +1,6 @@
+---
+'@shopify/theme-language-server-common': patch
+'@shopify/theme-check-common': patch
+---
+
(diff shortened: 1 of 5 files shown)


aswamy (approved):
Thank you for the fix ❤️ 

Definitely annoyed me. Didn't release it was this complicated.

PR #929, opened 2025-05-05
[BugFix] Update `doc` tag code completion template

aswamy:

Closes https://github.com/Shopify/developer-tools-team/issues/691

- Bug Fix: When you code complete `doc` tag, it does not place all of the example code in the docs as content

NOTE: There is a bug where `enddoc` does not code complete and this is known. This is a limitation on how our ohm rules are written.

## Tophat

- checkout this code
- `yarn install & yarn build`
- F5
- Create a block/snippet and attempt to add `doc` tag
  - If you select it from the dropdown, it should code complete the tag + it's associated end tag. The cursor should be in the middle of the `doc` tag
  - If you type `@`, it should code complete with `@param`, `@description`, and `@example` without forcing code completion using `ctrl + space`

## Release

<video src="https://github.com/user-attachments/assets/8a87a27e-4014-4b35-b072-7e745eaa77eb" />

### DIFF
--- a/.changeset/honest-queens-lie.md
+++ b/.changeset/honest-queens-lie.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-language-server-common': patch
+---
+
+[BugFix] Update `doc` tag code completion template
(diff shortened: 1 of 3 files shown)


jamesmengo:
sweet! Thanks for the quick fix 🔥

Issue #930, opened 2025-05-06
[Theme Check] MatchingTranslations rule handling Customer Accounts translations incorrectly


jamespeilow:
**Describe the bug**

When running `shopify theme check`, the `MatchingTranslations` reports errors for the `customer_accounts.` namespaced translations when a merchant has changed one or more of the Customer Accounts translations.

When the merchant changes one of the Customer Accounts translations, the changed string gets added to the theme locale JSON file under the `customer_accounts` namespace. However only the changed line gets added and only to the changed language's locale JSON file, causing Theme Check to then report the `MatchingTranslations` error as the translation is missing from the other locales.

The `customer_accounts.` translations all exist, and are visible in the Admin under in Languages and in Translate & Adapt, but they don't exist in a themes locales JSON files. This is the same behaviour as how  Checkout translations are handled.

**Expected behaviour**

The `MatchingTranslations` rule should ignore translations that start with `customer_accounts.`, similarly to how it ignores translations starting `shopify.`

**Actual behaviour**

The `MatchingTranslations` rule reports missing translations when a `customer_accounts.` translation does not appear in every locale file.

![Image](https://github.com/user-attachments/assets/fd5c18fe-7acc-4916-afea-5b0128c3fdb8)

**Additional context**

This issue isn't an problem with Checkout translations, as these get added under the `shopify` namespace, which is excluded by the `isShopifyPath` check in the `MatchingTranslations` rule - https://github.com/Shopify/theme-tools/blob/90577cd2b5381eb6b8c50cde00186dccf0088577/packages/theme-check-common/src/checks/matching-translations/index.ts#L51

Updating that function to also return true for paths starting `customer_accounts.` would fix the issue:

```js
    const isShopifyPath = (path: string) => path.startsWith('shopify.') || path.startsWith('customer_accounts.');
```

karreiro:
👋 Hey @jamespeilow,

Thanks for reporting this!

The `shopify.` is indeed a special case. However, it looks like in the `customer_accounts` scenario, the matching translations feature is correctly identifying and reporting missing translations, as we can see in the example below:

https://github.com/user-attachments/assets/832f66a9-0339-47a5-8804-dc75c7537b22

Thanks again for highlighting this! Please let me know if I've misunderstood the issue you're experiencing.

jamespeilow:
Hi @karreiro 

I believe that `customer_accounts` should also be handled as a special case, the same way that `shopify.` translations are. 

[...]

PR #931, opened 2025-05-06
Fix MatchingTranslations check for Accounts translations

jamespeilow:

Fixes: https://github.com/Shopify/theme-tools/issues/930

Update `MatchingTranslations` theme check to prevent reporting issues with Customer Accounts translations.
- Extends the `isShopifyPath` check in the matching translations check so that issues are not reported if the locale path starts with `shopify.` _or_ `.customer_accounts.`
- Adds tests

## Before you deploy


jamespeilow:
I have signed the CLA!

Issue #933, opened 2025-05-06
Bug: Block Allowances check should skip static blocks


miazbikowski:
**Describe the bug**
https://share.descript.com/view/65pNWIzRNpk

When checking for allowed blocks, we should skip blocks that are static, since they are not required to be in the schema's `"blocks"` list.

**Source**
http://github.com/Shopify/horizon/pull/4076/files
(see the Theme Check warning)

**Expected behaviour**
We can have a **static** block in the json template that is not in the allowances of its parent and not get a warning.

**Actual behaviour**
We are getting a Theme Check warning that the block is not allowed in the parent block.

PR #940, opened 2025-05-07
handle static blocks in template validation

NathanPJF:

Fixes: https://github.com/Shopify/theme-tools/issues/933

As the issue says, static blocks do not need to be in the `blocks` array. In fact, they often aren't to prevent adding multiple blocks intended only to be included once.

PR removes erroneous theme-check error in template files. However, it does not check if a static block referenced in the template is actually used by the Liquid section/block file.  New issue made: https://github.com/Shopify/theme-tools/issues/941

## Before you deploy

### DIFF
--- a/.changeset/rotten-dots-breathe.md
+++ b/.changeset/rotten-dots-breathe.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-check-common': patch
+---
+
+Fix validation for static blocks in JSON templates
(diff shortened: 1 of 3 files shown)


jamesmengo (approved):
Thanks for cleaning up the code as well!

🎩 went well and the code LGTM

NathanPJF:
@jamesmengo I just made an issue now :) https://github.com/Shopify/theme-tools/issues/941

PR #950, opened 2025-05-17
Add theme check to encourage routes object use

### LINKED ISSUE #442: Feature: Recommend usage of new routes object
A lot of developers still use the old method of hardcoding various Shopify routes.

Given there's a fairly small amount of routes, could a check be included to ensure liquid routes are being used?
image

```liquid
<form action="{{ routes.search_url }}">...</form>

Vs.
<form action="/search">...</form>
```

dejmedus:
- closes #442
- PR for associated [shopify-dev docs page](https://github.com/Shopify/shopify-dev/pull/57937)

[Liquid docs](https://shopify.dev/docs/api/liquid/objects/routes) recommend using the routes object (`{{ routes.search_url }}`) instead of hardcoding URLs (`/search`). This PR adds a theme check warning to encourage routes object use

Checks `link_to` filters and `href`/`action` attribute values for hardcoded routes that could be a routes object

harded-coded-routes-updated

hardcoded-routes-in-link-to

## How to test
- Add a hardcoded route to a file (snippet, section, etc)
``` ts
<a href="/cart/change"></a>
```
- Should display a warning until it is changed to a route obj url
``` ts
<a href="{{ routes.cart_change_url }}"></a>
```

- PR for theme check [documentation page in shopify-dev](https://github.com/Shopify/shopify-dev/pull/57937)

## Before you deploy

- [x] This PR includes a new check
    
    
    
    
  - [x] I've made a PR to update the [shopify.dev theme check docs](https://github.com/Shopify/shopify-dev/tree/main/content/storefronts/themes/tools/theme-check/checks) if applicable (https://github.com/Shopify/shopify-dev/pull/57937).


packages/theme-check-common/src/checks/hardcoded-routes/index.ts
```diff
@@ -0,0 +1,107 @@
+import { NodeTypes, TextNode } from '@shopify/liquid-html-parser';
+import { Severity, SourceCodeType, LiquidCheckDefinition } from '../../types';
+import { isAttr, isNodeOfType, isValuedHtmlAttribute, ValuedHtmlAttribute } from '../utils';
+
+const ROUTES: Record<string, string> = {
```
dejmedus, line 5: These are all the routes from the [routes obj docs](https://shopify.dev/docs/api/liquid/objects/routes) except `/`. 

I chose not to warn on `/` to align with how I saw route urls used in a best practices Shopify theme repo and because I wasn’t sure if it would have the same benefits (translations etc.)

EvilGenius13:
Small thing I noticed but if we try a script like this:
```html
<script>
  // Set up the /search route:
  let config = {
    baseUrl: '/search',
    otherParam: 'value',
  };
  console.log('Using base URL:', config.baseUrl);
</script>
```

It will highlight the commented version of `/search` and ignore the one in `baseUrl: '/search`. Not really sure anyone would comment like that but thought I'd point it out.

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
jamesmengo: I believe that it may still be worth it based off what I'm reading here - https://shopify.dev/docs/api/liquid/objects/routes#routes

You can see this on my store where the default language is set to french
`https://mammoth-matcha-redirect.shopitest.com` redirects me to `https://mammoth-matcha-redirect.shopitest.com/en` since my browser is set to English

dejmedus: That makes sense, thanks! Fixed

packages/theme-check-common/src/checks/hardcoded-routes/index.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+import { runLiquidCheck, highlightedOffenses } from '../../test';
+
+describe('Module: HardcodedRoutes', () => {
+  it('should report offenses for hardcoded routes', async () => {
+    const sourceCode = `
+      <a href="/"></a>
+      <form action="/search"></form>
+      <a href="/account/login"></a>
+      <a href="/collections/{{ collection.handle }}"></a>
+      <a href="/cart/change?line={{ forloop.index }}&amp;quantity=0"></a>
```
graygilmore, line 12: I found a use case that we should probably cover:

```liquid
{{ 'Cart' | link_to: '/cart' }}
```

I'm not sure how much the `link_to` filter is used but worth covering! Feel free to address this in a follow up PR.

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
```diff
@@ -0,0 +1,131 @@
...
+  '/cart/change': 'cart_change_url',
+  '/cart/clear': 'cart_clear_url',
+  '/cart/update': 'cart_update_url',
+  '/cart': 'cart_url',
+  '/collections': 'collections_url',
+  '/search/suggest': 'predictive_search_url',
+  '/recommendations/products': 'product_recommendations_url',
+  '/': 'root_url',
+  '/search': 'search_url',
+  '/customer_authentication/login': 'storefront_login_url',
```
graygilmore: Nit: prefer to alphabetize the list based on the key instead of the value.

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
```diff
@@ -0,0 +1,131 @@
...
+
+const HARDCODED_ROUTES = Object.keys(ROUTES);
+
+export const HardcodedRoutes: LiquidCheckDefinition = {
+  meta: {
+    code: 'HardcodedRoutes',
+    name: 'Hardcoded Routes',
+    docs: {
+      description:
+        'This check is aimed at encouraging use of the routes object instead of hardcoding URLs.',
```
graygilmore: Nit:

```suggestion
        'This check encourages using the routes object instead of hardcoding URLs.',
```

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
```diff
@@ -0,0 +1,131 @@
...
+
+          context.report({
+            message: `Use routes object {{ routes.${routeURL} }} instead of hardcoding ${route}`,
+            startIndex,
+            endIndex,
+          });
+        }
+      },
+
+      async HtmlRawNode(node) {
```
graygilmore: This check feels like the most fragile. Is this just for handling script tags? Might be worth adding a comments above here explaining what it's explicitly checking for?

graygilmore (commented):
Tophat went well! Some small comments. 

I wish I had a suggestion for cleaning up `HtmlRawNode` but nothing immediately springs to mind here.

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
dejmedus: Yes, its checking for hardcoded routes that are not commented out inside script tags

packages/theme-check-common/src/checks/hardcoded-routes/index.spec.ts
dejmedus: Thank you! I added in [a new block](https://github.com/Shopify/theme-tools/pull/950/commits/ebbbf9010b35206347705e64d5b42da0daec70d7) that will hopefully handle these cases

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
```diff
@@ -63,6 +63,35 @@ export const HardcodedRoutes: LiquidCheckDefinition = {
         }
       },
 
+      async LiquidVariableOutput(node) {
```
graygilmore: I think we could maybe use `LiquidFilter` here instead? That should be the node of the actual filter and then we could do something like:

```js
if (node.name !== 'link_to') return;
```

Then we can loop over the `node.args` from there. Should simplify this a little bit!

packages/theme-check-common/src/checks/hardcoded-routes/index.spec.ts
graygilmore: Niiiice!

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
graygilmore: @charlespwd you might have a good idea here!

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
dejmedus: Yes, thank you, that looks way better!

graygilmore (approved):
Very nice!

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
```diff
@@ -0,0 +1,155 @@
...
+  '/cart/add': 'cart_add_url',
+  '/cart/change': 'cart_change_url',
+  '/cart/clear': 'cart_clear_url',
+  '/cart/update': 'cart_update_url',
+  '/collections': 'collections_url',
+  '/collections/all': 'all_products_collection_url',
+  '/customer_authentication/login': 'storefront_login_url',
+  '/recommendations/products': 'product_recommendations_url',
+  '/search': 'search_url',
+  '/search/suggest': 'predictive_search_url',
```
aswamy, line 23: I was initially wondering why we're hardcoding this instead of fetching this list from theme-liquid-docs like so:

```
        const liquidObjects = (await context.themeDocset?.objects());
        const routeObj = liquidObjects?.find(
          (obj) => obj.name === 'routes'
        );
```

Then i realized we wouldn't have the paths 😅. This seems fair for now; not sure what we'd do if we had to update this list. Don't think it's worth it to have any cross-checks here 👍.

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
aswamy: I agree with Gray. I'm not sure if we need this especially since it doesn't really know the context of the string. The javascript string could be a normal text, not a URL path. E.g.

image

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
```diff
@@ -0,0 +1,155 @@
...
+            endIndex: startIndex + route.length,
+          });
+        }
+      },
+
+      async HtmlRawNode(node) {
+        // checks for hardcoded routes, that are not commented out, inside script tags
+        if (node.name !== 'script') {
+          return;
+        }
```
aswamy: If we're doing this, we should probably also do this for `javascript` tag, right?

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
aswamy: I'd actually be down for punting this part from this PR. The `LiquidFilter` and `HtmlElement` ones are good 👍 .

If we need to parse the content inside of the script tag, we should be doing this in a language parsing area, not here.

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
dejmedus: Yeah, I'm thinking it should be possible to add paths to [objects](https://github.com/shop/world/blob/7a3f67499b4f569328674bc38967a14c0f43544f/areas/core/storefront/db/liquid/docs/objects.json#L13526) somehow. If so I will make a follow up PR to create the mapping from themeDocset

aswamy: Im pretty sure those values are dynamic that's why they aren't part of the static `objects` list. So what you have now is good 👍

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
dejmedus: Sounds good, I removed HtmlRawNode!

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
dejmedus: Ohh I see, thanks!

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
```diff
@@ -0,0 +1,100 @@
...
+  },
+};
+
+function hardcodedRoute(attr: ValuedHtmlAttribute): string | undefined {
+  if (!isNodeOfType(NodeTypes.TextNode, attr.value[0])) return;
+
+  const value = attr.value[0].value;
+
+  return HARDCODED_ROUTES.find((route) =>
+    route === '/' ? route === value && attr.value.length === 1 : value.startsWith(route),
```
aswamy, line 98: Should this be `startsWith`?

When i have a spanish store, the translation happens AFTER the domain:
image

The current code corrects when  a developer writes `/collections`. Should we also correct them when they do `/<language>/collections`? I don't know if this gets too funky or not since not every language is a 2 character short-hand. We could skip it to keep the code clean 🤷

aswamy (approved):
1 question, but the rest looks good 👍

packages/theme-check-common/src/checks/hardcoded-routes/index.ts
dejmedus: Here’s my thinking: if someone wanted to hardcode in the locale, they couldn’t use the routes object since it [dynamically appends the current site language](https://shopify.dev/docs/storefronts/themes/markets/multiple-currencies-languages#locale-aware-urls)

ex: cart link the routes obj way: 
`<a href="{{ routes.cart_url }}">{{ "templates.cart.go_to_cart" | t }}</a>` would become `/es/cart`

but if someone wants to display specific locale link options, like this [cart drawer snippet](https://github.com/adnan-travljanin18/matratzen/blob/d597d82a780cbebbc46db0752840a768e1c31dbb/snippets/cart-drawer.liquid#L355-L363)

``` liquid
<a href="/cart" class="btn m-auto text-capitalize show-cart-btn german-show-cart-btn">
    {{- 'templates.cart.View_Shopping_cart' | t -}}
</a>
<a href="/en/cart" class="btn m-auto text-capitalize show-cart-btn english-show-cart-btn">
    {{- 'templates.cart.View_Shopping_cart' | t -}}
</a>
<a href="/nl/cart" class="btn m-auto text-capitalize show-cart-btn dutch-show-cart-btn">
    {{- 'templates.cart.View_Shopping_cart' | t -}}
</a>
```
and we warned the last two `/cart` to change to `{{ routes.cart_url }}` it would break those links

(Still a trade-off because their first `/cart` would be warned, but I’m assuming they want that to go to their default locale)

Whereas stuff like [this](https://github.com/search?q=href%3D%22%2Fcollections%22+language%3ALiquid&type=code&ref=advsearch) makes sense to catch 

Am I understanding this correctly do you think?

aswamy: oh i didn't think someone would want all different languages to appear side-by-side like that 😓

I guess they _have_ to explicitly type of those routes like that and have a "disable theme check line" above the `/cart` if the dev wanted the user to go to the dev's default locale.

PR #956, opened 2025-05-29
Introduce `@shopify/theme-graph`

charlespwd:
Introduce a package that computes a data structure that represents the links between all Liquid, JSON, CSS and JS files in your theme. 

A theme graph that spans all file types is the building block we need to do the following:

- Dead code identification (is this private block used?)
- Theme aware JS bundling optimizations (only load the JS that you need on every page)
- Theme aware CSS bundling optimizations (only load the CSS that you need on every page)
- File references in VS Code extension (jump to files _using_ the snippet from the snippet)
- Link to web component definition from a Liquid file 

## Demo

https://github.com/user-attachments/assets/21d5f5b2-0c46-49ab-8aa3-b7b0e0920b13

## Explanation

See README of `@shopify/theme-graph` :)

## TODO

- [x] Figure out cache invalidation
- [x] Figure out how to refresh the graph when a file is modified/created/deleted/renamed
- [x] Presets create explicit dependencies
- [x] Figure out how to best display that data in the UI
- [x] Figure out how to lazily do the requests iff the tree views are open (maybe automatic?)
- [x] Different icons in the sidebar for non-existent dependencies files (warning!) - clicks should ask to create?
- [x] Replace preset/indirect booleans with a singular thing (type probably) since those are mutually exclusive
- [x] Video demo in PR description
- [x] Changeset setup


packages/theme-check-node/src/NodeFileSystem.ts
```diff
@@ -18,10 +18,7 @@ export const NodeFileSystem: AbstractFileSystem = {
     // console.error('fs/readDirectory', uri);
     const files = await fs.readdir(path.fsPath(uri), { withFileTypes: true });
     return files.map((file) => {
-      return [
-        `${uri}/${file.name}`.replace('file:////', 'file:///'), // fun edge case
-        file.isDirectory() ? FileType.Directory : FileType.File,
-      ];
+      return [path.join(uri, file.name), file.isDirectory() ? FileType.Directory : FileType.File];
```
charlespwd, line 21: The `/` was causing windows tests to fail, ha! and path.join() also solves the `file:////` problem.

packages/theme-check-common/src/path.ts
```diff
@@ -33,7 +33,7 @@ export function dirname(uri: UriString | URI): UriString {
 
 export function basename(uri: UriString | URI, ext?: string): string {
   const base = Utils.basename(asUri(uri));
-  return ext ? base.replace(new RegExp(`${ext}$`), '') : base;
+  return ext ? base.replace(new RegExp(`${ext.replace(/\./g, '\\.')}$`), '') : base;
```
graygilmore, line 36: Ooo nice catch.

packages/theme-graph/bin/theme-graph
```diff
@@ -0,0 +1,107 @@
...
+    process.exit(1);
+  }
+
+  if (!path.isAbsolute(root)) {
+    root = path.resolve(process.cwd(), root);
+  }
+
+  const getSourceCode = makeGetSourceCode();
+
+  // Preload files to get a sense of what is slow
```
graygilmore, line 41: Not 100% sure I understand this comment

charlespwd: It's being compared vs buildGraph. If I don't preload, then `getSourceCode` is lazy and it means that any kind of timing I get from running buildThemeGraph would be load+parse+build graph. 

By preloading first and benching separately, I can tell how fast(or slow) buildThemeGraph is independently of parsing. 

Turns out parsing is our bottleneck, not buildThemeGraph.

packages/theme-graph/src/graph/build.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  const [templates, sectionSchemas] = await Promise.all([
+    findAllFiles(deps.fs, rootUri, ([uri]) => uri.startsWith(path.join(rootUri, 'templates'))),
+    findAllFiles(
+      deps.fs,
+      rootUri,
+      ([uri]) => uri.startsWith(path.join(rootUri, 'sections')) && uri.endsWith('.liquid'),
+    ).then((sections) =>
+      Promise.all(
+        sections.map((sectionUri) => deps.getSectionSchema(path.basename(sectionUri, '.liquid'))),
```
charlespwd: Just realized I no longer need this. I initially thought that I would only consider sections with schemas that had presets as entry points, but it seems like that assumption was wrong so I no longer need to parse the schemas up front, the findAllFiles should be enough.

graygilmore (approved):

packages/theme-language-server-common/src/server/ThemeGraphManager.ts
```diff
@@ -0,0 +1,255 @@
...
+    if (!graph) return [];
+
+    const module = graph.modules[uri];
+    if (!module) return [];
+
+    const includedTypes: (AugmentedReference['type'] | undefined)[] = [
+      'direct',
+      includeIndirect ? 'indirect' : undefined,
+      includePreset ? 'preset' : undefined,
+    ];
```
aswamy, line 67: Do we need the undefined values?

```suggestion
    const includedTypes: AugmentedReference['type'][] = [
      'direct',
      ...(includeIndirect ? ['indirect'] : []),
      ...(includePreset ? ['preset'] : []),
    ];
```

packages/vscode-extension/src/common/commands.ts
```diff
@@ -0,0 +1,51 @@
...
+  });
+}
+
+export function makeDeadCode(client: BaseLanguageClient) {
+  return async function deadCode() {
+    const uri = window.activeTextEditor?.document.uri.toString();
+    if (!uri) return;
+    const [rootUri, deadCode] = await Promise.all([
+      client.sendRequest<string>('themeGraph/rootUri', { uri }),
+      client.sendRequest<string[]>('themeGraph/deadCode', { uri }),
```
aswamy: Make sure you ask around if we're okay with that terminology. Not sure if "orphanedCode" is any better.

packages/theme-language-server-common/src/server/ThemeGraphManager.ts
charlespwd: We don't... but the second option is a bit more verbose and doesn't look as good (🙈 I'm vain like that :P) 

It's going to do `['direct',undefined,undefined].includes('foo')` in the end anyway, doesn't really change anything. But option 2 creates 2 arrays that it then needs to garbage collect (micro optimization, I know...)

aswamy: > We don't... but the second option is a bit more verbose and doesn't look as good 

LMAO, fair

> option 2 creates 2 arrays that it then needs to garbage collect (micro optimization, I know...)

_a single file changes_
_recreates whole graph from scratch_

Im messing with you, but all good 🤣

charlespwd: LOL hahaha FAIR!

packages/vscode-extension/src/common/commands.ts
charlespwd: We are!

PR #961, opened 2025-06-03
Warn developers that `{% stylesheet %}` and `{% javascript %}` tags only support static content

karreiro:

Using Liquid code inside `{% stylesheet %}` or `{% javascript %}` tags, as shown below, is not allowed:

```liquid
{% stylesheet %}
  .button {
    background-color: {{ settings.button_color }};
    color: white;
  }
{% endstylesheet %}
```

This is a common and easy mistake to make, so adding linting will help developers quickly catch and address this issue.

image

## What's next? Any follow-up issues?

After this PR is merged, I will create the corresponding documentation page on shopify.dev.

## Before you deploy

  - [ ] I've made a PR to update the [shopify.dev theme check docs](https://github.com/Shopify/shopify-dev/tree/main/content/storefronts/themes/tools/theme-check/checks) if applicable (I'm working on that)

### DIFF
--- a/packages/theme-check-common/src/checks/static-stylesheet-and-javascript-tags/index.ts
+++ b/packages/theme-check-common/src/checks/static-stylesheet-and-javascript-tags/index.ts
@@ -1,0 +1,48 @@
+import { NodeTypes } from '@shopify/liquid-html-parser';
+import { LiquidCheckDefinition, Severity, SourceCodeType } from '../../types';
+
+export const StaticStylesheetAndJavascriptTags: LiquidCheckDefinition = {
+  meta: {
+    code: 'StaticStylesheetAndJavascriptTags',
+    name: 'Reports non static stylesheet and javascript tags',
+    docs: {
+      description:
+        'Reports the usage of Liquid within {% stylesheet %} and {% javascript %} tags, which should only contain static content.',
+      recommended: true,
+      url: 'https://shopify.dev/docs/storefronts/themes/tools/theme-check/checks/static-stylesheet-and-javascript-tags',
+    },
+    type: SourceCodeType.LiquidHtml,
+    severity: Severity.ERROR,
+    schema: {},
+    targets: [],
+  },
+
+  create(context) {
+    return {
+      async LiquidRawTag(node) {
+        if (node.name !== 'stylesheet' && node.name !== 'javascript') {
+          return;
+        }
+
+        const liquidNodes = node.body.nodes.filter(
+          (childNode) =>
+            childNode.type === NodeTypes.LiquidVariableOutput ||
+            childNode.type === NodeTypes.LiquidTag ||
+            childNode.type === NodeTypes.LiquidRawTag,
+        );
+
+        for (const liquidNode of liquidNodes) {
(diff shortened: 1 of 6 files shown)


packages/theme-check-common/src/checks/static-stylesheet-and-javascript-tags/index.ts
```diff
@@ -0,0 +1,48 @@
...
+        if (node.name !== 'stylesheet' && node.name !== 'javascript') {
+          return;
+        }
+
+        const liquidNodes = node.body.nodes.filter(
+          (childNode) =>
+            childNode.type === NodeTypes.LiquidVariableOutput ||
+            childNode.type === NodeTypes.LiquidTag ||
+            childNode.type === NodeTypes.LiquidRawTag,
+        );
```
charlespwd, line 32: We should use a visitor here since you could have child nodes of LiquidTag theoretically, no? This only looks at the top level? 

Maybe we are doing this backwards? Maybe we should check LiquidVariableOutput's _ancestors_ and maybe find a stylesheet/js raw tag in there?

```ts
async LiquidTag(node, ancestors) {
  const invalidParentNode = ancestors.find((parent) => parent.type === NodeTypes.LiquidRawTag && ['stylesheet', 'javascript'].includes(parent.name));
  if (invalidParentNode) {
     // report
  }
}
```

aswamy: I didn't even know we could have `stylesheet` and `javascript` inside of a `liquid` tag... How would you even attempt to use a variable INSIDE a stylesheet if it was used in a `liquid` tag?

image

I do like checking the ancestors though 🤔 Seems the safest way.

aswamy: > We should use a visitor here since you could have child nodes of LiquidTag theoretically, no? This only looks at the top level?

Wait, what edge case are we missing here?

Isn't line 23-25 already seeing if we're inside of a stylesheet or javascript tag? If we have ANY liquid node inside here, we are throwing an error.

charlespwd: I mean the read squiggly should appear on the `{{ variant }}` variable output here:
```
{% stylesheet %}
  {% if cond %}
     .class--{{ variant }} { color: blue }
  {% endif %}
{% endstylesheet %}
```

The variant wouldn't be a direct child of stylesheet, it would be a child of the if

karreiro: The ancestor approach is definitely an alternative, but the Liquid nodes will always be at the top level in the `stylesheet` and `javascript` tags so, it's actually faster to just check at that level (that was my initial reasoning). What do you think, @charlespwd @aswamy?

karreiro: 👋 I talked with @charlespwd and @aswamy on Slack, and they both agreed to keep the current approach. The reasoning is that Liquid nodes will always be at the root level of the `stylesheet` and `javascript` tags, making it faster to check them there -- so closing this comment :)

Issue #751, opened 2025-02-03
Theme check should raise error when default block does not exist


aswamy:
**Describe the bug**
Theme check does not verify if the default block file exists, or is valid within scopes defined in `blocks` within schema tag

**Source**

<img width="608" alt="Image" src="https://github.com/user-attachments/assets/d23bb700-1e08-4783-ad0e-55f1c7fe6c97" />

**Expected behaviour**

- default block should throw the following error if the block isn't allowed for this schema

<img width="1095" alt="Image" src="https://github.com/user-attachments/assets/814ca28a-a3c7-4e0c-bf2d-9f9c5606890b" />

- default block should throw the following error if the block file does not exist

<img width="854" alt="Image" src="https://github.com/user-attachments/assets/1aa91d21-e8db-4909-9601-03c1569dafb2" />

**Actual behaviour**

- no errors are thrown

PR #964, opened 2025-06-05
Fix ValidBlockTarget for blocks in `default` and `presets`

aswamy:

Closes https://github.com/Shopify/theme-tools/issues/751

Solves two separate issues
- `presets` block
  - Did not show an error when the nested block does not exist (see tophat section)
- `default` block
  - Did not show an error when the block does not exist
  - Did not show an error when the block in default is a private block, but `"blocks": [{ "type": "@theme"}]` is enabled in the top-level schema

## Tophat

In the examples below:
- `nested-non-existent`, `_meadow`, `non-existent`, and `_private_and_non_existent` blocks do not exist in blocks folder
- `demo` and `nested-demo` blocks do exist

The schema in the section containing the following code looks like this:
```
{% schema %}
{
  "name": "Demo Section",
  "blocks": [{ "type": "@theme" }],
  "presets": [
    {
      // presets code goes here
    }
  ],
  "default": {
    // default code goes here
  },
}
{% endschema %}
```

| Before | After |
| - | - |
| image | image |
| image | image |

## Before you deploy

### DIFF
--- a/.changeset/many-planets-behave.md
+++ b/.changeset/many-planets-behave.md
@@ -1,0 +1,8 @@
+---
+'@shopify/theme-check-common': patch
+---
+
+Fix ValidBlockTarget for blocks in `default` and `presets`
+
+- Solves issue so nested blocks inside of `presets` are now validated
+- Solves issue so block types inside of `default` are now validated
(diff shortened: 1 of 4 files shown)


EvilGenius13 (approved):
Thanks for making it easier to top hat with the examples.

PR #967, opened 2025-06-09
Add DeprecatedFonts theme check

miazbikowski:

Video Demo:

Demo video: https://screenshot.click/17-29-7jiob-03dcd.mp4

Outdated:

This adds a check for section and block schemas to stop usage of deprecated fonts in:
1. settings
2. preset settings
3. preset block settings (including nested blocks)

We're doing this as part of the Monotype migration project which aims to move all deprecated fonts to a replacement by September.

Yes, settings_schema.json is most likely of all files to have fonts. We need to write a theme check for this. We may also, if possible, wish to add a check for json templates and settings_data.json to thoroughly warn against deprecated font usage.

## What did you learn?
That there are no existing checks for `settings_schema.json` or the other data files, so we don't have the building blocks for these. 

## Before you deploy

  - [x] I've made a PR to update the [shopify.dev theme check docs](https://github.com/Shopify/shopify-dev/tree/main/content/storefronts/themes/tools/theme-check/checks) if applicable ([link PR here](https://app.graphite.dev/github/pr/Shopify/shopify-dev/58532/Add-DeprecatedFontsOnSectionsAndBlocks-check-docs)).


miazbikowski:
* **#967** <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/967?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/967?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/theme-check-common/src/checks/deprecated-fonts-on-sections-and-blocks/index.ts
```diff
@@ -0,0 +1,278 @@
...
+        const offset = node.blockStartPosition.end;
+
+        // 1st: check schema settings for deprecated fonts
+        checkSchemaSettingsForDeprecatedFonts(validSchema, offset, ast, context);
+
+        // 2nd: check preset settings for deprecated fonts
+        checkPresetSettingsForDeprecatedFonts(validSchema, offset, ast, context);
+
+        // 3rd: check preset blocks' settings for deprecated fonts, including nested blocks
+        await checkPresetBlocksForDeprecatedFonts(validSchema, offset, ast, context);
```
aswamy: Do we also want to do it for `default` inside of section schemas?

https://shopify.dev/docs/storefronts/themes/architecture/sections/section-schema#default

miazbikowski: > Do we also want to do it for `default` inside of section schemas?
> 
> https://shopify.dev/docs/storefronts/themes/architecture/sections/section-schema#default

Yes, good catch, I wasn't aware of default in section schemas but I'll add it.

packages/theme-check-common/src/checks/deprecated-fonts-on-sections-and-blocks/index.ts
```diff
@@ -0,0 +1,278 @@
...
+        context,
+        preset,
+        basePath,
+      );
+    }
+  }
+}
+
+// only read setting type if value is a deprecated font
+async function checkBlocksHashForDeprecatedFonts(
```
aswamy: I feel like this function and its array equivalent can be merged into one - they share most of their code. The for-loop just has to be modified a little so we check if it's an array of hash.

charlespwd: yeah I do [this](https://github.com/Shopify/theme-tools/blob/main/packages/theme-graph/src/graph/traverse.ts#L335-L362) in theme-graph to traverse both array and hash form with the same for loop. 

I'm able to get the nodeAtPath in both forms too!

```ts
    // Iterate over array entries or object entries depending on how the blocks are defined
    const iterator = Array.isArray(preset.blocks)
      ? preset.blocks.entries()
      : Object.entries(preset.blocks!);

    for (const [keyOrIndex, block] of iterator) {
      const nodePath = ['presets', i, 'blocks', keyOrIndex];
      const node = nodeAtPath(ast, nodePath)! as ObjectNode;
      
      // ...
      
      }
```

miazbikowski: Thanks for the feedback! As a ruby dev who doesn't know typescript, I don't always have the most refined typescript code. But I incorporated this approach and was able to consolidate :)

packages/theme-check-common/src/checks/deprecated-fonts-on-sections-and-blocks/index.ts
miazbikowski: section schema "default" is done :)

packages/theme-check-common/src/utils/highlighting.ts
```diff
@@ -0,0 +1,21 @@
+import { getLocEnd, getLocStart, nodeAtPath } from '..';
+import { JSONNode } from '..';
+import { Context, SourceCodeType, ArrayNode } from '../types';
+
+export function reportWarning(
+  context: Context<SourceCodeType.LiquidHtml>,
+  offset: number,
+  ast: JSONNode,
+  ast_path: string[],
```
aswamy: id keep this camelCase like the other variable

packages/theme-check-common/src/utils/highlighting.ts
aswamy, line 1: Can we name this something more appropriate? This is more about warnings, not highlighting.

packages/theme-check-common/src/checks/deprecated-fonts-on-sections-and-blocks/index.ts
```diff
@@ -0,0 +1,230 @@
...
+      setting.type === 'font_picker' &&
+      setting.default &&
+      DEPRECATED_FONT_HANDLES.has(setting.default)
+    ) {
+      const warning_ast_path = ['settings', String(index), 'default'];
+      reportWarning(
+        context,
+        offset,
+        ast as JSONNode,
+        warning_ast_path,
```
aswamy: No need to fix `SchemaPresetsBlockOrder`, but we should keep everything camelCase

packages/theme-check-common/src/checks/deprecated-fonts-on-sections-and-blocks/index.ts
```diff
@@ -0,0 +1,230 @@
...
+    let blockSchema = null; // This is used to avoid reading the schema multiple times.
+
+    // block_value is the hash which can have settings, blocks, etc.
+    for (const [setting_key, setting_value] of Object.entries(block.settings ?? {})) {
+      if (setting_value && DEPRECATED_FONT_HANDLES.has(setting_value as string)) {
+        // We found a setting value that is a deprecated font, we need to get the block schema to check the setting type is a font_picker.
+        if (!blockSchema) {
+          blockSchema = await context.getBlockSchema?.(block.type);
+          if (!blockSchema || blockSchema instanceof Error) continue;
+        }
```
aswamy: Why aren't we just fetching the block schema outside the loop?

Is it just to avoid loading the block if there are no deprecated fonts? If that's true, we don't have to really worry about it since the blocks are preloaded anyway, so it won't cost any extra cycles reading the file from the fs.

packages/theme-check-common/src/checks/deprecated-fonts-on-sections-and-blocks/index.ts
```diff
@@ -0,0 +1,230 @@
...
+
+        // 2nd: check preset settings for deprecated fonts
+        checkPresetSettingsForDeprecatedFonts(validSchema, offset, ast, context);
+
+        // 3rd: check preset blocks' settings for deprecated fonts, including nested blocks
+        await checkPresetBlocksForDeprecatedFonts(validSchema, offset, ast, context);
+
+        // 4th: check schema default (sections only) for deprecated fonts
+        if ('default' in validSchema) {
+          checkSchemaDefaultForDeprecatedFonts(validSchema as Section.Schema, offset, ast, context);
```
aswamy: I think we're also missing default block settings check

image

packages/theme-check-common/src/checks/deprecated-fonts-on-sections-and-blocks/index.ts
miazbikowski: Good to know, I'll move the reading outside the loop then :)

aswamy (approved):
🙏 thank you for your patience on this! This is amazing stuff!

packages/theme-check-common/src/utils/warnings.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export function reportWarning(
+  context: Context<SourceCodeType.LiquidHtml>,
+  offset: number,
+  ast: JSONNode,
+  ast_path: string[],
+  message: string,
+  fullHighlight: boolean = true,
+) {
+  const node = nodeAtPath(ast, ast_path)! as ArrayNode;
```
charlespwd: I feel like this casting is dangerous given this method is extracted as a util 🤔?

Why not call context.report directly and do the casting when you need it instead? 

Would remove the need for the fullHighlight thing? 

reportWarning is pretty generic and it would feel weird to try to use that in a Liquid check, I'm not sure this deserves to be extracted?

miazbikowski: Fair point re: extracting, so, un-extracting :)

PR #985, opened 2025-06-16
Gracefully handle files that are not from a theme

### LINKED ISSUE #973: findThemeRoot should throw or return undefined when it can't find one
**Describe the bug**
I think findThemeRootURI assumes that it will always find a root. That's not true. Say you are in your theme and you take a look at your vscode's settings.json file, you won't be able to find the theme root from that URI. 

That assumption is wrong and is unfortunately used everywhere in the stack. 

We probably should change the return type of `findThemeRootURI` to possibly return `undefined` when it can't find a root.

Then we'll need to adjust all the call sites and add handling for when that happens. Usually that means returning early, returning no results, etc. 

It's too big of a task to include in #956 

That kind of assumption is what causes the dependency graph to load 4000 files when the focused file is settings.json in $HOME/Library/...

charlespwd:
Make findThemeRootURI return null when it can't find a root

Make everywhere handle null when no root is found

Should fix "preloading" of theme when looking at `$HOME/Library/Application Support/Code/User/settings.json`

Also, `snippets` can be found in `$HOME/Library/Application Support/Code/User` and is a VS Code-owned, non theme thing. So had to modify the findRoot algo to search for assets _and_ snippets folder existence.

Fixes #973

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/checks/valid-visible-if/visible-if-utils.ts
+++ b/packages/theme-check-common/src/checks/valid-visible-if/visible-if-utils.ts
@@ -1,13 +1,6 @@
-import {
-  type LiquidVariableLookup,
-  toLiquidHtmlAST,
-  NodeTypes,
-  LiquidVariableOutput,
-} from '@shopify/liquid-html-parser';
+import { type LiquidVariableLookup, NodeTypes, toLiquidHtmlAST } from '@shopify/liquid-html-parser';
 import { Context, SourceCodeType } from '../..';
-import { findRoot } from '../../find-root';
 import { parseJSON } from '../../json';
-import { join } from '../../path';
 import { visit } from '../../visitor';
 
 export type Vars = { [key: string]: Vars | true };
@@ -140,11 +133,8 @@ export async function getGlobalSettings(context: Context<SourceCodeType>) {
   const globalSettings: string[] = [];
 
   try {
-    const path = join(
-      await findRoot(context.file.uri, context.fileExists),
-      'config/settings_schema.json',
-    );
-    const settingsFile = await context.fs.readFile(path);
+    const uri = context.toUri('config/settings_schema.json');
+    const settingsFile = await context.fs.readFile(uri);
     const settings = parseJSON(settingsFile);
     if (Array.isArray(settings)) {
       for (const group of settings) {
--- a/packages/theme-check-common/src/find-root.ts
+++ b/packages/theme-check-common/src/find-root.ts
@@ -7,12 +7,19 @@ async function isRoot(dir: UriString, fileExists: FileExists) {
   return or(
     fileExists(path.join(dir, 'shopify.extension.toml')), // for theme-app-extensions
     fileExists(path.join(dir, '.theme-check.yml')),
-    fileExists(path.join(dir, '.git')),
 
-    // zip files and TAEs might not have config files, but they should have a
-    // snippets directory but in case they do specify a .theme-check.yml a
+    // Maybe the root of the workspace has an assets + snippets directory, we'll accept that
+    and(
+      fileExists(path.join(dir, '.git')),
+      fileExists(path.join(dir, 'assets')),
+      fileExists(path.join(dir, 'snippets')),
+    ),
+
+    // zip files and TAEs might not have config files, but they should have an
+    // assets & snippets directory but in case they do specify a .theme-check.yml a
     // couple of directories up, we should respect that
     and(
+      fileExists(path.join(dir, 'assets')),
       fileExists(path.join(dir, 'snippets')),
       not(fileExists(path.join(path.dirname(dir), '.theme-check.yml'))),
       not(fileExists(path.join(path.dirname(path.dirname(dir)), '.theme-check.yml'))),
@@ -49,7 +56,7 @@ async function not(ap: Promise<boolean>) {
  * Note: that this is not the theme root. The config file might have a `root` entry in it
  * that points to somewhere else.
  */
-export async function findRoot(curr: UriString, fileExists: FileExists): Promise<UriString> {
+export async function findRoot(curr: UriString, fileExists: FileExists): Promise<UriString | null> {
   const currIsRoot = await isRoot(curr, fileExists);
   if (currIsRoot) {
     return curr;
@@ -58,7 +65,7 @@ export async function findRoot(curr: UriString, fileExists: FileExists): Promise
   const dir = path.dirname(curr);
   const currIsAbsoluteRoot = dir === curr;
   if (currIsAbsoluteRoot) {
-    return curr;
+    return null; // Root not found.
   }
 
   return findRoot(dir, fileExists);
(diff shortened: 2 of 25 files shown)


packages/theme-check-common/src/checks/valid-visible-if/visible-if-utils.ts
```diff
@@ -140,11 +133,8 @@ export async function getGlobalSettings(context: Context<SourceCodeType>) {
   const globalSettings: string[] = [];
 
   try {
-    const path = join(
-      await findRoot(context.file.uri, context.fileExists),
-      'config/settings_schema.json',
-    );
-    const settingsFile = await context.fs.readFile(path);
```
charlespwd, line 147: toURI does this better.

packages/theme-check-common/src/test/MockFileSystem.ts
```diff
@@ -46,32 +48,37 @@ export class MockFileSystem implements AbstractFileSystem {
...
+      let current = result;
+      for (let i = 0; i < segments.length - 1; i++) {
+        const segment = segments[i];
+        current[segment] ??= {};
+        current = current[segment] as FileTree;
       }
+      current[segments[segments.length - 1]] = source;
     }
-    return this._fileTree;
+    return result;
```
charlespwd, line 81: Had to remove memoization because of a test that mutated this.mockTheme. It's only for tests, perf doesn't matter :)

packages/theme-language-server-common/src/diagnostics/runChecks.ts
```diff
@@ -43,7 +43,7 @@ export function makeRunChecks(
     // then we recheck theme1
     const fileExists = makeFileExists(fs);
     const rootURIs = await Promise.all(triggerURIs.map((uri) => findRoot(uri, fileExists)));
-    const deduplicatedRootURIs = new Set(rootURIs);
+    const deduplicatedRootURIs = new Set<string>(rootURIs.filter((x): x is string => !!x));
     await Promise.all([...deduplicatedRootURIs].map(runChecksForRoot));
```
charlespwd, line 47: Impact: runChecks won't run when triggered from a file that you can't find a root for.

packages/theme-language-server-common/src/documentLinks/DocumentLinksProvider.ts
```diff
@@ -24,6 +25,10 @@ export class DocumentLinksProvider {
     }
 
     const rootUri = await this.findThemeRootURI(uriString);
+    if (!rootUri) {
+      return [];
+    }
+
```
charlespwd, line 31: No document links when you can't find a root.

packages/theme-language-server-common/src/rename/providers/LiquidVariableRenameProvider.ts
```diff
@@ -71,7 +72,7 @@ export class LiquidVariableRenameProvider implements BaseRenameProvider {
     const rootUri = await this.findThemeRootURI(params.textDocument.uri);
     const textDocument = document?.textDocument;
 
-    if (!textDocument || !node || !ancestors) return null;
+    if (!rootUri || !textDocument || !node || !ancestors) return null;
```
charlespwd, line 75: No variable rename when you can't find a root.

packages/theme-language-server-common/src/renamed/handlers/AssetRenameHandler.ts
```diff
@@ -44,6 +45,7 @@ export class AssetRenameHandler implements BaseRenameHandler {
     if (relevantRenames.length !== 1) return;
     const rename = relevantRenames[0];
     const rootUri = await this.findThemeRootURI(path.dirname(params.files[0].oldUri));
+    if (!rootUri) return;
```
charlespwd, line 48: No asset rename when you can't find a root

packages/theme-language-server-common/src/server/ThemeGraphManager.ts
```diff
@@ -33,11 +34,15 @@ export class ThemeGraphManager {
...
     private documentManager: DocumentManager,
     private fs: AbstractFileSystem,
-    private findThemeRootURI: (uri: string) => Promise<string>,
+    private findThemeRootURI: FindThemeRootURI,
   ) {}
 
   async getThemeGraphForURI(uri: string) {
     const rootUri = await this.findThemeRootURI(uri);
+    if (!rootUri) {
+      return undefined;
```
charlespwd, line 43: No theme graph for files that are not part of a theme

packages/theme-language-server-common/src/server/startServer.spec.ts
```diff
@@ -58,7 +58,7 @@ describe('Module: server', () => {
       }
     });
 
-    fileTree = { 'snippets/code.liquid': fileContents };
+    fileTree = { '.theme-check.yml': '', 'snippets/code.liquid': fileContents };
```
charlespwd, line 61: Since `Application Support/Code/User/snippets` exists and since `Application Suport/Code/User/settings.json` is a file you're likely to edit, we were inferring `Application Support/Code/User` to be the root of a theme. That was not very good. So I changed the algo to look for `snippets` _and_ `assets` (as well as looking for `.theme-chekc.yml`).

For the unit tests, it was easiest to simply put a `.theme-check.yml`

packages/theme-language-server-common/src/server/startServer.ts
```diff
@@ -158,8 +160,13 @@ export function startServer(
...
   );
 
-  const getMetafieldDefinitions = async (uri: string) => {
+  const getMetafieldDefinitions: NonNullable<Dependencies['getMetafieldDefinitions']> = async (
+    uri: string,
+  ) => {
     const rootUri = await findThemeRootURI(uri);
+    if (!rootUri) {
+      return {} as MetafieldDefinitionMap;
+    }
```
charlespwd, line 169: no root URI means no `.shopify` folder gets generated in non-themes.

packages/theme-check-common/src/find-root.ts
```diff
@@ -7,12 +7,19 @@ async function isRoot(dir: UriString, fileExists: FileExists) {
...
     fileExists(path.join(dir, '.theme-check.yml')),
-    fileExists(path.join(dir, '.git')),
 
-    // zip files and TAEs might not have config files, but they should have a
-    // snippets directory but in case they do specify a .theme-check.yml a
+    // Maybe the root of the workspace has an assets + snippets directory, we'll accept that
+    and(
+      fileExists(path.join(dir, '.git')),
+      fileExists(path.join(dir, 'assets')),
+      fileExists(path.join(dir, 'shippets')),
```
aswamy: `shippets`? 😅

packages/theme-language-server-common/src/server/startServer.ts
aswamy: Metafields file never got created on my system but i feel like it's a problem outside of the rootUri... can you check yours to see if it's working. It's failing on the `exec` function in `packages/theme-language-server-node/src/metafieldDefinitions.ts` line 46.

packages/theme-check-common/src/find-root.ts
charlespwd: oh god

packages/theme-language-server-common/src/server/startServer.ts
charlespwd: They load fine for me 🤔

Issue #981, opened 2025-06-13
theme graph does check `inline_asset_content` filter


aswamy:
**Describe the bug**

- When you open an SVG file in the editor, it doesn't show any entries in the "References" panel
- When you open a file that uses an SVG (e.g. `inline_asset_content`), it doesn't show up in the "Dependencies" panel

NOTE: Please check other filters we might have missed for loading assets

**Source**

![Image](https://github.com/user-attachments/assets/c12a19f2-b69a-4e45-ae45-93f7edc4a1a5)

![Image](https://github.com/user-attachments/assets/1840a3e8-341b-4bfa-823f-43e8f9af2c59)

**Expected behaviour**
- [ ] When you open the svg file, it should show its references
- [ ] When you open a file that imports the svg, it should show up as a dependency

**Actual behaviour**
- it doesn't do ^

PR #988, opened 2025-06-18
Support for SVGs and images in theme-graph

aswamy:

Fixes https://github.com/Shopify/theme-tools/issues/981
- Added support for SVGs and Images on theme-graph

## Tophat

- Check out branch
- Press F5
- Open a theme repo
- Find a theme file that calls out an SVG or an image file in the assets folder
- It should appear in the Dependencies panel 

image

## Before you deploy

### DIFF
--- a/packages/theme-graph/src/graph/traverse.ts
+++ b/packages/theme-graph/src/graph/traverse.ts
@@ -73,7 +73,9 @@ export async function traverseModule(
       return; // TODO graph import/exports ?
     }
 
-    case ModuleType.Css: {
+    case ModuleType.Css:
+    case ModuleType.Svg:
+    case ModuleType.Image: {
       return; // Nothing to do??
     }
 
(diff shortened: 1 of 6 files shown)


packages/theme-graph/src/graph/traverse.ts
```diff
@@ -97,9 +99,10 @@ async function traverseLiquidModule(
     { target: ThemeModule; sourceRange: Range; targetRange?: Range }
   > = {
     // {{ 'theme.js' | asset_url }}
-    // {{ 'theme.css' | asset_url }}
+    // {{ 'image.png' | asset_img_url }}
+    // {{ 'icon.svg' | inline_asset_content }}
     LiquidFilter: (node, ancestors) => {
-      if (node.name === 'asset_url') {
+      if (['asset_url', 'asset_img_url', 'inline_asset_content'].includes(node.name)) {
```
aswamy, line 105: These are the only filters i knew that reference images or svgs that are in assets folder, let me know if i missed anything

packages/theme-graph/src/types.ts
```diff
@@ -136,13 +159,29 @@ export const enum LiquidModuleKind {
   Template = 'template',
 }
 
+export const SUPPORTED_ASSET_IMAGE_EXTENSIONS = ['jpg', 'jpeg', 'png', 'gif', 'webp', 'heic'];
```
aswamy: Im not sure if we support other image assets in the asset folder, but could add more if anyone knows more

EvilGenius13 (approved):
Sweet! 🎩 'd the changes and could see dependencies including images.

packages/theme-graph/src/types.ts
EvilGenius13: Would we ever want `.ico`?

packages/theme-graph/src/types.ts
```diff
@@ -24,7 +24,14 @@ export interface IDependencies {
...
   getSourceCode?: (
     uri: UriString,
-  ) => Promise<JSONSourceCode | LiquidSourceCode | JsSourceCode | CssSourceCode>;
+  ) => Promise<
+    | JSONSourceCode
+    | LiquidSourceCode
+    | JsSourceCode
+    | CssSourceCode
+    | ImageSourceCode
+    | SvgSourceCode
```
charlespwd: Should probably type alias this union :P

Issue #536, opened 2024-10-22
Feature Request: Add Emmet support


david-w-shopify:
**Is your feature request related to a problem? Please describe.**
Writing HTML is slow, and would be much faster if I could use emmet expansions.

**Describe the solution you'd like**
Support for Emmet expansions (examples here https://code.visualstudio.com/docs/editor/emmet).

As a basic example, I could then type 

```
.my-element#id
```

After pressing tab, it would be expanded to

```html
<div class=”my-element” id=”id”></div>
```

charlespwd:
It seems like we don't need to do anything. The following `.vscode/config.json` works (This is what the other extension does:)

```jsonc
{
      "emmet.triggerExpansionOnTab": true,
      "emmet.includeLanguages": {
        "liquid": "html"
      }
}
```

There's just one thing I don't like though: this adds the emmet completion options _everywhere_. Even when you're typing Liquid code. They also appear to have a very high priority in the completion window. 

![Image](https://github.com/user-attachments/assets/d14fa2a0-4851-4508-a542-291785beb139)

Maybe we simply tell people through this issue that you can turn that on yourself? There would be 0 code changes other than switching this to on by default with the extension.

david-w-shopify:
🤦 Wish I'd realised it was that simple.

Also I'd never noticed the completions being included within Liquid contexts before. Can't be unseen and I totally agree that I don't like that. I imagine letting people know they can enable it this way if they want to, with this caveat is fine for now, but think it would be better enabled by default in the future _if_ the issue you raised can be resolved.

PR #991, opened 2025-06-27
Added support for emmet for liquid files in extentions

NzakiCodes:
fixes: #536


m74278803-cmyk (commented):
#1094 NzakiCodes:emmet-default-support-vscode-extention

PR #1023, opened 2025-08-12
Update our parsing rules to support boolean expressions

aswamy:
Part of https://github.com/Shopify/developer-tools-team/issues/807

- Update parser so we can detect boolean expressions in theme code
- Auto-fix boolean expressions when used in assigns, echos, liquid output tag `{{ .. }}`
- When you hover over a variable that's been assigned a boolean expression, we will identify it as a boolean (even though it isn't supported)

- Find and fix more issues where we allow invalid syntax when we shouldn't

## Tophat

Main feature
- Run this branch
- `yarn install`
- `yarn build`
- F5
- Open a theme
- Create a liquid file with some boolean expressions (see below)
- Run fix and the latter portion of the logical expression or conditional expression be erased
  - If there are multiple expressions, remove all but the first expression

```
{% # liquid output %}
{{ 'example1' == var }}
{{ 'example1' == 'example2' }}
{{ var == 'example2' }}
{{ var1 == var2 }}

{% # echos %}
{% echo 'example1' == var %}
{% echo 'example1' == 'example2' %}
{% echo var == 'example1' %}
{% echo var1 == var2 %}

{% # assigns %}
{% assign var3 = 'example1' == var %}
{% assign var4 = 'example1' == 'example2' %}
{% assign var5 = var == 'example2' %}
{% assign var6 = var1 == var2 %}

{% # Checking inside a liquid block %}
{% liquid
  assign var7 = true
  assign var7 = true == true
  assign var7 = what == '123'
  assign var7 = this and that == 8

  echo true
  echo true == true
  echo what == '123'
  echo this and that == 8

  echo a == b and c == d
  echo sand and c or do
  echo 'test' or 'something' > 5
%}

{% # This is not considered a boolean expression - and we won't be fixing them since they are supported today %}
{% if something > 5 %}{% endif %}
{% if a and b %}{% endif %}
```

Side features
- Hover over a variable with a boolean expression, and it should say it's a boolean
image

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/checks/liquid-html-syntax-error/index.ts
+++ b/packages/theme-check-common/src/checks/liquid-html-syntax-error/index.ts
@@ -1,4 +1,5 @@
-import { Severity, SourceCodeType, LiquidCheckDefinition } from '../../types';
+import { NodeTypes } from '@shopify/liquid-html-parser';
+import { Severity, SourceCodeType, LiquidCheckDefinition, Check, Context } from '../../types';
 import { getOffset, isError } from '../../utils';
 
 type LineColPosition = {
@@ -34,25 +35,29 @@ export const LiquidHTMLSyntaxError: LiquidCheckDefinition = {
   },
 
   create(context) {
-    const error = context.file.ast;
-    if (!isError(error)) return {};
+    const ast = context.file.ast;
+    if (!isError(ast)) {
+      return {
+        ...detectInvalidBooleanExpressions(context),
+      };
+    }
 
     return {
       async onCodePathStart(file) {
-        if (isParsingErrorWithLocation(error)) {
-          const { start, end } = error.loc;
+        if (isParsingErrorWithLocation(ast)) {
+          const { start, end } = ast.loc;
           const startIndex = getOffset(file.source, start.line, start.column);
           let endIndex = getOffset(file.source, end.line, end.column);
           if (startIndex === endIndex) endIndex += 1;
           const highlight = file.source.slice(startIndex, endIndex);
           context.report({
-            message: cleanErrorMessage(error.message, highlight),
+            message: cleanErrorMessage(ast.message, highlight),
             startIndex,
             endIndex: endIndex,
           });
         } else {
           context.report({
-            message: error.message,
+            message: ast.message,
             startIndex: 0,
             endIndex: file.source.length,
           });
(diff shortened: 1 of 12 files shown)


packages/theme-check-common/src/checks/liquid-html-syntax-error/index.ts
```diff
@@ -61,3 +66,33 @@ export const LiquidHTMLSyntaxError: LiquidCheckDefinition = {
...
+        condition.type !== NodeTypes.Comparison &&
+        condition.type !== NodeTypes.LogicalExpression
+      ) {
+        return;
+      }
+
+      context.report({
+        message: 'Syntax is not supported',
+        startIndex: node.position.start,
+        endIndex: node.position.end,
```
aswamy, line 87: Maybe the underline should only be under the part we are reporting as "invalid syntax" 🤔

```suggestion
        startIndex: condition.right.position.start,
        endIndex: condition.right.position.end,
```

graygilmore: I have waffled a few times on my reply to this. I think I like it as-is?

graygilmore (approved):
🎩 went great though I won't claim to be the expert here. I wonder if a pairing session to go over the ohm/cst/ast changes would be good?

packages/theme-check-common/src/checks/liquid-html-syntax-error/index.ts
```diff
@@ -61,3 +66,33 @@ export const LiquidHTMLSyntaxError: LiquidCheckDefinition = {
...
+  return {
+    async BooleanExpression(node) {
+      const condition = node.condition;
+
+      if (
+        condition.type !== NodeTypes.Comparison &&
+        condition.type !== NodeTypes.LogicalExpression
+      ) {
+        return;
+      }
```
karreiro, line 82: It's excellent that we're identifying comparison and logical expressions, but since `{{ a == b }}` is just as invalid as `{{ a === b }}` today, I think we should lint both. Otherwise, we may lead users on thinking the second scenario is valid:

scenario

karreiro (approved):
Thank you, @aswamy! Excellent stuff 🔥 I've left only one comment, we may apply in a following PR :)

EvilGenius13 (approved):

packages/theme-check-common/src/checks/liquid-html-syntax-error/index.ts
aswamy: I really don't want to add `===` to our grammar just so we can parse it. This will pollute our grammar files. We might just do some fancy regex to find unsupported comparison operators like `!==` and `===` in the follow-up PR.

aswamy:
> I wonder if a team pairing session to go over the ohm/cst/ast changes would be good?

Josh and I will be working on a few of these this week - maybe we can setup a mid-week session so whoever is interested in these can join.

PR #1036, opened 2025-08-26
Conditional boolean check

EvilGenius13:

Adding additional logic to look for invalid boolean expressions within if/else/unless statements.
Liquid will ignore certain patterns when running in lax mode. We want to surface those patterns and offer fixes.
The patterns are identified in the file but examples are:

Pattern 1: Left-side evaluation - parser stops at truthy value, ignoring rest
`{% if 7 1 > 100 %} → stops at 7, ignores  1 > 100`
Pattern 2: Malformed expressions starting with invalid tokens
`{% if > 2 %} → starts with invalid operator >`
Pattern 3: Trailing junk after complete comparisons
`{% if 1 == 2 foobar %} → ignores trailing foobar`

There was also an object collision issue since some of our checks are returning data for the same type of branch. Added a new utility to ensure we don't have that problem.

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidConditionalNode.ts
+++ b/packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidConditionalNode.ts
@@ -1,11 +1,34 @@
 import { LiquidBranch, LiquidTag } from '@shopify/liquid-html-parser';
-import { Check, Context, SourceCodeType, Problem } from '../../..';
+import { SourceCodeType, Problem } from '../../..';
+import { getValuesInMarkup } from './utils';
 
-const LeftSideEvaluation =
-  /^(['"][^'"]*['"]|\d+(?:\.\d+)?|true|false|nil|empty|blank)\s+((?![a-zA-Z_])\S.*)/;
-const TrailingJunk = /^(.+?(?:==|!=|>=|<=|>|<|contains)\s+\S+)\s+(.+?)(?:\s+(?:and|or)\s|$)/;
-const Comparison = /^\s*(==|!=|>=|<=|>|<|contains)\s/;
-const LogicalOperator = /^(?:and|or)\s/;
+type TokenType = 'variable' | keyof typeof TOKEN_PATTERNS;
+
+interface Token {
+  value: string;
+  type: TokenType;
+}
+
+interface ExpressionIssue {
+  message: string;
+  fix: string;
+}
+
+const TOKEN_PATTERNS = {
+  logical: /^(and|or)$/,
+  comparison: /^(==|!=|>=|<=|>|<|contains)$/,
+  invalid: /^[@#$&]$/,
+  literal: /^(['"][^'"]*['"]|\d+(?:\.\d+)?|true|false|nil|empty|blank)$/,
+} as const;
+
+function classifyToken(value: string): TokenType {
+  for (const [type, pattern] of Object.entries(TOKEN_PATTERNS)) {
+    if (pattern.test(value)) {
+      return type as TokenType;
+    }
+  }
+  return 'variable';
+}
 
 export function detectInvalidConditionalNode(
   node: LiquidBranch | LiquidTag,
@@ -16,7 +39,7 @@ export function detectInvalidConditionalNode(
   const markup = node.markup;
   if (typeof markup !== 'string' || !markup.trim()) return;
 
-  const issue = checkLaxParsing(markup);
+  const issue = analyzeConditionalExpression(markup);
   if (!issue) return;
 
   const openingTagRange = node.blockStartPosition || node.position;
@@ -28,7 +51,7 @@ export function detectInvalidConditionalNode(
   const endIndex = startIndex + markup.length;
 
   return {
-    message: `Liquid lax parsing issue: ${issue.issue}`,
+    message: `Liquid lax parsing issue: ${issue.message}`,
     startIndex,
     endIndex,
     fix: (corrector) => {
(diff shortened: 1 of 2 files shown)


packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidConditionalBooleanExpression.ts
```diff
@@ -0,0 +1,91 @@
...
+const LeftSideEvaluation =
+  /^(['"][^'"]*['"]|\d+(?:\.\d+)?|true|false|nil|empty|blank)\s+((?![a-zA-Z_])\S.*)/;
+const TrailingJunk = /^(.+?(?:==|!=|>=|<=|>|<|contains)\s+\S+)\s+(.+?)(?:\s+(?:and|or)\s|$)/;
+const Comparison = /^\s*(==|!=|>=|<=|>|<|contains)\s/;
+const LogicalOperator = /^(?:and|or)\s/;
+
+export function detectInvalidConditionalBooleanExpression(
+  context: Context<SourceCodeType.LiquidHtml>,
+): Check<SourceCodeType.LiquidHtml> {
+  function checkLaxParsing(markup: string): { issue: string; fix: string } | null {
```
aswamy: let's move this function and the one below outside parent function

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidConditionalBooleanExpression.ts
```diff
@@ -0,0 +1,91 @@
+import { Check, Context, SourceCodeType } from '../../..';
+
+const LeftSideEvaluation =
```
aswamy: I did a few others tests around if-statements and still failed for certain scenarios. Not sure how likely it is, but here they are:

```
{% if wat | something == something %}
  {{ 'ello' }}
{% endif %}

{% if "str" blue == something %}
  {{ 'ello' }}
{% endif %}

{% if red blue > something %}
  {{ 'ello' }}
{% endif %}
```

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidConditionalBooleanExpression.ts
```diff
@@ -0,0 +1,91 @@
...
+          issue: `Trailing tokens ignored after comparison: '${trailing.trim()}'`,
+          fix: comparison.trim(),
+        };
+      }
+    }
+
+    return null;
+  }
+
+  async function checkConditional(node: any) {
```
aswamy: After rebasing, get rid of all these `any` types unless absolutely necessary

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidConditionalBooleanExpression.ts
aswamy, line 1: Let's not call this `boolean expression` since we are going to use it for expressions like `a == b` and `a and b` outside conditional statements. Maybe `InvalidConditionalNode`.

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidConditionalBooleanExpression.ts
EvilGenius13: Okay I added them all as tests in the test suite. 

The first passes lax parsing as false, should this have been caught?
Two and Three would both be syntax or unknown operator errors even in lax mode so I don't catch those.

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidConditionalNode.ts
```diff
@@ -0,0 +1,142 @@
+import { LiquidBranch, LiquidTag } from '@shopify/liquid-html-parser';
+import { Check, Context, SourceCodeType, Problem } from '../../..';
+import { getValuesInMarkup } from './utils';
+
+type TokenType = 'literal' | 'variable' | 'comparison' | 'logical' | 'invalid';
```
aswamy:
```suggestion
type TokenType = 'variable' | keyof typeof TOKEN_PATTERNS;
```

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidConditionalNode.ts
```diff
@@ -0,0 +1,142 @@
+import { LiquidBranch, LiquidTag } from '@shopify/liquid-html-parser';
+import { Check, Context, SourceCodeType, Problem } from '../../..';
```
aswamy:
```suggestion
import { SourceCodeType, Problem } from '../../..';
```

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidConditionalNode.ts
```diff
@@ -0,0 +1,142 @@
...
+        !remaining.some(isOperatorToken)) {
+
+      const validExpr = tokens.slice(0, i + COMPARISON_LENGTH).map(t => t.value).join(' ');
+      const junk = remaining.map(t => t.value).join(' ');
+      return {
+        message: `Trailing tokens ignored after comparison: '${junk.trim()}'`,
+        fix: validExpr,
+      };
+    }
+  }
```
aswamy: I feel like the assumption of minTokenForTrailing is incorrect here. It fails to capture when there is a missing operator AFTER a variable/literal:

```
{% # mistyped `and` %}
{% if "wat" == "squat"  adn "wat" == "squat" %}
  Hello, world
{% endif %}

{% # non-existing logical operator %}
{% if "wat" == "squat"  xor "wat" == "squat" %}
  Hello, world
{% endif %}
```

aswamy (commented):
I assume we're going to do a v2 of this PR where we remove `&&` and `||` operators. For now, we need to at least fix those two scenarios i mentioned before merging.

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidConditionalNode.ts
```diff
@@ -0,0 +1,144 @@
...
+    const [v1, op, v2] = tokens.slice(i, i + 3);
+    const remaining = tokens.slice(i + 3);
+
+    if (isValueToken(v1) && op.type === 'comparison' && isValueToken(v2)) {
+      if (remaining.length > 0) {
+        if (remaining[0].type !== 'logical') {
+          const validExpr = tokens.slice(0, i + 3).map(t => t.value).join(' ');
+          const junk = remaining.map(t => t.value).join(' ');
+          return {
+            message: `Trailing tokens ignored after comparison: '${junk.trim()}'`,
```
aswamy, line 99: @karreiro we should all should unify how we want to report these error messages. For boolean expressions, we have a generic messaged like "Syntax is unsupported". Do we want that for ALL liquid syntax errors?

aswamy (approved):
Amazing stuff! Thank you for all the iterations and thorough testing in the test files 🙏 

I did more testing with our liquid gem to see what else is allowed with the lax parser that should be corrected, but i think you caught them all.

PR #1037, opened 2025-08-26
Find/fix multiple `echo` values

aswamy:

Part of https://github.com/Shopify/developer-tools-team/issues/812#issue-3339001872

- Find and fix bad use of `echo` where multiple values are provided
  - fix should remove all values past the first one (since that's what the Liquid gem does)
  - the fix will be strict parsable
  - also supports liquid output version of "echoing" `{{ }}`

## Tophat
- yarn install
- yarn build
- F5
- Try a few invalid syntax and see auto-fix erases the bad part of echos:
```
{% echo var1 var2 %}
{{ output1 output2 }}
{% echo var1 var2 | upcase %}
{{ output1 output2 | upcase }}
```
image

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/MultipleEchoValues.ts
+++ b/packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/MultipleEchoValues.ts
@@ -1,0 +1,68 @@
+import { LiquidTag, LiquidVariableOutput, NodeTypes } from '@shopify/liquid-html-parser';
+import { Problem, SourceCodeType } from '../../..';
+import { ensureValidAst, getFirstValueInMarkup, INVALID_SYNTAX_MESSAGE } from './utils';
+
+export function detectMultipleEchoValues(
+  node: LiquidTag | LiquidVariableOutput,
+): Problem<SourceCodeType.LiquidHtml> | undefined {
+  // We've broken it up into two groups:
+  // 1. The variable(s)
+  // 2. The filter section (non-captured)
+  const ECHO_MARKUP_REGEX = /([^|]*)(?:\s*\|\s*.*)?$/m;
+
+  if (node.type === NodeTypes.LiquidTag && node.name !== 'echo') {
+    return;
+  }
+
+  const markup = node.markup;
+
+  if (typeof markup !== 'string') {
+    return;
+  }
+
+  const match = markup.match(ECHO_MARKUP_REGEX);
+  if (!match) {
+    return;
+  }
+
+  const [, echoValue] = match;
+
+  const firstEchoValue = getFirstValueInMarkup(echoValue);
+
+  if (!firstEchoValue) {
+    return;
+  }
+
+  const removalIndices = (source: string) => {
+    const offset = source.indexOf(markup);
+
+    return {
+      startIndex: offset + firstEchoValue.length,
+      endIndex: offset + echoValue.trimEnd().length,
+    };
+  };
+
+  const tagSource = node.source.slice(node.position.start, node.position.end);
+  const { startIndex: tagSourceRemovalStartIndex, endIndex: tagSourceRemovalEndIndex } =
+    removalIndices(tagSource);
+
+  if (
+    !ensureValidAst(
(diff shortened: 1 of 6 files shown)


packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/MultipleEchoValues.ts
```diff
@@ -0,0 +1,68 @@
...
+  const markup = node.markup;
+
+  if (typeof markup !== 'string') {
+    return;
+  }
+
+  const match = markup.match(ECHO_MARKUP_REGEX);
+  if (!match) {
+    return;
+  }
```
karreiro, line 26: (optional) I'd personally call this variable, `hasValue`

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/MultipleEchoValues.ts
```diff
@@ -0,0 +1,68 @@
...
+import { Problem, SourceCodeType } from '../../..';
+import { ensureValidAst, getFirstValueInMarkup, INVALID_SYNTAX_MESSAGE } from './utils';
+
+export function detectMultipleEchoValues(
+  node: LiquidTag | LiquidVariableOutput,
+): Problem<SourceCodeType.LiquidHtml> | undefined {
+  // We've broken it up into two groups:
+  // 1. The variable(s)
+  // 2. The filter section (non-captured)
+  const ECHO_MARKUP_REGEX = /([^|]*)(?:\s*\|\s*.*)?$/m;
```
karreiro, line 11: Is this regex extracted from `Shopify/liquid`?

It seems like it's something we'd extract from [lib/liquid/variable.rb](https://github.com/Shopify/liquid/blob/9bb7fbf123e6e2bd61e00189b1c83159f375d3f3/lib/liquid/variable.rb#L15-L19) because the [echo implementation](https://github.com/Shopify/liquid/blob/9bb7fbf123e6e2bd61e00189b1c83159f375d3f3/lib/liquid/tags/echo.rb#L26). It seems like they are very similar, but I wanted to double-check to sure our regex it's capturing what we want :)

karreiro (commented):
Thank you, @aswamy! Great stuff 🔥  I've left only one minor question about the one regex :)

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/MultipleEchoValues.ts
aswamy: I would if it returned a boolean - but it returns a `RegExpMatchArray | null`

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/MultipleEchoValues.ts
aswamy: Originally i tried to copy exactly the regexes from Shopify/liquid, but because of how the variables are stored across multiple files like lib/liquid.rb and lib/liquid/variable.rb, i just copied the "un-ruby-fied" + "compiled" version of the regex. The reason ours has extra `*` characters are more because of how we're doing error tolerance.

On my next PR I will want to capture when there is an empty string passed into an echo tag (e.g. `{{  | upcase }}`). Thus, i put a `*` after `[^|]` even though _strict_ liquid would have `+`.

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/MultipleEchoValues.ts
karreiro: Ohh! I see -- thank you for the clarification, @aswamy!

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/MultipleEchoValues.ts
karreiro: Nice! Thanks for the extra context, @aswamy!

PR #1038, opened 2025-08-28
Find+fix echo tags & variable output with no value

aswamy:

Part of https://github.com/Shopify/developer-tools-team/issues/812#issue-3339001872

- Find and fix empty `echo` values
  - If no value is provided but still has filters, just replace it with `blank`
  - If no value is provided and no filters are provided, we can keep it as is

## Tophat
- yarn install
- yarn build
- F5
- Try a few invalid syntax and see auto-fix patches the echo statements:
```
{% echo | upcase %}
{{ | upcase }}
{% liquid echo | upcase %}
```

## Before you deploy

### DIFF
--- a/.changeset/hip-scissors-accept.md
+++ b/.changeset/hip-scissors-accept.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-check-common': patch
+---
+
+Find+fix echo tags & variable output with no value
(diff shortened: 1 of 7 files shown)


EvilGenius13 (approved):
I think this util could help in other areas. I tried one for conditionals `'foo 1 > 100 baz bux'` and it grabbed everything.

PR #1040, opened 2025-09-03
Detect and report invalid ranges in loops

aswamy:

- Detect invalid use of ranges in loops (`for` and `tablerow` tags)
- Strict parser currently allows spaces within ranges (e.g. valid: `(     1..5)`, `(1     ..    5)`). The lax parser doesn't complain but does fail to render this. This is usually the reverse of what happens between strict/lax parsing. For consistency, we will just force all ranges to exist in the format `(<start>..<end>)`
  - This is why some times the markup is a `string` (when the entire range is invalid) and some times it is recognized as a valid `ForMarkup` from our ohm parser, but fails to render it when using our Liquid gem.

## Tophat
- Ensure all ranges conform to the following standard: `(<start>..<end>)`
- Here are examples of bad ranges in loops

```
{% for num in (1...5) %}
  {{ num }}
{% endfor %}

{% for num in (1 .. 5) %}
  {{ num }}
{% endfor %}

{% for num in ( 1..5) %}
  {{ num }}
{% endfor %}

{% tablerow x in (a ..b) %}
  {{ x }}
{% endtablerow %}
```

## Before you deploy

### DIFF
--- a/.changeset/new-tomatoes-sort.md
+++ b/.changeset/new-tomatoes-sort.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-check-common': patch
+---
+
+Detect and report invalid ranges in loops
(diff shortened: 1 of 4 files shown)


graygilmore (approved):
🎩 👍🏻

PR #1041, opened 2025-09-03
Added orphaned file theme check

shoplando:

This PR adds a new `OrphanedSnippet` check to detect snippet files that are not referenced by any other files in the theme. This helps identify unused snippets that can be safely removed.

The implementation includes:
- A new `OrphanedSnippet` check with comprehensive tests
- Integration with the theme graph system to detect file references
- Moved `Reference` type definitions to `theme-check-common`, avoiding circular dependencies in order to support the check
- Updated configuration files to enable the check by default

The check reports a warning when it finds snippet files that aren't referenced anywhere else in the theme, helping developers maintain cleaner codebases by identifying unused files.

## Before you deploy

  - [x] I've made a PR to update the [shopify.dev theme check docs](https://github.com/Shopify/shopify-dev/tree/main/content/storefronts/themes/tools/theme-check/checks) if applicable (link PR here).

### DIFF
--- a/.changeset/nine-experts-peel.md
+++ b/.changeset/nine-experts-peel.md
@@ -1,0 +1,8 @@
+---
+'@shopify/theme-language-server-common': minor
+'@shopify/theme-check-common': minor
+'@shopify/theme-check-node': minor
+'@shopify/theme-graph': minor
+---
+
+Added a new OrphanedSnippet theme check to discover and warn if a snippet isn't referenced anywhere
--- a/packages/theme-check-common/src/checks/orphaned-snippet/index.spec.ts
+++ b/packages/theme-check-common/src/checks/orphaned-snippet/index.spec.ts
@@ -1,0 +1,58 @@
+import { expect, describe, it } from 'vitest';
+import { OrphanedSnippet } from './index';
+import { runLiquidCheck } from '../../test';
+
+describe('Module: OrphanedSnippet', () => {
+  describe('when the snippet is not referenced by any files', () => {
+    it('should report a warning', async () => {
+      const sourceCode = `<div>Orphaned content</div>`;
+
+      const offenses = await runLiquidCheck(
+        OrphanedSnippet,
+        sourceCode,
+        'snippets/orphaned.liquid',
+        {
+          getReferences: async (uri) => [],
+        },
+      );
+
+      expect(offenses).toHaveLength(1);
+      expect(offenses[0].message).toBe('This snippet is not referenced by any other files');
+    });
+  });
+
+  describe('when the snippet is referenced by other files', () => {
+    it('should not report a warning', async () => {
+      const sourceCode = `<div>Referenced snippet</div>`;
+
+      const offenses = await runLiquidCheck(
+        OrphanedSnippet,
+        sourceCode,
+        'snippets/product-card.liquid',
+        {
+          getReferences: async (uri) => [
+            {
+              source: { uri: 'templates/product.liquid' },
+              target: { uri: 'snippets/product-card.liquid' },
+              type: 'direct',
+            },
+          ],
+        },
+      );
+
+      expect(offenses).toHaveLength(0);
+    });
+  });
+
+  describe('when checking non-snippet files', () => {
+    it('should not report warnings for templates', async () => {
+      const sourceCode = `{% section 'header' %}`;
+
+      const offenses = await runLiquidCheck(OrphanedSnippet, sourceCode, 'templates/index.liquid', {
+        getReferences: async (uri) => [],
+      });
+
+      expect(offenses).toHaveLength(0);
+    });
+  });
+});
--- a/packages/theme-check-common/src/checks/orphaned-snippet/index.ts
+++ b/packages/theme-check-common/src/checks/orphaned-snippet/index.ts
@@ -1,0 +1,43 @@
+import { isSnippet } from '../../to-schema';
+import { LiquidCheckDefinition, Severity, SourceCodeType } from '../../types';
+
+export const OrphanedSnippet: LiquidCheckDefinition = {
+  meta: {
+    code: 'OrphanedSnippet',
+    name: 'Prevent orphaned snippets',
+    docs: {
+      description: 'This check exists to prevent orphaned snippets in themes.',
+      recommended: true,
+      url: 'https://shopify.dev/docs/storefronts/themes/tools/theme-check/checks/orphaned-snippet',
+    },
+    type: SourceCodeType.LiquidHtml,
+    severity: Severity.WARNING,
+    schema: {},
+    targets: [],
+  },
+
+  create(context) {
+    return {
+      async onCodePathEnd() {
+        const { getReferences } = context;
+
+        if (!getReferences) {
+          return;
+        }
+
+        const fileUri = context.file.uri;
+        if (isSnippet(fileUri)) {
+          const references = await getReferences(fileUri);
+
+          if (references.length === 0) {
+            context.report({
+              message: `This snippet is not referenced by any other files`,
+              startIndex: 0,
+              endIndex: 1,
+            });
+          }
+        }
+      },
+    };
+  },
+};
--- a/packages/theme-check-common/src/types.ts
+++ b/packages/theme-check-common/src/types.ts
@@ -306,6 +306,44 @@ export type Translations = {
   [k in string]: string | Translations;
 };
 
+/**
+ * A reference is a link between two modules.
+ *
+ * @example
+ *
+ * It could be a specific range that points to a whole file
+ * {
+ *   source: { uri: 'file:///templates/index.json', range: [167, 190] },
+ *   target: 'file:///sections/custom-section.liquid'
+ * }
+ *
+ * It could be a specific range that points to a specific range
+ * {
+ *   // e.g. `<parent-component></parent-component>`
+ *   source: { uri: 'file:///snippets/parent.liquid', range: [167, 190] },
+ *
+ *   // e.g. window.customElements.define('parent-component', ParentComponent);
+ *   target: { uri: 'file:///assets/theme.js', range: [0, undefined] }
+ * }
+ */
+export type Reference = {
+  source: Location;
+  target: Location;
+
+  type:
+    | 'direct' // explicit dependency, e.g. {% render 'child' %}
+    | 'indirect' // indirect dependency, e.g. "blocks": [{ "type": "@theme" }]
+    | 'preset'; // preset dependency
+};
+
+export type Range = [start: number, end: number]; // represents a range in the source code
+export type Location = {
+  /** The URI of the module */
+  uri: UriString;
+  /** Optional range inside that module */
+  range?: Range;
+};
+
 export interface Dependencies {
   /** The file system abstraction used to read files. */
   fs: AbstractFileSystem;
--- a/packages/theme-check-common/src/types.ts
+++ b/packages/theme-check-common/src/types.ts
@@ -386,6 +386,12 @@ export interface Dependencies {
    * Used in theme-checks for cross-file checks rather that going through fs.
    */
   getDocDefinition?: (relativePath: string) => Promise<DocDefinition | undefined>;
+
+  /**
+   * Get references to a file (which files reference this file)
+   * Returns an empty array if no files reference this file
+   */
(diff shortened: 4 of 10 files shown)


packages/theme-check-common/src/checks/orphaned-file/index.ts
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        const fileUri = context.file.uri;
+
+        const [references, dependencies] = await Promise.all([
+          getReferences(fileUri),
+          getDependencies(fileUri),
+        ]);
+
+        // A file is orphaned if it has NO references AND NO dependencies
```
graygilmore: Thinking out loud: is this true? If a snippet is not referenced anywhere it doesn't matter what dependencies it has, no?

packages/theme-check-common/src/checks/orphaned-file/index.ts
```diff
@@ -0,0 +1,45 @@
...
+        if (references.length === 0 && dependencies.length === 0) {
+          context.report({
+            message: `This file is not referenced by any other files and does not reference any files itself`,
+            startIndex: 0,
+            endIndex: 1,
```
graygilmore: I wonder if we could make this easier to spot. It's a little hard to find the squiggle:
image

packages/theme-check-common/src/checks/orphaned-file/index.ts
graygilmore, line 1: (🧵 threading here for easy conversation)

Two questions for clarity for myself!

Do you know where the magic is that makes this not complain about sections? e.g. I have a section with no references and no dependencies and it correctly does not tell me that it's orphaned (because sections can be inserted anywhere by the merchant)

I _think_ this is currently accidentally flagging blocks as orphaned but I think blocks can be rendered with `{% content_for 'blocks' %}` and therefore wouldn't have an explicit reference.

packages/theme-check-common/src/checks/orphaned-file/index.ts
shoplando: Yeah, Alok and I were also wondering if there is a way to just make the file getting warned, and nothing being "squiggled" in the file itself

packages/theme-check-common/src/checks/orphaned-file/index.ts
shoplando: You would be referring to dead code, which is unreachable files (we have a command for that too). This check will warn the user of a completely isolated file (no connections at all), and these files will be 100% safe to delete. Dead code (no references) may be part of a bigger cluster of files that need more investigation. The orphan theme check will guarantee a safe file deletion, but dead code might not be safe to delete.

packages/theme-check-common/src/checks/orphaned-file/index.ts
aswamy: Technically this is a file-level error. The only existing file-level error is [AppBlockMissingSchema](https://github.com/Shopify/theme-tools/blob/main/packages/theme-check-common/src/checks/app-block-missing-schema/index.ts#L37-L41) that did a similar approach (having startIndex and endIndex at 0).

I don't think underlying the whole file is right either, but maybe a showing the error only on the file (in the left panel in VSCode) is enough?

packages/theme-check-common/src/checks/orphaned-file/index.ts
shoplando: For the first question: Hmm, on my side, I still see the warning message. Should I implement that barrier for the /sections paths?
Second question: Ohh ok I'll test that out and see what I can do for that case
Thanks for the review btw! I really appreciate it!

shoplando: Also I noticed that every new file I create in the /blocks directory will automatically add two references, so no block file is ever being flagged as orphan.

.changeset/nine-experts-peel.md
```diff
@@ -0,0 +1,8 @@
+---
+'@shopify/theme-language-server-common': minor
+'@shopify/theme-check-common': minor
+'@shopify/theme-check-node': minor
+'@shopify/theme-graph': minor
+---
+
+Added orphaned-snippet theme check
```
graygilmore: We can be a little more descriptive here! This will be a public changelog so we can say something like:

```suggestion
Adds a new `OrphanedSnippet` theme check to discover and warn if a snippet isn't referenced anywhere
```

packages/theme-check-common/src/checks/orphaned-snippet/index.ts
```diff
@@ -0,0 +1,44 @@
+import { isSnippet } from '../../to-schema';
+import { LiquidCheckDefinition, Severity, SourceCodeType } from '../../types';
+
+export const OrphanedSnippet: LiquidCheckDefinition = {
+  meta: {
+    code: 'OrphanedSnippet',
+    name: 'Prevent orphaned files',
```
graygilmore:
```suggestion
    name: 'Prevent orphaned snippets',
```

packages/theme-check-common/src/checks/orphaned-snippet/index.ts
```diff
@@ -0,0 +1,44 @@
...
+
+        if (!getReferences) {
+          return;
+        }
+
+        const fileUri = context.file.uri;
+        if (isSnippet(fileUri)) {
+          const references = await getReferences(fileUri);
+
+          // A snippet is orphaned if it has NO references
```
graygilmore: We can remove this 😄. We generally want to reserve comments for things that might be misunderstood or need some extra explaining to make sense.
```suggestion
```

packages/theme-check-common/src/checks/orphaned-snippet/index.ts
```diff
@@ -0,0 +1,44 @@
...
+          if (references.length === 0) {
+            context.report({
+              message: `This snippet is not referenced by any other files`,
```
graygilmore, line 34: Do we want to go further and maybe suggest an action? I guess if you're in the middle of creating a new snippet that might be a weird warning to get 🤔 
```suggestion
              message: `This snippet is not referenced by any other files and can safely be removed`,
```

packages/theme-check-common/src/types.ts
graygilmore, line 1: (super nit) when thinking about structuring a PR sometimes we'll separate the changes into multiple commits for easier reviewing. For example, if I were writing this change I might split the this into two commits: 

1. Move theme graph types into a shared place (theme-check-common)
2. (bonus commit that probably only I'd do) Add ability for the language server to fetch references for a file
3. Add orphaned snippet theme check

packages/theme-check-common/src/checks/orphaned-snippet/index.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    const offenses = await runLiquidCheck(OrphanedSnippet, sourceCode, 'snippets/orphaned.liquid', {
+      getReferences: async (uri) => [],
+    });
+
+    expect(offenses).toHaveLength(1);
+    expect(offenses[0].message).toBe('This snippet is not referenced by any other files');
+  });
+
+  it('should NOT report snippets that have references', async () => {
```
graygilmore: Describe blocks are really handy for separating out different states!

```
describe("when the snippet is referenced in other files", () => {
  it("should not report a warning", async () => {
    // the test
  });
});
```

packages/theme-check-common/src/checks/orphaned-snippet/index.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+            type: 'direct',
+          },
+        ],
+      },
+    );
+
+    expect(offenses).toHaveLength(0);
+  });
+
+  it('should report snippets that have dependencies but no references', async () => {
```
graygilmore: Is this test meaningfully different from the test at the top?

packages/theme-check-common/src/checks/orphaned-snippet/index.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    const offenses = await runLiquidCheck(OrphanedSnippet, sourceCode, 'snippets/button.liquid', {
+      getReferences: async (uri) => [],
+    });
+
+    expect(offenses).toHaveLength(1);
+    expect(offenses[0].message).toBe('This snippet is not referenced by any other files');
+  });
+
+  it('should NOT report entry point snippets (templates)', async () => {
```
graygilmore:
```suggestion
  it('should NOT report a warning for any file that isn\'t a snippet', async () => {
```

packages/theme-check-common/src/checks/orphaned-snippet/index.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    const sourceCode = `{% section 'header' %}`;
+
+    const offenses = await runLiquidCheck(OrphanedSnippet, sourceCode, 'templates/index.liquid', {
+      getReferences: async (uri) => [],
+    });
+
+    expect(offenses).toHaveLength(0);
+  });
+
+  it('should handle missing getReferences gracefully (snippets)', async () => {
```
graygilmore: I'm not 100% sure what this is testing! Could you help me understand?

packages/theme-language-server-common/src/index.ts
```diff
@@ -32,5 +32,6 @@ export {
   FileStat,
   FileTuple,
   FileType,
+  Reference,
```
aswamy: I don't think we need to do this, right? You can remove all the changes from this file.

packages/theme-check-common/src/checks/orphaned-snippet/index.spec.ts
shoplando: True! I forgot to remove this after removing getDependencies(). Thanks!

packages/theme-check-common/src/checks/orphaned-snippet/index.ts
aswamy: I generally think it's a fair error to see while you're tying in the file. We do the same for Missing schema errors. As long as the error squiggly lines don't make it horrible to code (like underscoring the whole document), this is fair.

packages/theme-language-server-common/src/index.ts
shoplando: Since the Reference type got moved from theme-graph to theme-check-common, I had to make this change, I'll separate the changes into two commits as how Gray suggested.

aswamy: oh i guess we used to also export `Reference` from this file. I just meant idk who actually consumes `Reference` _FROM_ theme-language-server (they'd probably just consume it from theme-check-common or theme-graph instead). But i guess for consistency, we can keep it like this.

packages/theme-check-common/src/checks/orphaned-snippet/index.spec.ts
shoplando: We're trying to see if the function itself is available, but I guess this test won't do much since getReferences() should always be available.

packages/theme-check-common/src/checks/orphaned-file/index.ts
shoplando: We can maybe create a new issue on github for this

packages/theme-graph/src/types.ts
```diff
@@ -3,6 +3,9 @@ import {
   LiquidSourceCode,
   Dependencies as ThemeCheckDependencies,
   UriString,
+  Reference,
+  Location,
+  Range,
```
charlespwd, line 8: I think we could have re-exported those
```ts
export { Reference, Location, Range }
```

This would have made it so you didn't need to change so many import statements across the codebase

packages/theme-language-server-common/src/types.ts
```diff
@@ -8,7 +8,7 @@ import * as rpc from 'vscode-jsonrpc';
 
 import { WithOptional } from './utils';
 import { Range } from 'vscode-json-languageservice';
-import { Reference } from '@shopify/theme-graph';
+import { Reference } from '@shopify/theme-check-common';
```
charlespwd: re: https://github.com/Shopify/theme-tools/pull/1041/files#r2325810511

e.g. here

aswamy (approved):
Amazing work! 🎉 Very thorough in your investigation on how theme-graph and dependency injection works!

PR #1044, opened 2025-09-05
added invalid filter name check

EvilGenius13:

Adding a check for invalid filter names (not the same as an unknown filter)

Here are a few examples of liquid parsing of filters:

Properly named filter (filter activates):
```
irb(main):002> Liquid::Template.parse("{{ 'Hello' | append: 'world' }}").render
=> "Helloworld"
```
Filter with trailing numbers attached (filter doesn't activate)
```
irb(main):003> Liquid::Template.parse("{{ 'Hello' | append123: 'world' }}").render
=> "Hello"
```
Filter with trailing letters (filter doesn't activate):
```
irb(main):004> Liquid::Template.parse("{{ 'Hello' | unknownfilter: 'world' }}").render
=> "Hello"
```
Filter with trailing special characters (filter activates)
```
irb(main):007> Liquid::Template.parse("{{ 'Hello' | append!@($: 'world' }}").render
=> "Helloworld"
```

This gives us two types of scenarios, either we have an improperly named filter that still works, or we have an improperly named filter that never worked.

- For filters that work, we fix them by removing the trailing junk.
- For filters that don't work, we remove the entire filter.

I left some comments in the code 

## Before you deploy


packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidFilterName.ts
```diff
@@ -0,0 +1,60 @@
...
+    return;
+  }
+
+  const filterName = node.name;
+  const knownFilters = await context.themeDocset.filters();
+
+  let matchingFilter = null;
+  let longestMatchLength = 0;
+
+  for (const filter of knownFilters) {
```
EvilGenius13: This finds the longest filter that matches the start of the filter name

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidFilterName.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (filterName.startsWith(filter.name) && filter.name.length > longestMatchLength) {
+      matchingFilter = filter;
+      longestMatchLength = filter.name.length;
+    }
+  }
+
+  if (!matchingFilter) {
```
EvilGenius13: Any filter that isn't a "known filter" will be handled by the `UnknownFilter` check.
EX: `{{ 'hello' | notarealcheck }} will land here.

karreiro (commented):
Thank you for this PR, @EvilGenius13!

We can actually strictly parse this file as unknown filters can be parsed strictly:
```
{{ 'HELLO' | append123: 'world' }}
{{ 'HELLO' | appendme: 'world' }}
{{ 'hello' | append: 'world' }}
{{ 'hello' | completely_unknown_filter: 'world' }}
{{ 'test' | append123: 'world' | color_to_rgb456 }}
{{ 'test' | append123: 'world' | color_to_rgb }}
{{ 'test' | append: 'world' | color_to_rgb123 }}
{{ 'test' | append: 'world' | color_to_rgb123 | item_count_for_variant | link_to_type456 }}
{{ 'test' | append123 | color_to_rgb456 | item_count_for_variantabc }}
{{ 'test' | append123: 'mixed' | color_to_rgbABC | link_to_type999xyz }}
{{ 'test' | appendXYZ | append: 'good' | color_to_rgb123 | color_to_rgb }}
{{ 'test' | append999 | append: 'step2' | color_to_rgb | item_count_for_variant }}
{{ 'test' | append: 'step1' | color_to_rgb666 | item_count_for_variant | link_to_type }}
{{ 'HELLO' | append123: 'world' }}
{{ 'HELLO' | appendme }}
{{ 'test' | append123: 'world' | color_to_rgb }}
{{ 'test' | append: 'world' | color_to_rgb123 }}
{{ 'test' | append123: 'world' | color_to_rgb456 }}
{{ 'test' | append: 'good' | color_to_rgb123 | item_count_for_variant }}
```

I think we could repurpose this PR to cover scenarios like this (involving special characters):
```
input:
{{ 'HELLO' | append!: 'world' }}

expected:
{{ 'HELLO' | append: 'world' }}
```

What do you think?

Thanks again for this PR!

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidEchoValue.ts
```diff
@@ -59,6 +59,10 @@ export function detectInvalidEchoValue(
 
   const { startIndex, endIndex } = removalIndices(node.source, node.position.start);
 
+  if (endIndex <= startIndex) {
```
EvilGenius13, line 62: We were flagging a separate Syntax issue on invalid filters coming from the `InvalidEchoValue` check. If there is nothing but spaces before the pipe such as `{{ 'hello' |`, we won't report it.

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidFilterName.ts
```diff
@@ -0,0 +1,55 @@
...
+    const markupStartInSource = node.position.start + 3;
+    const trailingStartInSource = markupStartInSource + filterEndIndex;
+    const trailingEndInSource = trailingStartInSource + trailingChars.length;
+
+    problems.push({
+      message: `${INVALID_SYNTAX_MESSAGE} Filter '${filterName}' has trailing characters '${trailingChars}' that should be removed.`,
+      startIndex: trailingStartInSource,
+      endIndex: trailingEndInSource,
+      fix: (corrector) => {
+        corrector.replace(trailingStartInSource, trailingEndInSource, '');
```
graygilmore, line 76: Might need a tiny tweak here!
fix-all

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidEchoValue.ts
aswamy: Good catch! I saw that error too but glad you caught it in your PR.

Can you add a test for reporting an InvalidEchoValue error when no filters are provided?

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidFilterName.ts
```diff
@@ -0,0 +1,55 @@
...
+    const filterName = match[1];
+
+    if (!knownFilters.some((filter) => filter.name === filterName)) {
+      continue;
+    }
+
+    const filterEndIndex = match.index! + match[0].length;
+    const afterFilter = markup.slice(filterEndIndex);
+
+    const trailingChars = afterFilter.match(/^(?!\s*(?::|$|\|\s*[a-zA-Z]))([^:|]+)/)?.[1];
```
aswamy: 😵💫 This regex kills me

packages/theme-check-common/src/checks/liquid-html-syntax-error/index.ts
```diff
@@ -74,13 +75,15 @@ export const LiquidHTMLSyntaxError: LiquidCheckDefinition = {
           context.report(problem);
         },
         async LiquidVariableOutput(node) {
-          const problem = detectInvalidEchoValue(node);
-
-          if (!problem) {
-            return;
+          const filterProblems = await detectInvalidFilterName(node, context);
```
aswamy: Are we only including this for `{{ }}`? Should we also have this for `assign` and `echo` tags?

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidFilterName.ts
```diff
@@ -0,0 +1,55 @@
...
+    if (!trailingChars) {
+      continue;
+    }
+
+    const markupStartInSource = node.position.start + 3;
```
aswamy: What's with the +3? Is it to offset the `{{ `? If that's it, this wouldn't be great because it could be `{{-` or `{{` or `{{       `

Consider using `node.source.indexOf(node.markup, node.position.start)` instead:
- It looks through the source for markup, but only starts AFTER `node.position.start`

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidFilterName.ts
```diff
@@ -0,0 +1,55 @@
...
+): Promise<Problem<SourceCodeType.LiquidHtml>[]> {
+  if (
+    node.type !== NodeTypes.LiquidVariableOutput ||
+    !context.themeDocset ||
+    typeof node.markup !== 'string'
+  ) {
+    return [];
+  }
+
+  const knownFilters = await context.themeDocset.filters();
```
aswamy: I'd prefer to keep the whole method sync, and move this outside like i did here:
https://github.com/Shopify/theme-tools/pull/1045/files#diff-a6b6a40113e389c51506ec71986af5141e8f97a315ee91b416128d26ad4ba5ccR44

That way, we only call these await once per theme check instead of once per node traversal

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidFilterName.ts
aswamy: We've talked about formatting a lot around our theme checks. Only do it if it doesn't make the code uglier (from a readability) perspective. We should rely on the dev to format their editor.

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidEchoValue.ts
aswamy: Can you also patch it for Assign tag as well 😅 
image

PR #1045, opened 2025-09-05
temporary

aswamy:
## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidLoopArguments.ts
+++ b/packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidLoopArguments.ts
@@ -1,0 +1,70 @@
+import { LiquidTag, LiquidTagFor, LiquidTagTablerow, NodeTypes } from '@shopify/liquid-html-parser';
+import { Problem, SourceCodeType, TagEntry } from '../../..';
+import { doesFragmentContainUnsupportedParentheses, getFragmentsInMarkup, fragmentKeyValuePair } from './utils';
+import { isLoopLiquidTag } from '../../utils';
+
+export function detectInvalidLoopArguments(
+  node: LiquidTag,
+  tags: TagEntry[] = [],
+): Problem<SourceCodeType.LiquidHtml> | undefined {
+  if (node.type === NodeTypes.LiquidTag && !isLoopLiquidTag(node)) {
+    return;
+  }
+
+  const markup = (node as LiquidTagFor | LiquidTagTablerow).markup;
+
+  if (!markup) {
+    return;
+  }
+
+  if (typeof markup !== 'string') return;
+
+  const fragments = getFragmentsInMarkup(markup);
+
+  // The first three fragments are: `<variable> in <array>`
+  // If they haven't started typing the tag, we don't want to report an error.
+  if (fragments.length <= 3) {
+    return;
+  }
+
+  const markupIndex = node.source.indexOf(markup, node.position.start);
+  let startIndex = markupIndex + fragments.at(3)!.index!;
+  let endIndex = markupIndex + fragments.at(-1)!.index! + fragments.at(-1)!.value.length;
+
+  const filteredFragments: string[] = [];
+
+  for (let i = 3; i < fragments.length; i++) {
+    const fragment = fragments[i];
+
+    if (doesFragmentContainUnsupportedParentheses(fragment.value)) {
+      continue;
+    }
+
+    const keyValuePair = fragmentKeyValuePair(fragment.value);
+
+    if (keyValuePair && !isSupportedTagArgument(tags, node.name, keyValuePair.key, false)) {
+      continue;
+    }
+
+    if (!isSupportedTagArgument(tags, node.name, fragment.value, true)) {
+      continue;
+    }
+
+    // TODO: Consider reordering the fragments so that positional arguments are before named arguments
+    // But this might change the functionality of the loop
+    filteredFragments.push(fragment.value);
+  }
+
+  return {
+    message: 'Loop arguments must be `reversed` or named arguments.',
+    startIndex,
+    endIndex,
+    fix: (corrector) => {
+      corrector.replace(startIndex, endIndex, filteredFragments.join(' '));
+    },
+  };
+}
+
+function isSupportedTagArgument(tags: TagEntry[], tagName: string, key: string, positional: boolean) {
+  return tags.find((tag) => tag.name === tagName)?.parameters?.some((parameter) => parameter.name === key && parameter.positional === positional) || false;
+}
--- a/packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/utils.ts
+++ b/packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/utils.ts
@@ -5,7 +5,7 @@ export const INVALID_SYNTAX_MESSAGE = 'Syntax is not supported';
 
 export function ensureValidAst(source: string) {
   try {
-    const ast = toLiquidAST(source, { allowUnclosedDocumentNode: false, mode: 'strict' });
+    const ast = toLiquidAST(source, { allowUnclosedDocumentNode: false, mode: 'tolerant' });
     return !isError(ast);
   } catch (_error) {
     return false;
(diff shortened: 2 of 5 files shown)


packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/utils.ts
```diff
@@ -5,16 +5,73 @@ export const INVALID_SYNTAX_MESSAGE = 'Syntax is not supported';
...
   }));
 }
+
+const DOUBLE_QUOTED_STRING = `"[^"]*"`;
+const SINGLE_QUOTED_STRING = `'[^']*'`;
+// Lax parser does NOT complain about leading/trailing spaces inside ranges (e.g. `(1 .. 10 )`) and
+// within the parenthesis (e.g. `( 1 .. 10 )`), but fails to render the liquid when using the gem.
+// Strict parser does NOT complain, but still renders it.
+// To avoid any issues, we will remove extra spaces.
+const RANGE_MARKUP_REGEX = `\\(\\s*(\\w+)\\s*(\\.{2,})\\s*(\\w+)\\s*\\)`;
```
karreiro: Could we update this regex to prevent this scenario `{% for i in (1..10.1) %}{{ i }}{% endfor %}`?

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidLoopArguments.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+    for (const sourceCode of testCases) {
+      const offenses = await testCheck(sourceCode);
+      expect(offenses).to.have.length(0);
+    }
+  });
+
+  it('should report when invalid args are found', async () => {
+    const testCases = [
+      [
+        `{% for i in array limit1: 10 %}{% endfor %}`,
```
karreiro, line 23: Here we're showing this message:

> Loop arguments invalid. Positional arguments must be provided before named arguments. Invalid arguments: limit1: 10

However, that doesn't seem so accurate, because we actually don't have positional arguments involved, we have an unknown of invalid named argument. Am I missing something?

karreiro: We have a similar error message for `{% for i in array reversed invalid limit: 10 %}`. However, the problem is that **invalid** is not a supported argument, so maybe we want to adopt a more generic message to prevent ambiguity.

karreiro (commented):
Thank you, @aswamy! Excellent stuff 🔥 it's working very well!

I just dropped a few minor comments. Please, let me know what you think :)

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/utils.ts
aswamy: 1000% updated and added tests for bad ranges + ensuring we dont have false positives like nested variables (e.g. `(1..obj.child.inner)`)

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidLoopArguments.spec.ts
aswamy: Made it more generic! Matched how i did it for invalid ranges! Thanks for the advice.

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/utils.ts
```diff
@@ -5,16 +5,74 @@ export const INVALID_SYNTAX_MESSAGE = 'Syntax is not supported';
 
 export function ensureValidAst(source: string) {
   try {
-    const ast = toLiquidAST(source, { allowUnclosedDocumentNode: false, mode: 'strict' });
+    const ast = toLiquidAST(source, { allowUnclosedDocumentNode: false, mode: 'tolerant' });
     return !isError(ast);
   } catch (_error) {
     return false;
```
charlespwd, line 11: I have slight perf concerns about doing that in every check since that could be a multiplicative number of N files that this happen (given parsing is still the slowest part of our pipeline). 

Are we sure we do this only if strictly necessary?

charlespwd: Since corrector.replace is lazy, maybe we could put that in the`fix` function instead? 🤔

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidLoopArguments.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  let markup: ForMarkup | string = (node as LiquidTagFor | LiquidTagTablerow).markup;
+
+  if (!markup) {
+    return;
+  }
+
+  if (typeof markup !== 'string') {
+    markup = markup.source.slice(markup.position.start, markup.position.end);
```
charlespwd, line 31: If markup is not a string, doesn't that mean that our LiquidHTML parser has correctly parsed the node? Why double check?

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/utils.ts
aswamy: I talked to @EvilGenius13 about this too - Originally when we found a bad node (e.g. `{{ one two }}`) in a theme file, we would ONLY check the node's source to see if our suggested fix is valid Liquid (e.g. `toLiquidAST("{{ one }}"...)`). The premise was:
- When we offer suggestions, make sure it is valid Liquid AST

It wouldn't be _TOO_ slow since we aren't checking the whole file. But the more i use it i released we shouldn't be doing this because if a single node has MULTIPLE error points (e.g. we have two different checks in for-loops - one to check if the range is valid, and one to check if there are bad arguments passed in), it is possible that our suggested fixes offered from the fixers aren't individually correct - but are correct AFTER both are applied.

e.g.
If the user wrote `{% for i in (1....10) reversed invalid %}` there are two errors (the range is invalid, and they passed an unknown arg)

InvalidLoopArguments would offer `{% for i in (1....10) reversed %}` as a solution
InvalidLoopRange would offer `{% for i in (1..10) reversed invalid %}` as a solution

Both of those offered solutions aren't valid Liquid AST (in strict mode). We would need to check AST AFTER both are applied.

Potential solution:
A) Check the whole file's AST after all the patches are done
B) Don't bother checking AST and be confident what we offer is correct

Im leaning towards B because if A fails a check because we don't correct _every_ error, we would have to revert every fix.

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidLoopArguments.ts
aswamy: Valid question!

TLDR: We have a pretty inconsistent way we parse positional args and named args when passed into loops
- so even though it is "valid" markup, it won't pass Liquid gem's strict parser
- when you pass invalid positional args, the markup will be a string
- when you pass invalid named args, the markup will be `ForMarkup`


E.g.

`{% for i in (1..5) reversed arg: 1 %}{{i}}{% endfor %}`
- Valid Markup in theme-tools (it will be a `ForMarkup` not `string`)
- Valid when parsed via Liquid Lax Parser via gem (`arg: 1` will be ignored)
- Invalid when parsed via Liquid Strict Parser via gem

image

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/utils.ts
EvilGenius13: Personal fan of B

charlespwd: Yeah. I don't think it's a huge deal if we have to rerun more actions either. B feels better and eliminates the perf cost risk

aswamy: Sounds good, I will make a follow-up PR to remove this because I think @EvilGenius13 and I have a few instance of this method, and it's easier to cleanup and tophat all of them at once.

PR #1050, opened 2025-09-12
Remove 'CaptureOnContentForBlock' check

karreiro:

This PR removes the `CaptureOnContentForBlock` check from the codebase. As theme development standards have evolved, this check is now fully deactivated on all major themes.

With the growing adoption of blocks, using this approach has become a standard way to encourage reusability and prevent duplicate markup.

- Remove this page https://shopify.dev/docs/storefronts/themes/tools/theme-check/checks/capture-on-content-for-block from the documentation

### DIFF
--- a/.changeset/curvy-teachers-enjoy.md
+++ b/.changeset/curvy-teachers-enjoy.md
@@ -1,6 +1,6 @@
 ---
-'@shopify/theme-check-common': patch
-'@shopify/theme-check-node': patch
+'@shopify/theme-check-common': minor
+'@shopify/theme-check-node': minor
 ---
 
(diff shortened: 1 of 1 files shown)


.changeset/curvy-teachers-enjoy.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/theme-check-common': patch
+'@shopify/theme-check-node': patch
```
karreiro: Updating this to minor as it includes changes in the configuration of a check:

```suggestion
'@shopify/theme-check-common': minor
'@shopify/theme-check-node': minor
```

Issue #968, opened 2025-06-11
Rework `DocumentLinks` to use `@shopify/theme-graph`


aswamy:
**Is your feature request related to a problem? Please describe.**

[Shopify/theme-graph](https://github.com/Shopify/theme-tools/pull/956) allowed each file to be able to tell if it has a reference or dependency to another file.

Instead of explicitly looking through Liquid Tags and checking to see if it's a variable or liquid string, we can just underline all references in liquid files using theme-graph's reference.

File in question: https://github.com/Shopify/theme-tools/blob/main/packages/theme-language-server-common/src/documentLinks/DocumentLinksProvider.ts#L37-L50

- [ ] Refactor provider to include theme-graph
- [ ] Introduce document linking from Liquid schemas to blocks/sections it's referencing
- [ ] Introduce document linking from web components to the javascript file it's referencing

**Describe the solution you'd like**
- Ideally we can keep this file extremely minimal and only check the theme-graph for references and where it's located in the file
- Introduces schema level document linking (which currently doesn't exist)

PR #1051, opened 2025-09-12
Added document links inside schema tags

shoplando:

Added document links support for schema tags in Liquid files. This enhancement allows navigation from schema tags within Liquid files to referenced blocks.

The implementation includes:
- A new `JsonDocumentLinksProvider` class for handling JSON document links
- Integration with the existing `DocumentLinksProvider` to support schema tags
- Special handling for block types (ignoring special types like `@app` and `@theme`)

- Implement theme graph in DocumentLinks as described [here](https://github.com/Shopify/theme-tools/issues/968)

## What did you learn?

- Implementing a visitor in an AST
- It was an interesting challenge when it came to implementing a visitor inside the current visitor. 

### DIFF
--- a/packages/theme-language-server-common/src/json/JSONLanguageService.ts
+++ b/packages/theme-language-server-common/src/json/JSONLanguageService.ts
@@ -8,12 +8,16 @@ import {
   SourceCodeType,
   findCurrentNode,
   isValid,
+  toJSONAST,
+  visit,
 } from '@shopify/theme-check-common';
 import { JSONDocument, LanguageService, getLanguageService } from 'vscode-json-languageservice';
 import {
   CompletionItem,
   CompletionList,
   CompletionParams,
+  DocumentLink,
+  DocumentLinkParams,
   Hover,
   HoverParams,
   ClientCapabilities as LSPClientCapabilities,
@@ -22,6 +26,9 @@ import { TextDocument } from 'vscode-languageserver-textdocument';
 import { DocumentManager } from '../documents';
 import { GetTranslationsForURI } from '../translations';
 import { JSONContributions, GetThemeBlockNames, GetThemeBlockSchema } from './JSONContributions';
+import { createJSONDocumentLinksVisitor } from './documentLinks/DocumentLinksProvider';
+import { URI } from 'vscode-uri';
+import { FindThemeRootURI } from '../internal-types';
 
 export class JSONLanguageService {
   // We index by Mode here because I don't want to reconfigure the service depending on the URI.
@@ -45,6 +52,7 @@ export class JSONLanguageService {
     private getModeForURI: (uri: string) => Promise<Mode>,
     private getThemeBlockNames: GetThemeBlockNames,
     private getThemeBlockSchema: GetThemeBlockSchema,
+    private findThemeRootURI: FindThemeRootURI,
   ) {
     this.services = Object.fromEntries(Modes.map((mode) => [mode, null])) as typeof this.services;
     this.schemas = {};
@@ -125,6 +133,39 @@ export class JSONLanguageService {
     if (!documents) return null;
     const [jsonTextDocument, jsonDocument] = documents;
     return service.doHover(jsonTextDocument, params.position, jsonDocument);
+  }
+
+  async documentLinks(params: DocumentLinkParams): Promise<DocumentLink[]> {
+    await this.initialized;
+    const rootUri = await this.findThemeRootURI(params.textDocument.uri);
+    if (!rootUri) return [];
+
+    const document = this.documentManager.get(params.textDocument.uri);
+    if (!document) return [];
+
+    switch (document.type) {
+      case SourceCodeType.LiquidHtml: {
+        if (document.ast instanceof Error) return [];
+        const textDocument = document.textDocument;
+        const links: DocumentLink[] = [];
+
+        const schema = await document.getSchema();
+        if (schema && !(schema.ast instanceof Error)) {
+          const visitor = createJSONDocumentLinksVisitor(
+            textDocument,
+            URI.parse(rootUri),
+            schema.offset,
+          );
+          const schemaLinks = visit(schema.ast, visitor);
+          links.push(...schemaLinks);
+        }
+
+        return links;
+      }
+
+      default:
+        return [];
+    }
   }
 
   public isValidSchema = async (uri: string, jsonString: string) => {
--- a/packages/theme-language-server-common/src/json/documentLinks/DocumentLinksProvider.spec.ts
+++ b/packages/theme-language-server-common/src/json/documentLinks/DocumentLinksProvider.spec.ts
@@ -1,0 +1,141 @@
+import { beforeEach, describe, expect, it } from 'vitest';
+import { DocumentManager } from '../../documents';
+import { createJSONDocumentLinksVisitor } from './DocumentLinksProvider';
+import { toJSONAST, visit } from '@shopify/theme-check-common';
+import { URI } from 'vscode-uri';
+
+describe('JSON Document Links in Schema Tags', () => {
+  let documentManager: DocumentManager;
+  let rootUri: string;
+  let uriString: string;
+
+  beforeEach(() => {
+    documentManager = new DocumentManager();
+  });
+
+  it('should return document links for block types in schema tags', async () => {
+    uriString = 'file:///path/to/sections/header.liquid';
+    rootUri = 'file:///path/to/project';
+
+    const schemaContent = `
+      {
+        "name": "Header",
+        "blocks": [
+          { "type": "logo" },
+          { "type": "navigation" }
+        ]
+      }
+    `;
+
+    const liquidContent = `
+      <div>Header content</div>
+      
+      {% schema %}
+      ${schemaContent}
+      {% endschema %}
+    `;
+
+    documentManager.open(uriString, liquidContent, 1);
+
+    const textDocument = documentManager.get(uriString)!.textDocument;
+    const schemaOffset = liquidContent.indexOf(schemaContent);
+
+    const visitor = createJSONDocumentLinksVisitor(textDocument, URI.parse(rootUri), schemaOffset);
+    const ast = toJSONAST(schemaContent);
+
+    if (ast instanceof Error) throw ast;
+    const result = visit(ast, visitor);
+
+    expect(result).toHaveLength(2);
+    expect(result[0].target).toBe('file:///path/to/project/blocks/logo.liquid');
+    expect(result[1].target).toBe('file:///path/to/project/blocks/navigation.liquid');
+  });
+
+  it('should handle section types in schema tags', async () => {
+    uriString = 'file:///path/to/sections/main-collection.liquid';
+    rootUri = 'file:///path/to/project';
+
+    const schemaContent = `
+      {
+        "name": "Collection",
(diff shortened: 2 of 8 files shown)


packages/theme-language-server-common/src/documentLinks/DocumentLinksProvider.ts
```diff
@@ -79,6 +80,22 @@ function documentLinksVisitor(
...
+        }
+
+        const visitor = createJSONDocumentLinksVisitor(
+          textDocument,
+          root,
+          node.body.position.start,
+        );
+        return visit(jsonAST, visitor);
+      }
+    },
```
aswamy: Why are we approaching the JSON parsing through the schema tag?

I think all the JSON supported are isolated in the `json` directory and Liquid providers don't directly call them. Consider how we do it for code completion - in the `startServer` file, we have a `onCompletion` listener and we feed in all the completion providers (liquid completion, json completion, and css completion) into it.

We should do the same for document links. Go to the `onDocumentLinks` method in that file, and add our json service there.

For reference: https://github.com/Shopify/theme-tools/blob/main/packages/theme-language-server-common/src/server/startServer.ts#L521-L528

packages/theme-language-server-common/src/json/documentLinks/DocumentLinksProvider.ts
```diff
@@ -0,0 +1,107 @@
...
+        );
+      }
+    },
+  };
+  return visitor;
+}
+
+function range(textDocument: TextDocument, node: JSONNode, offset: number): Range {
+  const start = textDocument.positionAt(offset + node.loc.start.offset + 1);
+  const end = textDocument.positionAt(offset + node.loc.end.offset - 1);
```
aswamy, line 44: You should put a comments on why we're doing +1/-1 here. What's it do?

packages/theme-language-server-common/src/json/documentLinks/DocumentLinksProvider.ts
```diff
@@ -0,0 +1,107 @@
...
+        return true;
+      }
+      if (ancestor.key.value === 'sections') {
+        // If we find sections first, we're not in a blocks context
+        return false;
+      }
+    }
+  }
+  return false;
+}
```
aswamy: Couldn't you just do `getOriginDirectory(ancestors: JSONNode[]): sections | blocks | undefined` instead of copying 99% same code 😅

aswamy (commented):
Couple concerns but looks amazing! Can't believe you made it work for JSON sections, and templates too! I thought you'd only do it for schemas! 😁

packages/theme-language-server-common/src/json/documentLinks/DocumentLinksProvider.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    const rootUri = await this.findThemeRootURI(uriString);
+    if (!rootUri) {
+      return [];
+    }
+
+    const visitor = createJSONDocumentLinksVisitor(sourceCode.textDocument, URI.parse(rootUri));
+    return visit(sourceCode.ast, visitor);
+  }
+}
```
aswamy: Is this even used? Or is this for the next PR? If that's true, you should put it in that PR.

packages/theme-language-server-common/src/json/documentLinks/DocumentLinksProvider.ts
```diff
@@ -0,0 +1,92 @@
...
+
+function range(textDocument: TextDocument, node: JSONNode, offset: number): Range {
+  // +1 and -1 to exclude the quotes
+  const start = textDocument.positionAt(offset + node.loc.start.offset + 1);
+  const end = textDocument.positionAt(offset + node.loc.end.offset - 1);
+  return Range.create(start, end);
+}
+
+// Find the nearest parent property that is either "sections" or "blocks"
+function isInContext(ancestors: JSONNode[], context: 'sections' | 'blocks'): boolean {
```
aswamy: The term "context" is quite overloaded in theme-tools - avoid using that term if you don't provide more in the name. Also change the signature of the method:

```suggestion
function jsonPropertyOriginDirectory(ancestors: JSONNode[]): 'sections' | 'blocks' | undefined {
```

packages/theme-language-server-common/src/json/documentLinks/DocumentLinksProvider.ts
```diff
@@ -0,0 +1,92 @@
...
+        isInContext(ancestors, 'blocks') &&
+        node.value.type === 'Literal' &&
+        typeof node.value.value === 'string' &&
+        !['@app', '@theme'].includes(node.value.value)
+      ) {
+        return DocumentLink.create(
+          range(textDocument, node.value, offset),
+          Utils.resolvePath(root, 'blocks', node.value.value + '.liquid').toString(),
+        );
+      }
```
aswamy: Lot of code repetition here. Also we could probably clean up a lot of logic with early returns. Try the following:

```suggestion
      const origin = jsonPropertyOriginDirectory(ancestors);

      if (!origin) return;

      if (!(node.key.value === 'type' &&
        node.value.type === 'Literal' &&
        typeof node.value.value === 'string')
      ) {
        return;
      }

      if (origin === 'blocks' && ['@app', '@theme'].includes(node.value.value)) {
        return;
      }

      return DocumentLink.create(
        range(textDocument, node.value, offset),
        Utils.resolvePath(root, origin, node.value.value + '.liquid').toString(),
      );
```

packages/theme-language-server-common/src/json/documentLinks/index.ts
```diff
@@ -0,0 +1 @@
+export { JsonDocumentLinksProvider } from './DocumentLinksProvider';
```
aswamy: We don't need this extra export, right?

packages/theme-language-server-common/src/server/startServer.ts
```diff
@@ -507,7 +508,19 @@ export function startServer(
...
+
+    if (document.type === SourceCodeType.LiquidHtml) {
+      const [liquidLinks, jsonLinks] = await Promise.all([
+        documentLinksProvider.documentLinks(params.textDocument.uri),
+        jsonLanguageService.documentLinks(params),
+      ]);
+      return [...liquidLinks, ...jsonLinks];
+    }
+
+    return [];
```
aswamy: Try doing quick return to avoid multiple returns. Also do we need that LiquidHtml check?

```suggestion
    if (!document || document.type !== SourceCodeType.LiquidHtml) return [];

    const [liquidLinks, jsonLinks] = await Promise.all([
      documentLinksProvider.documentLinks(params.textDocument.uri),
      jsonLanguageService.documentLinks(params),
    ]);
    return [...liquidLinks, ...jsonLinks];
```

packages/theme-language-server-common/src/json/JSONLanguageService.ts
```diff
@@ -127,6 +135,45 @@ export class JSONLanguageService {
...
+            }
+          },
+        });
+
+        return links;
+      }
+
+      default:
+        return [];
+    }
```
aswamy: I noticed there is a `getDocuments` in the same file. Would it help to use that here? Or will using that method create more confusion in the code?

packages/theme-language-server-common/src/json/documentLinks/index.ts
shoplando: I was following the documentLinks pattern in the original implementation, but yeah I just realized that I'm exporting this function twice with this barrel file 😓

packages/theme-language-server-common/src/server/startServer.ts
shoplando: Ooh good catch, documentLinksProvider already checks internally if the document is LiquidHtml, so yeah the check is not needed, thanks!

packages/theme-language-server-common/src/json/JSONLanguageService.ts
shoplando: I think it would be better to use documentLinks mainly because the return value of getDocument will not work with the visitor pattern used for document link, which expects `JSONNode` (the exact AST structure) instead of getDocument's `TextDocument | JSONDocument` types

packages/theme-language-server-common/src/json/documentLinks/DocumentLinksProvider.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+      {% endschema %}
+    `;
+
+    documentManager.open(uriString, liquidContent, 1);
+
+    const textDocument = documentManager.get(uriString)!.textDocument;
+    const schemaStart = liquidContent.indexOf('{% schema %}') + '{% schema %}'.length;
+    const schemaEnd = liquidContent.indexOf('{% endschema %}');
+    const schemaContent = liquidContent.substring(schemaStart, schemaEnd).trim();
+    const schemaOffset = liquidContent.indexOf(schemaContent);
```
aswamy: We can reduce some of the fluff in this test case so it's easier to tell what we're actually testing. We shouldn't be doing all these schema offset calculations.

```suggestion
    const schemaContent = `
      {
        "name": "Header",
        "blocks": [
          { "type": "logo" },
          { "type": "navigation" }
        ]
      }
    `;

    const liquidContent = `
      <div>Header content</div>
      
      {% schema %}
      ${schemaContent}
      {% endschema %}
    `;

    documentManager.open(uriString, liquidContent, 1);

    const textDocument = documentManager.get(uriString)!.textDocument;
    const schemaOffset = liquidContent.indexOf(schemaContent);

    const visitor = createJSONDocumentLinksVisitor(textDocument, URI.parse(rootUri), schemaOffset);
```

packages/theme-language-server-common/src/json/documentLinks/DocumentLinksProvider.spec.ts
```diff
@@ -0,0 +1,138 @@
...
```
aswamy: Same comment as the other test

packages/theme-language-server-common/src/json/documentLinks/DocumentLinksProvider.spec.ts
```diff
@@ -0,0 +1,138 @@
...
```
aswamy: Same comment as the other test

packages/theme-language-server-common/src/json/JSONLanguageService.ts
```diff
@@ -127,6 +135,45 @@ export class JSONLanguageService {
...
+    switch (document.type) {
+      case SourceCodeType.LiquidHtml: {
+        if (document.ast instanceof Error) return [];
+        const textDocument = document.textDocument;
+        const links: DocumentLink[] = [];
+
+        visit(document.ast, {
+          LiquidRawTag(node: LiquidRawTag) {
+            if (node.name === 'schema') {
+              const jsonAST = toJSONAST(node.body.value);
```
charlespwd: I feel like document already has that cached. 🤔

charlespwd: Something like `document.getSchema()`

shoplando:
### Merge activity

* **Sep 19, 7:26 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/theme-tools/1051).

PR #1053, opened 2025-09-15
Added document linking inside json files

shoplando:

Added document linking support inside JSON files. This allow faster navigation and better dev experience.

The implementation:
- Extends the DocumentLinksProvider to handle JSON files in addition to Liquid HTML files
- Creates links for section type references in templates
- Creates links for block type references in section schemas
- Properly ignores special block types like "@app" and "@theme"
- Handles nested structures correctly

## What did you learn?

Working with document links across different file types requires careful handling of URI paths and proper type checking to ensure the correct linking behavior.

## Before you deploy

### DIFF
--- a/.changeset/sixty-queens-float.md
+++ b/.changeset/sixty-queens-float.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-language-server-common': minor
+---
+
+Added document linking inside json files
--- a/packages/theme-language-server-common/src/json/JSONLanguageService.ts
+++ b/packages/theme-language-server-common/src/json/JSONLanguageService.ts
@@ -144,6 +144,12 @@ export class JSONLanguageService {
     if (!document) return [];
 
     switch (document.type) {
+      case SourceCodeType.JSON: {
+        if (document.ast instanceof Error) return [];
+        const visitor = createJSONDocumentLinksVisitor(document.textDocument, URI.parse(rootUri));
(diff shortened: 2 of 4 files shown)


packages/theme-language-server-common/src/json/documentLinks/DocumentLinksProvider.ts
```diff
@@ -2,7 +2,28 @@ import { JSONNode, SourceCodeType } from '@shopify/theme-check-common';
...
+      return [];
+    }
+    const rootUri = await this.findThemeRootURI(uriString);
+    if (!rootUri) {
+      return [];
+    }
+    const visitor = createJSONDocumentLinksVisitor(sourceCode.textDocument, URI.parse(rootUri));
+    return visit(sourceCode.ast, visitor);
+  }
+}
```
aswamy: This is defined and tested, but where is this even called?

shoplando: Oh yeah you're right, this was added when schema tag was handled inside the main `DocumentLinksProvider` with a nested visitor. I'll remove this.

shoplando:
### Merge activity

* **Sep 19, 7:39 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Shopify/theme-tools/1053).

PR #1054, opened 2025-09-15
Invalid pipe check

EvilGenius13:

Leftover scenario for catching parsing issues which are double pipes (`||`) or a trailing pipe (`|').

I was originally trying to add this into the `invalidFilterName` check but it wasn't playing nicely so now it's got it's own check.

Examples it will catch
`{{ 'hello' | upcase | | downcase }}`
`{% assign result = 'hello' | upcase | %}`

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidFilterName.ts
+++ b/packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidFilterName.ts
@@ -57,8 +57,13 @@ async function detectInvalidFilterNameInMarkup(
     const filterEndIndex = match.index! + match[0].length;
     const afterFilter = trimmedMarkup.slice(filterEndIndex);
 
+    // This regex finds invalid trailing characters after a filter name using lookaheads:
+    // 1. Skip valid syntax like ": parameter" or "| nextfilter"
+    // 2. Capture any junk characters that shouldn't be there
+    // 3. Stop before valid boundaries like colons or pipes
+    // e.g. "upcase xyz" finds "xyz", but "upcase | downcase" is ignored as valid
     const invalidSegment = afterFilter.match(
-      /^(?!\s*(?::|$|\|\s*[a-zA-Z]|\|\s*\||\s*\|\s*(?:[}%]|$|[\r\n])))([^:|]+?)(?=\s*(?::|$|\|))/,
+      /^(?!\s*(?::|$|\|\s*[a-zA-Z]|\|\s*\||\s*\|\s*(?:[}%]|$)))([^:|]+?)(?=\s*(?::|$|\|))/,
     )?.[1];
     if (!invalidSegment) {
       continue;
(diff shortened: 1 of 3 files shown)


packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidPipeSyntax.ts
```diff
@@ -0,0 +1,86 @@
...
+
+async function detectPipeSyntaxInMarkup(
+  node: LiquidVariableOutput | LiquidTag,
+  markup: string,
+): Promise<Problem<SourceCodeType.LiquidHtml>[]> {
+  const problems: Problem<SourceCodeType.LiquidHtml>[] = [];
+  const trimmedMarkup = markup.trim();
+
+  // Check for double pipes
+  const doublePipePattern = /\|\s*\|\s*/g;
```
aswamy: You could update the regex so multiple pipes area also caught. But then, you'd have to also update the message to say something more like "Empty filter found. Remove extra `|` character(s)."
```suggestion
  const extraPipesPattern = /(\|\s*){2,}/g;
```

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidPipeSyntax.ts
```diff
@@ -0,0 +1,86 @@
...
+    const markupStartInSource = node.source.indexOf(markup, node.position.start);
+    if (markupStartInSource === -1) {
+      return problems;
+    }
+    const trailingPipeStartInMarkup = trailingPipeMatch.index!;
+    const problemStartInSource = markupStartInSource + trailingPipeStartInMarkup;
+    const problemEndInSource = problemStartInSource + trailingPipeMatch[0].length;
+
+    problems.push({
+      message: `${INVALID_SYNTAX_MESSAGE} Trailing pipe detected. Remove the trailing pipe.`,
```
aswamy: Both of the invalid syntax messages you have don't have a `.` after the `${INVALID_SYNTAX_MESSAGE}`.

Also, i think we can just have one sentence after the invalid syntax message.

```suggestion
      message: `${INVALID_SYNTAX_MESSAGE}. Remove the trailing \`|\` character.`,
```

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/InvalidFilterName.ts
```diff
@@ -58,7 +58,7 @@ async function detectInvalidFilterNameInMarkup(
     const afterFilter = trimmedMarkup.slice(filterEndIndex);
 
     const invalidSegment = afterFilter.match(
-      /^(?!\s*(?::|$|\|\s*[a-zA-Z]))([^:|]+?)(?=\s*(?::|$|\|))/,
+      /^(?!\s*(?::|$|\|\s*[a-zA-Z]|\|\s*\||\s*\|\s*(?:[}%]|$|[\r\n])))([^:|]+?)(?=\s*(?::|$|\|))/,
```
aswamy: What's with the new line characters in our regex? `[\r\n]`. We have those in our markup?

Also this capture group hurts my head, can we get some comments please 💀

Issue #379, opened 2023-04-12
.prettierignore file is ignored by the plugin


kfa-spyro:
**Is your feature request related to a problem? Please describe.**
Yes, I have `*.css.liquid` files that I don't want to be formatted by Shopify VSCODE plugin. Tried using the `.prettierignore` file with proper pattern, but this doesn't work.

**Describe the solution you'd like**
Full support for .prettierignore file.

PR #1056, opened 2025-09-16
Respect .prettierignore when formatting local files

dejmedus:
- partially fixes https://github.com/Shopify/theme-tools/issues/379

Previously, the extension used liquid prettier plugin to format all files regardless of whether it was ignored by a `.prettierignore` file in the workspace. Although we aren't able to respect`.prettierignore` in all environments (for instance in [browser](https://prettier.io/docs/en/browser) or remote filesystems) we can in `nodePrettierFormat`

This PR checks for a `.prettierignore` at the workspace root and does not format files that `.getFileInfo` determines should be ignored

<details>
<summary>demo</summary>

https://github.com/user-attachments/assets/5a868e6d-9bd5-4123-a3ce-a495634b25a9

</details>

## What did you learn?

- Unlike other prettier config:
  - there is only one possible ignore [file name](https://prettier.io/docs/ignore)
  - prettier looks for `.prettierignore` files at workspace root (though its being [discussed](https://github.com/prettier/prettier/issues/4081))

## Before you deploy

### DIFF
--- a/packages/vscode-extension/src/node/formatter.spec.ts
+++ b/packages/vscode-extension/src/node/formatter.spec.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect, vi, beforeEach } from 'vitest';
 import type * as vscode from 'vscode';
 import * as prettier from 'prettier';
-import { vscodePrettierFormat, nodePrettierFormat } from '../../node/formatter';
+import { vscodePrettierFormat, nodePrettierFormat } from './formatter';
 
 vi.mock('prettier', () => ({
   format: vi.fn(),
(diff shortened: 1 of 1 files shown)


packages/vscode-extension/src/node/formatter.spec.ts
```diff
@@ -0,0 +1,74 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
```
dejmedus, line 1: It _seemed_ okay to add a test here as `src/*.spec.ts` are [excluded](https://github.com/Shopify/theme-tools/blob/main/packages/vscode-extension/tsconfig.json#L24), but i'm happy to move or remove if preferred!

PR #1061, opened 2025-09-24
Fix false reporting of UndefinedObject within visible_if

### LINKED ISSUE #685: Liquid inside `{% schema %}` should not be visited by theme-check.
> I think this is somehow related. But since theme blocks are introduced (I think) we can now connect a block value as a default. For instance, inside a section, I have this block:
> 
> ```
> "blocks": [
>     {
>       "type": "collection",
>       "name": "t:sections.featured_collections.blocks.collection.name",
>       "settings": [
>         {
>           "type": "collection",
>           "id": "collection",
>           "label": "t:sections.featured_collections.blocks.collection.collection"
>         },
>         {
>           "type": "text",
>           "id": "title",
>           "label": "t:global.text.title",
>           "default": "{{ block.settings.collection.title }}"
>         }
>   }
> ]
> ```
> 
> This is convenient and allow to automatically connect the title with the collection's title of the block.
> 
> Theme check report does not currently allow that though:
> 
> Image
> 
> I am not sure if this is expected @charlespwd  

 _Originally posted by @bakura10 in [#624](https://github.com/Shopify/theme-tools/issues/624#issuecomment-2558708446)_

This is a raw tag that considers liquid differently.

charlespwd:

We were visiting those nodes as though they were part of the execution of the file when in fact those are not parsed by the backend but stripped away.

I first tried doing that at the parsing layer by putting the body as a TextNode (the same way `{% raw %}` does it) but quickly discovered that we do depend on it being parsed for the StaticStylesheetAndJavascriptTags check.

So instead, I just added a helper that both UndefinedObject and UnusedAssign use to skip visiting of nodes that are parent'ed by a "raw-like" tag

Fixes #685

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/checks/liquid-html-syntax-error/index.ts
+++ b/packages/theme-check-common/src/checks/liquid-html-syntax-error/index.ts
@@ -9,6 +9,7 @@ import { detectInvalidLoopArguments } from './checks/InvalidLoopArguments';
 import { detectConditionalNodeUnsupportedParenthesis } from './checks/InvalidConditionalNodeParenthesis';
 import { detectInvalidFilterName } from './checks/InvalidFilterName';
 import { detectInvalidPipeSyntax } from './checks/InvalidPipeSyntax';
+import { isWithinRawTagThatDoesNotParseItsContents } from '../utils';
 
 type LineColPosition = {
   line: number;
@@ -50,6 +51,8 @@ export const LiquidHTMLSyntaxError: LiquidCheckDefinition = {
     if (!isError(ast)) {
       return {
         async BooleanExpression(node, ancestors) {
+          if (isWithinRawTagThatDoesNotParseItsContents(ancestors)) return;
+
           const problem = detectInvalidBooleanExpressions(node, ancestors);
 
           if (!problem) {
@@ -58,7 +61,9 @@ export const LiquidHTMLSyntaxError: LiquidCheckDefinition = {
 
           context.report(problem);
         },
-        async LiquidTag(node) {
+        async LiquidTag(node, ancestors) {
+          if (isWithinRawTagThatDoesNotParseItsContents(ancestors)) return;
+
           const problems = [
             detectMultipleAssignValues(node),
             detectInvalidEchoValue(node),
@@ -87,7 +92,10 @@ export const LiquidHTMLSyntaxError: LiquidCheckDefinition = {
             pipeProblems.forEach((pipeProblem) => context.report(pipeProblem));
           }
         },
-        async LiquidBranch(node) {
+
+        async LiquidBranch(node, ancestors) {
+          if (isWithinRawTagThatDoesNotParseItsContents(ancestors)) return;
+
           const problem = detectInvalidConditionalNode(node);
 
           if (!problem) {
@@ -96,7 +104,10 @@ export const LiquidHTMLSyntaxError: LiquidCheckDefinition = {
 
           context.report(problem);
         },
-        async LiquidVariableOutput(node) {
+
+        async LiquidVariableOutput(node, ancestors) {
+          if (isWithinRawTagThatDoesNotParseItsContents(ancestors)) return;
+
           const filterProblems = await detectInvalidFilterName(node, await filtersPromise);
           if (filterProblems.length > 0) {
             filterProblems.forEach((problem) => context.report(problem));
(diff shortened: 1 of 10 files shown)


packages/theme-check-common/src/checks/unused-assign/index.spec.ts
```diff
@@ -85,7 +85,7 @@ describe('Module: UnusedAssign', () => {
   });
 
   it('should not report unused assigns for things used in a raw-like tag', async () => {
-    const tags = ['style', 'javascript', 'stylesheet'];
+    const tags = ['style'];
```
charlespwd, line 88: that was wrong. 

> Liquid isn't rendered in {% javascript %} or {% stylesheet %} tags. Including Liquid code in these tags can cause syntax errors, or prevent styles from being applied to the theme.

[source](https://shopify.dev/docs/storefronts/themes/best-practices/javascript-and-stylesheet-tags#stylesheet:~:text=Liquid%20isn%27t%20rendered%20in%20%7B%25%20javascript%20%25%7D%20or%20%7B%25%20stylesheet%20%25%7D%20tags.%20Including%20Liquid%20code%20in%20these%20tags%20can%20cause%20syntax%20errors%2C%20or%20prevent%20styles%20from%20being%20applied%20to%20the%20theme.)

packages/theme-check-common/src/checks/utils.ts
```diff
@@ -103,3 +103,15 @@ export function isLoopScopedVariable(variableName: string, ancestors: LiquidHtml
 export function isLoopLiquidTag(tag: LiquidTag): tag is LiquidTagFor | LiquidTagTablerow {
   return LoopNamedTags.includes(tag.name as any);
 }
+
+const RawTagsThatDoNotParseTheirContents = ['raw', 'stylesheet', 'javascript', 'schema'];
```
karreiro, line 107: I wonder if we should include `doc` and `comment` here too. What do you think?

packages/theme-check-common/src/checks/undefined-object/index.spec.ts
```diff
@@ -432,4 +432,31 @@ describe('Module: UndefinedObject', () => {
...
+        {
+          "type": "text",
+          "label": "foo",
+          "id": "foo",
+        }
+        {
+          "type": "text",
+          "label": "bar",
+          "id": "bar",
+          "visible_if": "{{ block.settings.foo }}",
```
karreiro, line 450: In a section file, we don't get the undefined object behavior (because the `section` drop is available there):

image

But, if we try to use the `block`, we get the error:
image

It feels like a nice behavior we'd like to keep. Am I missing something?

packages/theme-check-common/src/checks/utils.ts
charlespwd: doc and comment do not parse their bodies the same way those do. I don't think we need that.

packages/theme-check-common/src/checks/undefined-object/index.spec.ts
charlespwd: No because it's wrong :D 

You _could_ use `block` in a visible_if in a section file for OS2 blocks
```
sections/os2-section.liquid
{% schema %}
{ 
  "settings": [];
  "blocks": [
    {
      "type": "foo",
      "name": "foo",
      "settings": [
         // this is actually valid in a section file
         { "id": "text", "label": "foo", "visible_if": "{{ block.setting.foo }}" }
       ]
    }
}
{% endschema %}
```

charlespwd: UndefinedObject shouldn't try to visit the schema tag because it's a coincidence that there's a section variable available there and that it works inside visible_if.

charlespwd: VisibleIf reporting the problem is actually a different bug I need to fix :P

packages/theme-check-common/src/checks/liquid-html-syntax-error/index.ts
```diff
@@ -96,7 +97,10 @@ export const LiquidHTMLSyntaxError: LiquidCheckDefinition = {
 
           context.report(problem);
         },
-        async LiquidVariableOutput(node) {
+        async LiquidVariableOutput(node, ancestors) {
+          // Do not report {{ foo == bar }} errors inside {% schema %} where they are supported
+          if (isWithinRawTagThatDoesNotParseItsContents(ancestors)) return;
```
charlespwd, line 109: Image

That's a false positive, since we do support those

packages/theme-check-common/src/checks/utils.ts
karreiro: Solved in a pairing

packages/theme-check-common/src/checks/undefined-object/index.spec.ts
karreiro: Solved in a pairing

PR #1065, opened 2025-10-01
Bump to TS 5

### LINKED ISSUE #1064: TypeScript 5.2 bump

charlespwd:

Bump to TypeScript version 5.

Only had minor issues,

1. We were pushing to the offenses array typed `Offense[]` an `Offense<T>` and it didn't like that.
2. We were using the `s` regex flag which only works in ES2018+, the compiler didn't catch that before. So we updated the tsconfig for the prettier plugin to target ES2018 as it should. 
3. Some of the node packages started return Buffer | string and we needed to adapt
4. Some of the `@types` were outdated and didn't work in TS 5

Thanks to https://github.com/microsoft/TypeScript/issues/33133#issuecomment-526277334, I was able to debug what was up because tsc was crashing and giving no valuable information

Fixes #1064 

## Before you deploy

No changeset since this is an internal change.

### DIFF
--- a/packages/release-orchestrator/src/utils.ts
+++ b/packages/release-orchestrator/src/utils.ts
@@ -1,11 +1,11 @@
 import { exec } from 'child_process';
 import crypto from 'crypto';
-import fs from 'fs';
+import fs from 'fs/promises';
 import { promisify } from 'node:util';
 import type { RunOptions, StepFunction } from './types';
 
 const execAsync = promisify(exec);
-export const readFile = promisify(fs.readFile);
+export const readFile = fs.readFile;
 
 /**
  * Creates a memoized version of a function.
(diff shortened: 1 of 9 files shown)


packages/theme-check-common/src/index.ts
```diff
@@ -157,7 +157,7 @@ function createContext<T extends SourceCodeType, S extends Schema>(
         end: getPosition(file.source, problem.endIndex),
         fix: problem.fix,
         suggest: problem.suggest,
-      } as Offense<T>);
+      } as Offense<T> as Offense);
```
charlespwd, line 160: Compiler got confused and crashed here without that extra `as Offense` since `offenses` is `Offense[]`

`Offense<T>` for some reason stopped collapsing to `Offense` in 5

packages/release-orchestrator/src/utils.ts
```diff
@@ -54,7 +54,9 @@ export const run = async (cmd: string, runInRepoRoot = true): Promise<string> =>
   const execOptions: RunOptions = runInRepoRoot
     ? { encoding: 'utf-8', cwd: await getRepoRoot() }
     : { encoding: 'utf-8' };
-  const { stdout, stderr } = await execAsync(cmd, execOptions);
+  let { stdout, stderr } = await execAsync(cmd, execOptions);
+  stdout = Buffer.isBuffer(stdout) ? stdout.toString('utf-8') : stdout;
+  stderr = Buffer.isBuffer(stderr) ? stderr.toString('utf-8') : stderr;
```
charlespwd, line 59: Apparently stdout can be a buffer, so this makes it behave like what we wanted to.

packages/theme-check-common/src/index.ts
graygilmore: Could we simplify to this then?

```suggestion
      } as Offense);
```

graygilmore (approved):

packages/theme-check-common/src/index.ts
charlespwd: Probably, but I think i wanted to make sure a JSON check didn't push a Liquid offense.

Checked this real quick and it seems like it's doing that 🤔?

PR #1063, opened 2025-09-30
introduced snippet tag in ohm

shoplando:
## Before you deploy

- [ ] This PR includes a new checks or changes the configuration of a check
  - [ ] I included a minor bump `changeset`
  - [ ] It's in the `allChecks` array in `src/checks/index.ts`
  - [ ] I ran `yarn build` and committed the updated configuration files
    
    
    
    - [ ] If applicable, I've updated the `theme-app-extension.yml` config
  - [ ] I've made a PR to update the [shopify.dev theme check docs](https://github.com/Shopify/shopify-dev/tree/main/content/storefronts/themes/tools/theme-check/checks) if applicable (link PR here).

- [ ] I included a minor bump `changeset`
- [ ] My feature is backward compatible

- [ ] I included a patch bump `changeset`

PR #1067, opened 2025-10-02
Added doc tag support logic inside snippet tag

shoplando:
Solves [#826](https://github.com/Shopify/developer-tools-team/issues/826)
Updated theme check to support `doc` tag inside of `snippet` tag. This change allows the `doc` tag to be used within inline snippets, in addition to the previously supported snippet and block contexts.

This PR also makes the theme check more rigorous by ensuring the doc tag is either:
- at the top level of a snippet/block file, OR
- a direct child of the snippet tag

The error message has been updated to reflect this new support, and tests have been added to verify that `doc` tags inside inline snippets don't trigger errors.

Update the theme check docs

## Before you deploy

 - [ ] I've made a PR to update the [shopify.dev theme check docs](https://github.com/Shopify/shopify-dev/tree/main/content/storefronts/themes/tools/theme-check/checks) if applicable.

### DIFF
--- a/packages/theme-check-common/src/checks/unsupported-doc-tag/index.ts
+++ b/packages/theme-check-common/src/checks/unsupported-doc-tag/index.ts
@@ -1,6 +1,7 @@
 import { isSnippet, isBlock } from '../../to-schema';
 import { LiquidCheckDefinition, Severity, SourceCodeType } from '../../types';
 import { filePathSupportsLiquidDoc } from '../../liquid-doc/utils';
+import { LiquidRawTag, NodeTypes, LiquidHtmlNode } from '@shopify/liquid-html-parser';
 
 export const UnsupportedDocTag: LiquidCheckDefinition = {
   meta: {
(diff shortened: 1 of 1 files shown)


shoplando:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/1067?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1067** <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/1067?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/1067?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1063** <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/1063?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/theme-check-common/src/checks/unsupported-doc-tag/index.ts
```diff
@@ -19,18 +20,26 @@ export const UnsupportedDocTag: LiquidCheckDefinition = {
...
+        const isDirectChildOfSnippetTag =
+          immediateParent?.type === NodeTypes.LiquidTag && immediateParent.name === snippetTagName;
+
+        if ((isInSnippetOrBlockFile && isTopLevelInFile) || isDirectChildOfSnippetTag) {
+          return;
+        }
+
         context.report({
-          message: `The \`${docTagName}\` tag can only be used within a snippet or block.`,
+          message: `The \`${docTagName}\` tag can only be used at the top level of a snippet or block file, or as a direct child of an inline snippet tag.`,
```
graygilmore: (nit) Wondering if this could be clearer but nothing better is springing to mind. Might be worth workshopping a little bit to get "at the top level of" clearer.

shoplando: I totally agree, I'll try to come up with something clearer

shoplando: How does "The \`${docTagName}\` tag cannot be nested inside Liquid tags except for snippet tags." sound?

EvilGenius13: Throwing in my two cents. Would it make sense to offer two different messages?
We keep the message for doc tags in snippets and blocks as `The \${docTagName}\ tag can only be used at the top level of a snippet or block file`
and then have a separate one triggered for inline snippets like `The ${docTagName} tag can only be used inside an inline snippet at the top level. (I don't have a good message for this)

My only reason for this is that the doc tags are now serving a purpose in different ways that would be confusing in one message.

We're top level only within blocks and snippet files but if we have an inline snippet that can be placed anywhere, it actually goes top level _within_ the snippet itself.

shoplando: Hmm, yeah that's actually a great idea, I'll make two separate messages for the two cases

Issue #827, opened 2025-02-27
Invalid completion for `metaobject` type schema settings


david-w-shopify:
**Describe the bug**
On adding a `metaobject` schema setting in a section, for example, the `default` key is suggested.

![Image](https://github.com/user-attachments/assets/21f13e2d-2c2a-4a30-a169-cbc7a78ee0b0)

On trying to upload a setting with this key, the CLI reports that `default` is not a valid attribute

<img width="570" alt="Image" src="https://github.com/user-attachments/assets/3d45d43f-7ebf-44a0-b06a-4ad178cee722" />

**Source**

```liquid
// sections/test.liquid

{% schema %}
{
  "name": "Test",
  "settings": [
    {
      "type": "metaobject",
      "id": "test",
      "label": "Test",
      "metaobject_type": "test",
      "default": "test"
    }
  ]
}
{% endschema %}
```

**Expected behaviour**
`default` is not suggested as a key for `metaobject` schema settings

**Actual behaviour**
`default` is suggested as a key for `metaobject` schema settings

**Debugging information**
 - OS: masOs
 - OS Version: Sequoia 15.3.1

PR #1068, opened 2025-10-08
Added support for render tag completion for inline snippets

shoplando:
Solves [#827](https://github.com/Shopify/developer-tools-team/issues/827)
Added completion support for inline snippets in the render tag. Now when using the `{% render %}` tag, the language server will suggest both file-based snippets and inline snippets defined within the same file. Inline snippets are displayed with a helpful documentation note indicating they are "defined in this file".

### DIFF
--- a/packages/theme-language-server-common/src/completions/providers/RenderSnippetCompletionProvider.ts
+++ b/packages/theme-language-server-common/src/completions/providers/RenderSnippetCompletionProvider.ts
@@ -1,7 +1,9 @@
-import { NodeTypes } from '@shopify/liquid-html-parser';
+import { LiquidHtmlNode, LiquidTag, NodeTypes } from '@shopify/liquid-html-parser';
 import { CompletionItem, CompletionItemKind } from 'vscode-languageserver';
 import { LiquidCompletionParams } from '../params';
 import { Provider } from './common';
+import { SourceCodeType, visit, isError } from '@shopify/theme-check-common';
+import { findInlineSnippetAncestor } from '@shopify/theme-check-common/dist/checks/utils';
 
 export type GetSnippetNamesForURI = (uri: string) => Promise<string[]>;
 
@@ -14,21 +16,13 @@ export class RenderSnippetCompletionProvider implements Provider {
     const { node, ancestors } = params.completionContext;
     const parentNode = ancestors.at(-1);
 
-    if (
-      !node ||
-      !parentNode ||
-      node.type !== NodeTypes.String ||
-      parentNode.type !== NodeTypes.RenderMarkup
-    ) {
+    if (!node || !parentNode || parentNode.type !== NodeTypes.RenderMarkup) {
       return [];
     }
 
-    const options = await this.getSnippetNamesForURI(params.textDocument.uri);
-    const partial = node.value;
-
-    return options
-      .filter((option) => option.startsWith(partial))
-      .map(
+    if (node.type === NodeTypes.String) {
+      const fileSnippets = await this.getSnippetNamesForURI(params.textDocument.uri);
+      const fileCompletionItems = fileSnippets.map(
         (option: string): CompletionItem => ({
           label: option,
           kind: CompletionItemKind.Snippet,
@@ -38,5 +32,42 @@ export class RenderSnippetCompletionProvider implements Provider {
           },
         }),
       );
+      return fileCompletionItems;
+    } else if (node.type === NodeTypes.VariableLookup) {
+      const containingSnippet = findInlineSnippetAncestor(ancestors);
+      const containingSnippetName =
+        containingSnippet && typeof containingSnippet.markup !== 'string'
+          ? containingSnippet.markup.name
+          : null;
+
+      const fullAst = params.document.ast;
+      const allInlineSnippets = isError(fullAst) ? [] : getInlineSnippetsNames(fullAst);
+
+      const inlineSnippets = allInlineSnippets.filter((name) => name !== containingSnippetName);
+
+      const inlineCompletionItems = inlineSnippets.map(
+        (option: string): CompletionItem => ({
+          label: option,
+          kind: CompletionItemKind.Snippet,
+          documentation: {
+            kind: 'markdown',
+            value: `Inline snippet "${option}"`,
+          },
+        }),
+      );
+      return inlineCompletionItems;
+    } else {
+      return [];
+    }
   }
+}
+
+function getInlineSnippetsNames(ast: LiquidHtmlNode): string[] {
+  return visit<SourceCodeType.LiquidHtml, string>(ast, {
+    LiquidTag(node: LiquidTag) {
+      if (node.name === 'snippet' && typeof node.markup !== 'string' && node.markup.name) {
+        return node.markup.name;
+      }
+    },
+  });
 }
--- a/.changeset/flat-geckos-breathe.md
+++ b/.changeset/flat-geckos-breathe.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-language-server-common': minor
+---
+
+Added completion support for inline snippets in the `render` tag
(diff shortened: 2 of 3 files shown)


shoplando:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/1068?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#1068** <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/1068?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/1068?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1067** <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/1067?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1063** <a href="https://app.graphite.dev/github/pr/Shopify/theme-tools/1063?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/theme-language-server-common/src/completions/providers/RenderSnippetCompletionProvider.ts
```diff
@@ -14,29 +15,59 @@ export class RenderSnippetCompletionProvider implements Provider {
...
+
+  const inlineSnippetNames = visit<SourceCodeType.LiquidHtml, string>(ast, {
+    LiquidTag(node: LiquidTag) {
+      if (node.name === 'snippet' && typeof node.markup !== 'string' && node.markup.name) {
+        return node.markup.name;
+      }
+    },
+  });
+
+  return inlineSnippetNames.filter((name): name is string => !!name);
```
aswamy:
```suggestion
  return visit<SourceCodeType.LiquidHtml, string>(ast, {
    LiquidTag(node: LiquidTag) {
      if (node.name === 'snippet' && typeof node.markup !== 'string' && node.markup.name) {
        return node.markup.name;
      }
    },
  });
```

packages/theme-language-server-common/src/completions/providers/RenderSnippetCompletionProvider.ts
```diff
@@ -14,29 +15,59 @@ export class RenderSnippetCompletionProvider implements Provider {
...
+      const inlineSnippets = getInlineSnippetsNames(params.completionContext.partialAst);
+      const inlineCompletionItems = inlineSnippets
+        .filter((option) => option.startsWith(partial))
+        .map(
+          (option: string): CompletionItem => ({
+            label: option,
+            kind: CompletionItemKind.Snippet,
+            documentation: {
+              kind: 'markdown',
+              value: `Inline snippet (defined in this file)`,
```
aswamy: Is this temporary documentation until you pick up `doc` tag's description?

packages/theme-language-server-common/src/completions/providers/RenderSnippetCompletionProvider.ts
```diff
@@ -14,29 +15,59 @@ export class RenderSnippetCompletionProvider implements Provider {
...
+              value: `snippets/${option}.liquid`,
+            },
+          }),
+        );
+      return fileCompletionItems;
+    } else if (node.type === NodeTypes.VariableLookup) {
+      partial = node.name || '';
+      const inlineSnippets = getInlineSnippetsNames(params.completionContext.partialAst);
+      const inlineCompletionItems = inlineSnippets
+        .filter((option) => option.startsWith(partial))
```
aswamy: I think you should remove that this `startsWith` filter entirely for this one, and the one above yours because VSCode will determine the best way a completion can be complete. e.g. what if you typo a variable, or what if the characters you were looking for are in the middle of a word.

aswamy (approved):
Minor comments, but overall looks great! 🥳

packages/theme-language-server-common/src/completions/providers/RenderSnippetCompletionProvider.ts
shoplando: I wasn't sure what to put as the description, and since the regular snippet's description is just `snippets/${option}.liquid`, I thought saying that the inline snippet defined in the file would be enough. 
Should we include the `doc` tag's description? And if we do, should we include that for regular snippets files?

aswamy: Oh dang, i didn't know `doc` tag isn't used when you do code-completion for file snippets...
image

That might have been a mistake since on hover, we do show the docs:
image

How about this, keep your description, but match the file snippet description style (maybe change it to `Inline snippet "<name-of-snippet>"`. But also, make a bug ticket and fix it for both snippets to use the doc tag. If the file snippet or inline snippet has a doc tag, use its description on code completion, if not, use the default string you have.

packages/theme-language-server-common/src/completions/providers/RenderSnippetCompletionProvider.ts
shoplando: I've removed the filter, but now the tests in `RenderSnippetCompletionProvider.spec.ts` doesn't pass anymore since the filtering logic is done on VSCode:
The test expects only `product-cart` to appear as completion suggestion when we write `{% render 'product`, but now without the filter, `image` also shows up in the test.
I was wondering if we should change the tests to match the language server behaviour (no filtering), or try to somehow mock a VSCode environment so that it can properly test the actual behaviour that we'd see on VSCode?

aswamy: I think the test is incorrect and we should remove it - i dont think we should be doing the `startsWith` anymore. I know we (most likely me) used to do this, but VSCode is pretty good at handling options without us filtering it down first.

packages/theme-language-server-common/src/completions/providers/RenderSnippetCompletionProvider.ts
```diff
@@ -38,5 +31,32 @@ export class RenderSnippetCompletionProvider implements Provider {
...
+        }),
+      );
+      return inlineCompletionItems;
+    } else {
+      return [];
+    }
   }
 }
+
+function getInlineSnippetsNames(ast: LiquidHtmlNode): string[] {
```
dejmedus, line 65: Another scope/nesting question: do we want to limit the options to "current level" snippets? For example, should the below only offer `header_inner`?

<details>
<summary>example</summary>

Image

</details>

dejmedus (approved):
Looks great to me! I left you a question about nested snippets but its not blocking

<details><summary> 🎩  </summary>

Image

</details>

packages/theme-language-server-common/src/completions/providers/RenderSnippetCompletionProvider.ts
shoplando: Oooh yeah you're right! Thanks for catching that 🙏

shoplando: Now I'm wondering if we should include it in other snippets? ie:

dejmedus: Yeah this is tricky, `hi-snip` _could_ be passed in via render to `hi-snip-2`, but should tooling expect that? Maybe we could just limit completion to prevent a snippet being offered inside itself and leave the rest as is? 

@aswamy do you think that would make sense?

aswamy: > Maybe we could just limit completion to prevent a snippet being offered inside itself and leave the rest as is?

I think you're right. For simplicity, we should only offer code completion to things that are explicitly defined within the snippet.

This includes ALL variables. The reason is that we aren't sure how the snippet is rendered, which can get very confusing.

PR #1071, opened 2025-10-09
Updated the undefined-object theme check to skip over inline-snippets

shoplando:

Disabled the "Undefined Object" theme check for the new `snippet` tag to prevent false positives.
This change also includes making the check scope-aware based on inline snippets' doc tag parameters.

## Before you deploy

### DIFF
--- a/packages/theme-check-common/src/checks/undefined-object/index.ts
+++ b/packages/theme-check-common/src/checks/undefined-object/index.ts
@@ -7,6 +7,7 @@ import {
   LiquidTagDecrement,
   LiquidTagFor,
   LiquidTagIncrement,
+  LiquidTagSnippet,
   LiquidTagTablerow,
   LiquidVariableLookup,
   NamedTags,
@@ -16,7 +17,7 @@ import {
 import { LiquidCheckDefinition, Severity, SourceCodeType, ThemeDocset } from '../../types';
 import { isError, last } from '../../utils';
 import { hasLiquidDoc } from '../../liquid-doc/liquidDoc';
-import { isWithinRawTagThatDoesNotParseItsContents } from '../utils';
+import { isWithinRawTagThatDoesNotParseItsContents, findInlineSnippetAncestor } from '../utils';
 
 type Scope = { start?: number; end?: number };
 
@@ -38,13 +39,14 @@ export const UndefinedObject: LiquidCheckDefinition = {
   create(context) {
     const relativePath = context.toRelativePath(context.file.uri);
     const ast = context.file.ast;
+    const isSnippetFile = relativePath.startsWith('snippets/');
 
     if (isError(ast)) return {};
 
     /**
      * Skip this check when a snippet does not have the presence of doc tags.
      */
-    if (relativePath.startsWith('snippets/') && !hasLiquidDoc(ast)) return {};
+    if (isSnippetFile && !hasLiquidDoc(ast)) return {};
 
     /**
      * Skip this check when definitions for global objects are unavailable.
@@ -56,7 +58,9 @@ export const UndefinedObject: LiquidCheckDefinition = {
     const themeDocset = context.themeDocset;
     const scopedVariables: Map<string, Scope[]> = new Map();
     const fileScopedVariables: Set<string> = new Set();
-    const variables: LiquidVariableLookup[] = [];
+    const variables: Array<{ node: LiquidVariableLookup; inlineSnippet: LiquidTag | null }> = [];
+    const inlineSnippetsWithDocTags: Set<number> = new Set();
+    let snippetFileHasDocTag = false;
 
     function indexVariableScope(variableName: string | null, scope: Scope) {
       if (!variableName) return;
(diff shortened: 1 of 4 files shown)


packages/theme-check-common/src/checks/undefined-object/index.ts
```diff
@@ -142,6 +142,10 @@ export const UndefinedObject: LiquidCheckDefinition = {
         const parent = last(ancestors);
         if (isLiquidTag(parent) && isLiquidTagCapture(parent)) return;
 
+        if (parent?.type === NodeTypes.RenderMarkup && parent.snippet === node) return;
+
+        if (isLiquidTag(parent) && parent.name === 'snippet' && parent.markup === node) return;
```
aswamy: Is this supposed to be a temporary fix or something instead of just adding it our list of valid liquid tags?

shoplando: I already added `snippet` to our list of valid liquid tags (in `packages/liquid-html-parser/src/types.ts`) but it still warns me 🤔

EvilGenius13: Is this because we load the tags from a different [repo](https://github.com/Shopify/theme-tools/blob/69282bb3deabaf6e2ee798064f09cf393dc0efc0/packages/theme-check-docs-updater/src/themeLiquidDocsManager.ts#L43)?

shoplando: I think that repo is used for the documentation, autocomplete and validation of the tag itself. And also the `capture` tag is using a similar approach by skipping over the newly defined object in the capture tag so that the theme check won't warn you about it (I might just use the same format as the capture case)

packages/theme-check-common/src/checks/undefined-object/index.ts
```diff
@@ -66,9 +67,16 @@ export const UndefinedObject: LiquidCheckDefinition = {
...
+      async LiquidDocParamNode(node: LiquidDocParamNode, ancestors: LiquidHtmlNode[]) {
         const paramName = node.paramName?.value;
-        if (paramName) {
+        if (!paramName) return;
+        const snippetAncestor = findInlineSnippetAncestor(ancestors);
+        if (snippetAncestor) {
+          indexVariableScope(paramName, {
+            start: snippetAncestor.blockStartPosition.end,
+            end: snippetAncestor.blockEndPosition?.start,
+          });
```
aswamy, line 92: TIL Cool didn't even know we had a variable scope like this

packages/theme-check-common/src/checks/undefined-object/index.ts
```diff
@@ -142,6 +154,8 @@ export const UndefinedObject: LiquidCheckDefinition = {
         const parent = last(ancestors);
         if (isLiquidTag(parent) && isLiquidTagCapture(parent)) return;
 
+        if (isLiquidTag(parent) && isLiquidTagSnippet(parent) && parent.markup === node) return;
```
aswamy: What's the `parent.markup === node` part supposed to catch?

packages/theme-check-common/src/checks/undefined-object/index.ts
aswamy, line 1: So i think you have the right logic, but i think you need to modify the logic on line 48:

```
if (relativePath.startsWith('snippets/') && !hasLiquidDoc(ast)) return {};
```

This might not make too much sense, but here is the logic:
- There exists real world snippets (especially old ones) that DONT have `doc` tag
- The code above ensures that we only have UndefinedObject errors on snippets IF they have `doc` tag because technically, you can pass ANYTHING into snippets, and without a `doc` tag we don't really know which one is Defined, and which one isn't
- But by introducing inline snippets, we have a slight problem: having a `doc` tag inside the inline snippets don't mean the top-level snippet file has one too...

So we need a robust solution that works for the following situations:

- What happens when you have no doc tags in the parent snippet AND the inline snippets
- What happens when you have doc tag in the parent, but not in the inline snippets
- What happens when you have doc tag in the inline snippets, but not in the parent
- What happens when you have doc tag in both

I think we need to layout each situation before full implementation of this

packages/theme-check-common/src/checks/undefined-object/index.ts
shoplando: That's an oversight on my part. I think I added that before creating the isLiquidTagSnippet() function and forgot to remove it. Thanks!

packages/theme-check-common/src/checks/undefined-object/index.ts
shoplando: Ok so basically we would want:
- no checks at all if both the snippet file and inline snippet don't have any doc tags
- yes checks when there is a doc tag at its respective scope, for example when there is a doc tag in the main snippet file and none inside the inline snippet:
```
{%- comment -%} snippets/my-snippet.liquid {%- endcomment -%}

{% doc %}
  @param {string} parent_param
{% enddoc %}

{{ parent_param }}         <- ✅ Should NOT warn
{{ undefined_in_parent }}  <- ❓ Should this warn? YES

{% snippet inline_one %}
  {{ undefined_in_inline }} <- ❓ Should this warn? NO
{% endsnippet %}
```
And when the doc tag is inside the inline snippet, the check will be activated only on the inside of the inline snippet, and not outside of it.
I'll implement this right now.

PR #829, opened 2025-02-27
Split ValidRenderSnippetParams into MissingRenderSnippetParams and UnrecognizedRenderSnippetParams to enable individually toggling behaviour

jamesmengo:
## Before you deploy

- [ ] This PR includes a new checks or changes the configuration of a check
  - [ ] I included a minor bump `changeset`
  - [ ] It's in the `allChecks` array in `src/checks/index.ts`
  - [ ] I ran `yarn build` and committed the updated configuration files
    
    
    
    - [ ] If applicable, I've updated the `theme-app-extension.yml` config
  - [ ] I've made a PR to update the [shopify.dev theme check docs](https://github.com/Shopify/shopify-dev/tree/main/content/storefronts/themes/tools/theme-check/checks) if applicable (link PR here).

- [ ] I included a minor bump `changeset`
- [ ] My feature is backward compatible

- [ ] I included a patch bump `changeset`

PR #1073, opened 2025-10-14
Added hover support for inline snippets on `render` tags

shoplando:
Solves [#829](https://github.com/Shopify/developer-tools-team/issues/829)
Added hover support for `render` handle and named arguments for inline snippets.

- [ x ] I included a minor bump `changeset`
- [ x ] My feature is backward compatible

### DIFF
--- a/packages/theme-language-server-common/src/hover/providers/RenderSnippetParameterHoverProvider.ts
+++ b/packages/theme-language-server-common/src/hover/providers/RenderSnippetParameterHoverProvider.ts
@@ -1,5 +1,10 @@
 import { NodeTypes } from '@shopify/liquid-html-parser';
-import { LiquidHtmlNode, GetDocDefinitionForURI } from '@shopify/theme-check-common';
+import {
+  LiquidHtmlNode,
+  GetDocDefinitionForURI,
+  findInlineSnippet,
+  extractDocDefinition,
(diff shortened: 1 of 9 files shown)


packages/theme-language-server-common/src/hover/providers/RenderSnippetParameterHoverProvider.ts
```diff
@@ -13,20 +18,31 @@ export class RenderSnippetParameterHoverProvider implements BaseHoverProvider {
...
+        params.textDocument.uri,
+        'snippets',
+        parentNode.snippet.value,
+      );
+    } else if (parentNode.snippet.type === NodeTypes.VariableLookup) {
+      const snippetName = parentNode.snippet.name || '';
+      const rootNode = ancestors[0];
+      const snippetNode = findInlineSnippet(rootNode, snippetName);
+      if (snippetNode) {
+        docDefinition = extractDocDefinition(params.textDocument.uri, snippetNode);
```
dejmedus, line 43: I ran into this with the arg completion: since `visit` grabs all the doc params (even nested), maybe we could re-use the [helper function](https://github.com/Shopify/theme-tools/pull/1075#discussion_r2445306327) from the other branch we talked about (as I think it grabs the immediate doc tag params)

<details><summary>example</summary>

Image

</details>

PR #1075, opened 2025-10-17
Support code completion for inline snippet render arguments

dejmedus:
- closes https://github.com/Shopify/developer-tools-team/issues/828

While rendering an inline snippet that contains a doc tag we should be offered argument completion suggestions 

inline-snippets-arg-completion

<details><summary>Demo</summary>

https://github.com/user-attachments/assets/154d911d-1669-4265-ba20-5f1b79e57ac0

</details> 

## How to test

Pull down the branch `gh pr checkout 1075` 

- Completion options defined a snippets doc tag should display as expected
- Only unused options should display
- Options from nested snippet doc tags should not be offered, only the current scope

> Notes: Since this branch isn't based off of the doc tags branch red squiggles will show up on snippets, but render completion should work as expected. VSCode displays param definitions, Cursor does not

## Before you deploy

### DIFF
--- a/packages/theme-language-server-common/src/completions/providers/RenderSnippetParameterCompletionProvider.ts
+++ b/packages/theme-language-server-common/src/completions/providers/RenderSnippetParameterCompletionProvider.ts
@@ -9,7 +9,11 @@ import {
 } from 'vscode-languageserver';
 import { CURSOR, LiquidCompletionParams } from '../params';
 import { Provider } from './common';
-import { formatLiquidDocParameter, getParameterCompletionTemplate } from '../../utils/liquidDoc';
+import {
+  formatLiquidDocParameter,
+  getInlineSnippetDocParams,
+  getParameterCompletionTemplate,
+} from '../../utils/liquidDoc';
 import { GetDocDefinitionForURI } from '@shopify/theme-check-common';
 
 export type GetSnippetNamesForURI = (uri: string) => Promise<string[]>;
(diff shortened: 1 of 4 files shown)


packages/theme-language-server-common/src/completions/providers/RenderSnippetParameterCompletionProvider.ts
```diff
@@ -77,3 +90,50 @@ export class RenderSnippetParameterCompletionProvider implements Provider {
...
+  return paramNodes.map(
+    (node): LiquidDocParameter => ({
+      nodeType: 'param',
+      name: node.paramName.value,
+      description: node.paramDescription?.value ?? null,
+      type: node.paramType?.value ?? null,
+      required: node.required,
+    }),
+  );
+}
```
shoplando: I think it might be worth moving this helper function into an utils file like [liquidDoc](https://github.com/Shopify/theme-tools/blob/main/packages/theme-language-server-common/src/utils/liquidDoc.ts) since this could be a very helpful function for hovering and some theme checks like `MissingRenderSnippetArguments` or `UnrecognizedRenderSnippetArgument`

shoplando (approved):
Left a nit comment, but overall looks good for me! Tophatting also worked as expected 👍

packages/theme-language-server-common/src/completions/providers/RenderSnippetParameterCompletionProvider.ts
dejmedus: Good call! I moved it out:)

EvilGenius13 (approved):

PR #1095, opened 2025-12-31
Allow disabling theme docset augmentation for vanilla liquid LSP

yoavbls:
Right now, there is no way to disable the augmentation of the docset.
This PR adds the option to disable augmentation and remove completions that are not part of the Liquid language specification.

Adds optional `augmentDocset?: boolean` dependency (default true) so consumers can run a vanilla Liquid LSP without wrapping the docset in `AugmentedThemeDocset`.

## Before you deploy

### DIFF
--- a/packages/theme-language-server-browser/package.json
+++ b/packages/theme-language-server-browser/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/theme-language-server-browser",
-  "version": "2.20.0",
+  "version": "2.20.1",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "author": "CP Clermont <cp.clermont@shopify.com>",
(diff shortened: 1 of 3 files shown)


packages/theme-language-server-common/src/types.ts
```diff
@@ -98,6 +98,14 @@ export interface RequiredDependencies {
    * This should only be used in node environments; not on the browser.
    */
   fetchMetafieldDefinitionsForURI: (uri: URI) => Promise<void>;
+
+  /**
+   * Controls augmentation of the provided ThemeDocset.
```
graygilmore, line 103: Maybe we could be a little more specific here. Something like:


```suggestion
   * Controls whether Shopify-specific Liquid should be included
```

And we could even rename the option to something like `includeShopifyTags` or something like that. @Shopify/developer-tools what do you think?

PR #1096, opened 2026-01-03
Allow Liquid code if input type is liquid

andershagbard:
This PR allows to use Liquid code, if the input type is "liquid". Prior to this, it would warn the that this was invalid.

E.g.:
```json
{
  "settings": [
    {
      "id": "text_value",
      "type": "liquid",
      "label": "Text Value",
      "default": "{% render 'block' %}"
    }
  ]
}
```

### DIFF
--- a/packages/theme-check-common/src/checks/liquid-free-settings/index.ts
+++ b/packages/theme-check-common/src/checks/liquid-free-settings/index.ts
@@ -74,18 +74,18 @@ function isLiteralNode(node: JSONNode): node is LiteralNode {
 }
 
 function isLiquidType(ancestors: JSONNode[]): boolean {
-  return ancestors.some((ancestor) => {
-    if (ancestor.type !== 'Object') {
+  const parentJsonNode = ancestors.at(-1)
+  
+  if (!parentJsonNode || parentJsonNode.type !== 'Object') {
+    return false;
+  }
+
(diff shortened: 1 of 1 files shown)


packages/theme-check-common/src/checks/liquid-free-settings/index.ts
```diff
@@ -67,6 +73,22 @@ function isLiteralNode(node: JSONNode): node is LiteralNode {
...
+    return ancestor.children.some(({ key, value }) => {
+      if (key.value !== 'type' || !isLiteralNode(value)) {
+        return false;
+      }
+
+      return value.value === 'liquid';
+    });
+  });
+}
+
```
aswamy, line 91: Your `isLiquidType` might be looking too far up the ancestor hierarchy. We want to make sure the block is of `type: "liquid"`, not any ancestor has `type: "liquid"`

```suggestion
function isLiquidType(ancestors: JSONNode[]): boolean {
  const parentJsonNode = ancestors.at(-1)
  
  if (!parentJsonNode || parentJsonNode.type !== 'Object') {
    return false;
  }

  return parentJsonNode.children.some(({ key, value }) => {
    if (key.value !== 'type' || !isLiteralNode(value)) {
      return false;
    }

    return value.value === 'liquid';
  });
}

```

PR #1112, opened 2026-01-28
Add browser integration test baseline for vscode-extension

charlespwd:
## Summary

Adds browser integration test infrastructure for the vscode-extension package, enabling end-to-end validation of the extension's functionality in a real browser environment.

### Browser Test Infrastructure (3a943b95)
- **Webpack test config** with browser polyfills for running tests in web environment
- **Mocha test entry point** for organizing and running browser-based tests
- **Activation smoke test** verifying the extension loads correctly in browser

### Watch Mode for Development (9cb3b434)
- **`build:test:watch`** script for auto-rebuild on file changes
- **`WATCH_MODE`** env var to keep browser open after tests complete (no auto-close)
- **E2e autocomplete test** that opens `layout/theme.liquid`, types `{{ prod`, and asserts "product" appears in autocomplete suggestions

## Scripts Added

| Script | Purpose |
|--------|---------|
| `test:web` | Run browser tests (CI mode, auto-closes) |
| `test:web:watch` | Run browser tests with debug delay |
| `build:test:watch` | Watch mode for test file changes |

### Running Tests
```bash
# Run tests once (CI mode)
yarn --cwd packages/vscode-extension test:web
```

### Watch Mode Development
```bash
# Terminal 1: Watch and rebuild on changes
yarn --cwd packages/vscode-extension build:test:watch

# Terminal 2: Run tests with browser kept open
yarn -cwd packages/vscode-extension test:web:watch
```

Edit test files, save, refresh browser to see results immediately.

## Test Plan

- [x] `yarn test` passes
- [x] `yarn --cwd packages/vscode-extension test:web` passes in CI mode
- [x] Watch mode rebuilds on file changes
- [x] `WATCH_MODE=1` keeps browser open after tests
- [x] Autocomplete e2e test validates "product" suggestion appears

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,30 @@ jobs:
       - name: Check formatting
         run: yarn format:check
 
+  e2e:
+    runs-on: ubuntu-latest
+    name: E2E tests
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          submodules: true
+
+      - name: Setup Node.js
+        uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0
+        with:
+          node-version: 20
+          cache: "yarn"
+          cache-dependency-path: "yarn.lock"
+
+      - name: Install the packages
+        run: yarn install --frozen-lockfile --ignore-engines
+
+      - name: Build the code
+        run: yarn build
+
+      - name: Run the E2E tests
+        run: yarn test:e2e
+
   tests:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "postrelease": "yarn --cwd packages/prettier-plugin-liquid prerelease",
     "reset": "find . -name \"node_modules\" -type d -prune -exec rm -rf '{}' +",
     "test": "vitest",
+    "test:e2e": "yarn test:e2e:browser",
     "test:e2e:browser": "yarn --cwd packages/vscode-extension test:web",
     "theme-check": "node packages/theme-check-node/dist/cli.js",
     "type-check": "yarn workspaces run type-check"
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -31,6 +31,30 @@ jobs:
...
+    name: E2E tests
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          submodules: true
+
+      - name: Setup Node.js
+        uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0
+        with:
+          node-version: 20
```
aswamy, line 45: I noticed in the same workflow file we used `node-version: ${{ matrix.node-version }}` instead. Unless we want this explicit version, i think we should let Claude know to follow existing standards when touching files.

packages/vscode-extension/.eslintrc.js
```diff
@@ -19,4 +19,12 @@ module.exports = {
     'node': true,
   },
   ignorePatterns: ['out', 'dist', '**/*.d.ts'],
+  overrides: [
+    {
+      files: ['test/**/*.ts'],
```
aswamy, line 24: What do you think about putting our e2e tests into an `e2e-test` folder instead?

I think we usually have our unit tests and unit test helpers in `src/test` folder. Just don't want them confused 🤷

.github/workflows/ci.yml
charlespwd: This job doesn't have a matrix. Using a matrix would return undefined.

packages/vscode-extension/.eslintrc.js
charlespwd: The vain part of me doesn't like `e2e-test` as a folder name because it looks ugly :P 

The other part is that they write very differently. I think that's fine? There's also a README in the test folder that explains how the E2E tests work. And it's generally how that's done with the yo generator for VS Code extensions. 

Do we usually use e2e-test as a folder in other repos?

aswamy: nah - just wondering if it matters when we want to explicitly say "unit test" or "e2e test". But readme is there so its fine.

PR #1114, opened 2026-01-30
Make liquid-html-parser ESM-compatible for @ohm-js/wasm

charlespwd:
## Summary

Resolves ESM/CJS incompatibility so liquid-html-parser can consume the ESM-only @ohm-js/wasm package.

## Key Changes

- **ESM tsconfig**: Configure TypeScript for ESM output (`module: "ESNext"`, `moduleResolution: "bundler"`)
- **@ohm-js/wasm/compat imports**: Use the compatibility API that works in both Node.js and browser
- **`res.use()` patterns**: Replace global `use()` calls with resource-scoped `res.use()` for grammar loading
- **Webpack fix for OHM_DEBUG**: Define `process.env.OHM_DEBUG` in browser bundle to prevent runtime errors

## Test Results

- **Node.js**: 1737/1740 tests pass (3 expected message format changes due to ohm-js version)
- **Browser**: 3/3 tests pass

## Context

Builds on Patrick's PR #1070 work for browser test infrastructure.


.github/workflows/ci.yml
```diff
@@ -15,12 +15,12 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0
         with:
-          node-version: ${{ matrix.node-version }}
+          node-version: 20
```
charlespwd, line 18: This one doesn't have a matrix. Makes sense!

PR #1118, opened 2026-02-03
Upgrade Prettier to v3

charlespwd:

Upgrade Prettier from v2.8.4 to v3.0.0. This update includes formatting changes across multiple files, primarily affecting interface declarations and conditional expressions. The changes are mostly related to line breaks and indentation.  
  
## Why?

Because ohm-js/wasm is making us newer features of ESM that prettier 2 doesn't support. 

## Before you deploy

`/aside` Claude did the work but I did 🎩 and had it fix a couple of typing issues over a couple of loops.

### DIFF
--- a/.changeset/sweet-needles-fetch.md
+++ b/.changeset/sweet-needles-fetch.md
@@ -1,0 +1,10 @@
+---
+"@shopify/theme-language-server-common": patch
+"@shopify/codemirror-language-client": patch
+"@shopify/prettier-plugin-liquid": patch
+"@shopify/liquid-html-parser": patch
+"@shopify/theme-check-common": patch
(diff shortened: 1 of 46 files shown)


charlespwd:
yessir. It's mostly just prettier3 at the root + yarn format. 

I _did_ have to deal with the prettier-plugin-liquid and how it supports prettier2 AND prettier3, but that stuff is stuffed in the webpack config changes. Basically we already imported prettier@2 as prettier2 in node_modules and prettier@# as prettier3. Most of the types in prettier-plugin-liquid were from prettier 2 so for the import to resolve correctly I had to change things there a little bit. 

Our entire test suite still passes + type-check + format:check. We're good :D

PR #1120, opened 2026-02-05
Fix const reassignment error in standalone build

### LINKED ISSUE #1109: Build error in standalone prettier plugin tries to reassign const
**Describe the bug**
Image
Plugin throws above issue when used.

**Expected behaviour**
Plugin should format code

**Actual behaviour**
https://github.com/Shopify/theme-tools/blob/b4cd18194fa12b5a9735aa0eeff0e15f4e6c831b/packages/prettier-plugin-liquid/src/printer/printer-liquid-html.ts#L631-L640
In the minified build, this becomes (reformatted for readability):
```js
function C(e, t) {
  const r = e[t];
  return (
    Array.isArray(r) ||
    (null !== (i = r) &&
      "object" == typeof i &&
      "type" in i &&
      "string" == typeof i.type)
  );
  var i;
}
```
The issue is because the "chunk" this function is inside already has `const i = r(849)` earlier, so instead of the `var` getting hoisted and assigned, it tries to reassign the `const` and throws an error. (in fact when i checked in the browser sources, the `var i;` statement was missing completely - possibly due to some DCE)

**Debugging information**
 - MacOS
 - Tahoe 26.2

charlespwd:
## Summary
- Inline `isNode` into `hasOrIsNode` to prevent terser from creating problematic const reassignment when building standalone bundle

## Problem
The standalone build fails with:
```
[ERROR] Cannot reassign "const" variable
```

This happens because terser inlines the single-use `isNode` function but generates invalid code that reassigns a const variable.

## Solution
Inline the `isNode` logic directly into `hasOrIsNode`, eliminating the separate function that terser was problematically inlining.

## Test plan
- [x] Standalone build completes without error
- [x] `import('@shopify/prettier-plugin-liquid/standalone')` works
- [x] Existing tests pass

Fixes #1109

### DIFF
--- a/packages/prettier-plugin-liquid/src/printer/printer-liquid-html.ts
+++ b/packages/prettier-plugin-liquid/src/printer/printer-liquid-html.ts
@@ -633,5 +633,8 @@ function hasOrIsNode<N extends LiquidHtmlNode, K extends keyof N>(node: N, key: 
   // this works because there's no ()[] type that is string | Node, it only
   // happens for singular nodes such as name: string | LiquidDrop, etc.
   // Note: isNode logic inlined to avoid terser const reassignment in standalone build
-  return Array.isArray(v) || (v !== null && typeof v === 'object' && 'type' in v && typeof v.type === 'string');
+  return (
+    Array.isArray(v) ||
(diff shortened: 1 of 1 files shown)


charlespwd:
Assuming report in #1109 is valid. This fix seems harmless.

PR #1121, opened 2026-02-05
Fix RemoteAsset false positives for hash URLs and variable lookups

### LINKED ISSUE #982: RemoteAsset shouldn't throw for hash URLs
**Describe the bug**

That's a perfectly valid URL. 

Image

**Source**

```liquid
    <link
      rel="expect"
      href="#MainContent"
      blocking="render"
      id="view-transition-render-blocker"
    >
```

**Expected behaviour**
No error. URLs that start with `#` are links to anchors on the page. This is not a remote URL as it is a link to the same page. 

**Actual behaviour**
Error

### LINKED ISSUE #995: RemoteAsset shouldn't throw for video_source.url
**Describe the bug**
source.url in the following code throw error: `Use one of the asset_url filters to serve assets for better performance. theme-check[RemoteAsset](https://shopify.dev/docs/storefronts/themes/tools/theme-check/checks/remote-asset)`

**Source**

```liquid
{% for source in video.sources %}
  <source src='{{ source.url }}' type='{{ source.mime_type }}'>
{% endfor %}
```

**Expected behaviour**
As the [video_source](https://shopify.dev/docs/api/liquid/objects/video_source#video_source).url property returns a CDN url, it doesn't return the right URL when the ` | asset_url` filter is added.

**Actual behaviour**
N/A

**Debugging information**
 - MacOS 15.4.1
 - @shopify/theme-check-common 3.19.0
 - shopify.theme-check-vscode 3.9.7

**Additional context**
N/A

charlespwd:
## Summary

Fixes #995, Fixes #982

This PR fixes two false positive issues in the `RemoteAsset` check:

1. **Hash URLs** (`#anchor`, `#section`) were incorrectly flagged as remote assets
2. **Variable-based URLs** (like `video_source.url`) were incorrectly flagged when static analysis can't determine what's in the variable

## Changes

- Add `isHashUrl()` helper to detect hash-only URLs and skip linting them
- Add `startsWithVariableLookup()` check to skip linting when attribute values start with a `VariableLookup` (since we can't statically know what's in variables)
- Add comprehensive test cases for both patterns

## Test plan

- [x] Added unit tests for hash URL patterns
- [x] Added unit tests for variable lookup patterns
- [x] Existing tests continue to pass

### DIFF
--- a/packages/theme-check-common/src/checks/remote-asset/index.spec.ts
+++ b/packages/theme-check-common/src/checks/remote-asset/index.spec.ts
@@ -183,19 +183,39 @@ describe('Module: RemoteAsset', () => {
     );
   });
 
-  it('should report an offense for image drops without img_url filter', async () => {
+  it('should not report an offense for variable-based URLs', async () => {
     const sourceCode = `
       <img src="{{ image }}">
       <img src="{{ image.src }}">
+      <source src="{{ source.url }}" type="{{ source.mime_type }}">
+      <source src="{{ video_source.url }}">
     `;
 
     const offenses = await runLiquidCheck(RemoteAsset, sourceCode);
-    expect(offenses).to.have.length(2);
-    offenses.forEach((offense) => {
-      expect(offense.message).to.equal(
-        'Use one of the asset_url filters to serve assets for better performance.',
-      );
-    });
+    expect(offenses).to.be.empty;
+  });
+
+  it('should not report an offense for hash URLs', async () => {
+    const sourceCode = `
+      <link rel="expect" href="#MainContent" blocking="render">
+      <link href="#section-header" rel="preload">
+      <img src="#placeholder">
+    `;
+
+    const offenses = await runLiquidCheck(RemoteAsset, sourceCode);
+    expect(offenses).to.be.empty;
+  });
+
+  it('should still report an offense for string literals without asset_url filter', async () => {
+    const sourceCode = `
+      <img src="{{ 'image.png' }}" />
+    `;
+
+    const offenses = await runLiquidCheck(RemoteAsset, sourceCode);
+    expect(offenses).to.have.length(1);
+    expect(offenses[0].message).to.equal(
+      'Use one of the asset_url filters to serve assets for better performance.',
+    );
   });
 
   it('should not report an offence if url is a shopify CDN', async () => {
--- a/packages/theme-check-common/src/checks/remote-asset/index.ts
+++ b/packages/theme-check-common/src/checks/remote-asset/index.ts
@@ -43,6 +43,29 @@ function isLiquidVariable(node: LiquidHtmlNode | string): node is LiquidVariable
   return typeof node !== 'string' && node.type === NodeTypes.LiquidVariable;
 }
 
+function isHashUrl(url: string): boolean {
+  return url.startsWith('#');
+}
+
+/**
+ * Checks if the attribute value starts with a variable lookup.
+ * When a value starts with a VariableLookup (e.g., {{ source.url }}, {{ image }}),
+ * we can't statically analyze what it contains, so we skip linting.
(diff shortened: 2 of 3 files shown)


packages/theme-check-common/src/checks/remote-asset/index.spec.ts
```diff
@@ -183,19 +183,43 @@ describe('Module: RemoteAsset', () => {
     );
   });
 
-  it('should report an offense for image drops without img_url filter', async () => {
+  it('should not report an offense for variable-based URLs (Fixes #995)', async () => {
```
graygilmore: Do we generally include PR numbers like this?

```suggestion
  it('should not report an offense for variable-based URLs', async () => {
```

packages/theme-check-common/src/checks/remote-asset/index.spec.ts
```diff
@@ -183,19 +183,43 @@ describe('Module: RemoteAsset', () => {
...
-    offenses.forEach((offense) => {
-      expect(offense.message).to.equal(
-        'Use one of the asset_url filters to serve assets for better performance.',
-      );
-    });
+    expect(offenses).to.be.empty;
+  });
+
+  it('should not report an offense for hash URLs (Fixes #982)', async () => {
+    // Hash URLs are internal page anchors, not remote assets
```
graygilmore:
```suggestion
```

packages/theme-check-common/src/checks/remote-asset/index.spec.ts
```diff
@@ -183,19 +183,43 @@ describe('Module: RemoteAsset', () => {
...
+      <link href="#section-header" rel="preload">
+      <img src="#placeholder">
+    `;
+
+    const offenses = await runLiquidCheck(RemoteAsset, sourceCode);
+    expect(offenses).to.be.empty;
+  });
+
+  it('should still report an offense for string literals without asset_url filter', async () => {
+    // String literals (not variables) should still be checked
```
graygilmore:
```suggestion
```

packages/theme-check-common/src/checks/remote-asset/index.ts
```diff
@@ -43,6 +43,30 @@ function isLiquidVariable(node: LiquidHtmlNode | string): node is LiquidVariable
...
+function startsWithVariableLookup(attr: ValuedHtmlAttribute): boolean {
+  const firstNode = attr.value[0];
+  if (!firstNode) return false;
+
+  if (!isLiquidVariableOutput(firstNode)) return false;
+
+  const variable = firstNode.markup;
+  if (!isLiquidVariable(variable)) return false;
+
+  // If the expression is a VariableLookup (not a String), skip linting
```
graygilmore:
```suggestion
```

packages/theme-check-common/src/checks/remote-asset/index.ts
```diff
@@ -157,6 +181,16 @@ export const RemoteAsset: LiquidCheckDefinition<typeof schema> = {
...
       if (!urlAttribute) return;
 
+      // Hash URLs are internal page anchors, not remote assets
+      const firstTextNode = urlAttribute.value.find(
+        (node): node is TextNode => node.type === NodeTypes.TextNode,
+      );
+      if (firstTextNode && isHashUrl(firstTextNode.value)) return;
+
+      // When value starts with a VariableLookup, we can't statically analyze it
+      // Skip linting to avoid false positives (e.g., video_source.url)
```
graygilmore:
```suggestion
```

packages/theme-check-common/src/checks/remote-asset/index.spec.ts
charlespwd: Fixed :)

Issue #942, opened 2025-05-09
{% schema %} checking of `t:` translations only seems to be checking `name` properties.


charlespwd:
**Describe the bug**
A setting of type `select` should throw errors if the `label` of its `options` don't exist. Right now we have no support for those. 

Looks like that's also a problem for a preset's `category` field

**Source**

```liquid
{% schema %}
{
  "name": "t:general.group",
  "blocks": [{ "type": "@theme" }],
  "settings": [
    {
      "type": "select",
      "id": "layout_direction",
      "label": "t:labels.layout_direction",
      "options": [
        { "value": "group--horizontal", "label": "t:options.direction.horizontal" },
        { "value": "group--vertical", "label": "t:options.direction.vertical" }
      ],
      "default": "group--vertical"
    },
    {
      "type": "select",
      "id": "alignment",
      "label": "t:labels.alignment",
      "options": [
        { "value": "group--left-aligned", "label": "t:options.alignment.left" },
        { "value": "group--center-aligned", "label": "t:options.alignment.center" },
        { "value": "group--right-aligned", "label": "t:options.alignment.right" }
      ],
      "visible_if": "{{ block.settings.layout_direction == 'group--vertical' }}"
    },
    {
      "type": "range",
      "id": "padding",
      "label": "t:labels.padding",
      "default": 0,
      "min": 0,
      "max": 100,
      "step": 1,
      "unit": "px",
    }
  ],
  "presets": [
    {
      "name": "t:general.column",
      "category": "t:categories.alignment.left",
      "settings": {
        "layout_direction": "group--vertical",
        "alignment": "group--left-aligned",
        "padding": 0
      }
    },
    {
      "name": "t:general.row",
      "category": "t:categories.layout",
      "settings": {
        "layout_direction": "group--horizontal",
        "padding": 0
      }
    }
  ]
}
{% endschema %}

```

**Expected behaviour**
I got a couple of things in there that should show errors because the labels don't exist.

**Actual behaviour**
`name` properties seem to be linted OK. 
`label` inside select options doesn't. 
`category` inside preset doesn't`

**Debugging information**
 - OS [e.g. Windows, Mac, Linux]
 - OS Version
 - Theme Check Version [e.g. 0.1.0]

**Additional context**

PR #1123, opened 2026-02-05
valid-schema-translations theme check

aswamy:

Part of https://github.com/Shopify/theme-tools/issues/942

Added a new theme check called `ValidSchemaTranslations` that ensures all translation keys (t:) in schema have matching entries in the default schema translations file. This check supersedes the existing `ValidSchemaName` which only checked the `schema.name` value (along with a few more things).

The PR:

- Creates a new check that recursively finds all string values starting with 't:' in schema JSON
- Verifies each translation key has a corresponding entry in the translations file
- Reports error messages for missing translations
- Moves the translation existence check from `ValidSchemaName` to this new dedicated check

## Tophat

Here is a schema you can test out the new theme-check (assuming you don't have `t:general.headerfoo` in your `en.default.schema.json`

```
{% schema %}
{
  "name": "t:general.headerfoo",
  "default": {
    "text": "t:general.headerfoo",
    "select": "option",
    "radio": "option",
  },
  "settings": [
    {
      "type": "text",
      "id": "text",
      "label": "t:general.headerfoo",
    },
    {
      "type": "select",
      "id": "select",
      "label": "t:general.headerfoo",
      "options": [
        {
          "group": "Select Group 1",
          "label": "t:general.headerfoo",
          "value": "option",
        },
        {
          "group": "Select Group 1",
          "label": "Option 2",
          "value": "option2",
        },
      ]
    },
    {
      "type": "radio",
      "id": "radio",
      "label": "Radio",
      "options": [
        {
          "label": "Option",
          "value": "option",
        }
      ],
    }
  ],
  "presets": [
    {
      "name": "Alok Section",
      "category": "t:general.headerfoo",
      "settings": {
        "text": "t:general.headerfoo",
        "select": "option",
        "radio": "option",
      }
    }
  ]
}
{% endschema %}
```

- Updating shopify-dev so that we have it documented

## Before you deploy

    - [ ] I've made a PR to update the [shopify.dev theme check docs](https://github.com/Shopify/shopify-dev/tree/main/content/storefronts/themes/tools/theme-check/checks) if applicable.
        - I will do so after this PR since we aren't cutting a release for this theme-check yet

### DIFF
--- a/.changeset/loud-hotels-teach.md
+++ b/.changeset/loud-hotels-teach.md
@@ -1,0 +1,6 @@
+---
+"@shopify/theme-check-common": minor
+"@shopify/theme-check-node": minor
+---
+
+New theme check to ensure translation values inside of `schema` tag are valid
--- a/packages/theme-check-common/src/checks/index.ts
+++ b/packages/theme-check-common/src/checks/index.ts
@@ -59,6 +59,7 @@ import { ValidLocalBlocks } from './valid-local-blocks';
 import { ValidRenderSnippetArgumentTypes } from './valid-render-snippet-argument-types';
 import { ValidSchema } from './valid-schema';
 import { ValidSchemaName } from './valid-schema-name';
+import { ValidSchemaTranslations } from './valid-schema-translations';
 import { ValidSettingsKey } from './valid-settings-key';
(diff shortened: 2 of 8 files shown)


charlespwd (approved):
Beautiful 👌

PR #1124, opened 2026-02-06
Fix rename cursor boundary for LiquidDoc param names

### LINKED ISSUE #936: Can't rename LiquidDoc param when cursor is at the first character of the var name
**Describe the bug**
F2 when the cursor is at the beginning of the param name shows "you can't rename this thing". Wondering if that's because of the same findCurentNode thing that I fixed last week. 

**Source**

```liquid
{% doc %}
  @param {collection} █collection - The collection object to display.
{% enddoc %}
```

**Expected behaviour**
Can rename the thing. (Works if the cursor is 1 to the right of it, e.g. `c█ollection`

**Actual behaviour**
The element can't be renamed.

**Debugging information**
 - OS [e.g. Windows, Mac, Linux]
 - OS Version
 - Theme Check Version [e.g. 0.1.0]

**Additional context**

charlespwd:
## Summary

Fixes #936

Rename does not trigger when the cursor is at the first character of a `@param` name in a LiquidDoc block. 

### DIFF
--- a/packages/theme-language-server-common/src/rename/providers/LiquidVariableRenameProvider.spec.ts
+++ b/packages/theme-language-server-common/src/rename/providers/LiquidVariableRenameProvider.spec.ts
@@ -156,6 +156,32 @@ describe('LiquidVariableRenameProvider', () => {
       );
     });
 
+    it('does not rename type annotation when type and param name share the same text', async () => {
+      const collisionSource = `{% doc %}@param {food} food - favorite food{% enddoc %}
+{% liquid
+  echo food
+%}`;
+      const collisionDoc = TextDocument.create(textDocumentUri, 'liquid', 1, collisionSource);
+      documentManager.open(collisionDoc.uri, collisionSource, 1);
+
+      const params = {
+        textDocument: collisionDoc,
+        position: Position.create(0, collisionSource.indexOf('} food') + 2),
+        newName: 'meal',
+      };
+
+      const result = await provider.rename(params);
+      assert(result);
+      assert(result.documentChanges);
+      expect((result.documentChanges[0] as TextDocumentEdit).edits).to.applyEdits(
+        collisionDoc,
+        `{% doc %}@param {food} meal - favorite food{% enddoc %}
+{% liquid
+  echo meal
+%}`,
+      );
+    });
+
     it('returns new name when cursor is at first character of liquid doc param name', async () => {
       const params = {
         textDocument,
--- a/packages/theme-language-server-common/src/rename/providers/LiquidVariableRenameProvider.ts
+++ b/packages/theme-language-server-common/src/rename/providers/LiquidVariableRenameProvider.ts
@@ -98,7 +98,9 @@ export class LiquidVariableRenameProvider implements BaseRenameProvider {
       AssignMarkup: replaceRange,
       ForMarkup: replaceRange,
       TextNode: (node: LiquidHtmlNode, ancestors: (LiquidHtmlNode | JSONNode)[]) => {
-        if (ancestors.at(-1)?.type !== NodeTypes.LiquidDocParamNode) return;
+        const parentNode = ancestors.at(-1);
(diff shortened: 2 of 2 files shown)


packages/theme-language-server-common/src/rename/providers/LiquidVariableRenameProvider.spec.ts
```diff
@@ -156,6 +156,31 @@ describe('LiquidVariableRenameProvider', () => {
       );
     });
 
+    it('returns new name when cursor is at first character of liquid doc param name', async () => {
+      const params = {
+        textDocument,
+        position: Position.create(0, 25),
```
charlespwd: Would be nicer to read if this was some kind of indexOf -> position instead.

packages/theme-language-server-common/src/rename/providers/LiquidVariableRenameProvider.ts
```diff
@@ -76,6 +80,13 @@ export class LiquidVariableRenameProvider implements BaseRenameProvider {
     if (document.ast instanceof Error) return null;
     if (!supportedTags(node, ancestors)) return null;
 
+    // When node is a LiquidDocParamNode, ensure cursor is on the param name, not the @ prefix
+    if (node.type === NodeTypes.LiquidDocParamNode) {
+      const cursorOffset = textDocument.offsetAt(params.position);
+      const nameEnd = node.paramName.position.start + node.paramName.value.length;
+      if (cursorOffset < node.paramName.position.start || cursorOffset > nameEnd) return null;
+    }
```
charlespwd: That confuses me. `@param` is literally the word before it. Should the comment be updated/clarified? Does this make any sense?

packages/theme-language-server-common/src/rename/providers/LiquidVariableRenameProvider.spec.ts
```diff
@@ -326,6 +351,38 @@ describe('LiquidVariableRenameProvider', () => {
...
+{% liquid
+  for prod in products
+    echo prod.title
+  endfor
+  echo ppp
+%}`,
+        );
+      });
+
+      it('does not trigger rename when cursor is at @ symbol position', async () => {
```
charlespwd: Should we add does not trigger rename on _everything_ before the name?
```
@param name
```

Why just check `@`?

packages/theme-language-server-common/src/rename/providers/LiquidVariableRenameProvider.ts
```diff
@@ -76,6 +80,13 @@ export class LiquidVariableRenameProvider implements BaseRenameProvider {
     if (document.ast instanceof Error) return null;
     if (!supportedTags(node, ancestors)) return null;
 
+    // When node is a LiquidDocParamNode, ensure cursor is on the param name, not the @param keyword or type annotation
+    if (node.type === NodeTypes.LiquidDocParamNode) {
+      const cursorOffset = textDocument.offsetAt(params.position);
+      const nameEnd = node.paramName.position.start + node.paramName.value.length;
+      if (cursorOffset < node.paramName.position.start || cursorOffset > nameEnd) return null;
+    }
```
charlespwd, line 88: I wonder if we should actually change the logic for findCurrentNode instead... should probably check how hover behave. If hover is also bugged we might want to revisit.

charlespwd: Oh we don't have a hover for that one.

charlespwd: Aight after some research the approach in the PR is better. 

- Hover is silly in the doc itself, since the doc is right after it.
- Completion doesn't make sense, since we're defining it.

Drawbacks of changing isCovered in visitor.ts:
- It would require changes to how `TextNode` calculates isCovered which would create an ambiguity for things like `<tag>TextNode</tag>`. Right now it's resolved by not matching on start so the `>` matches on the HtmlElement.

charlespwd:
Folded that fix in :) TY for reporting!

PR #1128, opened 2026-02-17
Integrate ohm-js v18 for faster parsing

pdubroy:
Same goals as #1116, but that PR identified several issues in the Ohm wasm packages that Claude had worked around.

The issues have been fixed upstream. I've also reorganized the packages somewhat, and expect it to be fairly stable now.


packages/liquid-html-parser/src/stage-1-cst.ts
```diff
@@ -531,21 +532,6 @@ export type LiquidDocConcreteNode =
...
-  type: ConcreteNodeTypes;
-  locStart: (node: Node[]) => number;
-  locEnd: (node: Node[]) => number;
-  source: string;
-  [k: string]: FunctionMapping | string | number | boolean | object | null;
-}
-
-type TopLevelFunctionMapping = (...nodes: Node[]) => any;
-type FunctionMapping = (nodes: Node[]) => any;
-
```
charlespwd, line 548: Oh that's nice :D

PR #1167, opened 2026-03-27
add tabstops to translations variables

### LINKED ISSUE #537: Feature Request: Tab stops for translations
After adding a translation, going backwards to add my variables is slow. It could be faster with built in tabstops.

On injecting a translation, add a tabstop for each variable that's already injected.

For example, given this in `locales/en.default.json`

```json
{
  "product": {
    "add_count_product_to_cart": "Add {{ count }} {{ product }} to cart"
  }
}
```

The following completion would be used

```
{{ 'product.add_count_product_to_cart' | t: count: ${1:count}, product: ${2:product} }}
```

This would mean I can immediately replace the "count" and "product" variables if I need to.

---

As more powerful alternative, the completion could look like this

```
{{ 'product.add_count_product_to_cart' | t: ${1:count: ${2:count}, }${3:product: ${4:product}} }}
```

This would mean I can remove a variable entirely if I desire. This is useful for scenarios where a translation will be rendered in JS, rather than rendered entirely via Liquid.

I am assuming that extensions use the same snippet syntax as user configured snippets here. If they don't and it's not clear, hopefully this video will help get the idea across

https://github.com/user-attachments/assets/6ed8ced9-ef5e-4c79-b188-89b4c32a50dc

FKauwe:
This fix adds tabstops to all the translation completions-- after selecting the autocomplete label from the dropdown, and pressing `tab`, the editor will now highlight the variable text to be filled by the user and the user will be able to type directly into this space, then `tab` forward and write text over the next variable (if there are multiple variables).
This will only work for translations that have variables.

I wrote a new function in `translations.ts` that creates the string with the Snippet syntax.  There are Plain Text strings and Snippet strings, the previously used function returned PlainText and the tabstops behavior is only available on a Snippet.

The function gets called in the TranslationCompletionProvider and the insertTextFormat flag also gets set to Snippet now, so the editor knows to interpret the string as Snippet and surface the tabstops behavior to the user. 

The existing spec file passes with my change and I added a new test to verify the new behavior

Fixes #537 
## How to Tophat
Off of main (old behavior): To trigger the translation key autocomplete with variables, go to any .liquid file and type:
`{{ 'search.res`
and pause for autocomplete to pop up. Select `search.results_for_html` and accept it, then notice that the variables are not highlighted and must be selected manually to be typed over (screenshots are in the linked ticket)
To see the new behavior, pull down my branch, function F5 to open the Extension Development instance in any theme and repeat the process.  Now you should see the variables highlighted, with the ability to tab to each one and type over. 
Here is a screen recording of the correct, tophatted behavior:

https://github.com/user-attachments/assets/212da252-4ad7-48dd-bd4e-f63bba067183

## What did you learn?

I learned about the block element  █ .
in the context of the tests in the spec file, the block element is used by the test framework to mock the cursor position so the test knows when to trigger the autocomplete functionality and must be used when testing autocomplete behavior (which my change falls under) but isn't needed to test things like the dropdown labels being present or the documentation pop-up. 

## Before you deploy

- [x] I included a patch `changeset`

### DIFF
--- a/packages/theme-language-server-common/src/translations.ts
+++ b/packages/theme-language-server-common/src/translations.ts
@@ -81,3 +81,8 @@ export function paramsString(params: string[]) {
   if (params.length === 0) return '';
   return `: ` + params.map((param) => `${param}: ${param}`).join(', ');
 }
+
+export function paramsWithTabstops(params: string[]) {
+  if (params.length === 0) return '';
+  return ': ' + params.map((param, index) => `${param}: \${${index + 1}:${param}}`).join(', ');
+}
--- a/.changeset/breezy-sheep-repair.md
+++ b/.changeset/breezy-sheep-repair.md
@@ -1,0 +1,5 @@
+---
+"@shopify/theme-language-server-common": patch
+---
+
+add tabstops to translations variables
--- a/packages/theme-language-server-common/src/completions/providers/TranslationCompletionProvider.spec.ts
+++ b/packages/theme-language-server-common/src/completions/providers/TranslationCompletionProvider.spec.ts
@@ -1,4 +1,5 @@
 import { describe, beforeEach, it, expect } from 'vitest';
+import { InsertTextFormat } from 'vscode-languageserver';
 import { DocumentManager } from '../../documents';
 import { CompletionsProvider } from '../CompletionsProvider';
 import { MetafieldDefinitionMap } from '@shopify/theme-check-common';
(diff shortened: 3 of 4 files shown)


packages/theme-language-server-common/src/translations.ts
```diff
@@ -81,3 +81,8 @@ export function paramsString(params: string[]) {
   if (params.length === 0) return '';
   return `: ` + params.map((param) => `${param}: ${param}`).join(', ');
 }
+
+export function paramsSnippetString(params: string[]) {
```
aswamy: `paramsWithTabstops` I just don't want to use the term "snippet" if it's not an actual "snippet" file

aswamy (approved):
Great stuff! Tophatted with 0, 1, and 2+ tabstops and everything works as expected. One small suggestion.

EvilGenius13 (approved):

packages/theme-language-server-common/src/translations.ts
FKauwe: summary of offline discussion:
there are snippet files which are related to themes and there are snippet strings (what is being used in this PR and completely different)
To avoid confusion, I've changed the name of the function to Alok's suggestion

PR #1168, opened 2026-04-02
Update theme checks for modern practices

graygilmore:
## Summary

Fixes 3 theme checks that produce false positives on valid Liquid patterns:

- **MatchingTranslations**: Exclude `customer_accounts.*` translation paths (like the existing `shopify.*` exclusion) — these are platform-provided translations that themes don't control. Also updates `TranslationCompletionProvider` to exclude `customer_accounts.*` from autocomplete suggestions.
- **ValidBlockTarget**: Allow local blocks (type `_name`) in presets when the schema has no root-level `blocks` declaration. Uses `hasRootBlocksDeclaration` from `getBlocks()` instead of checking `rootLevelThemeBlocks.length` to correctly handle schemas where the only root block is `@app`. Same fix applied to `isInvalidDefaultBlock`.
- **UnclosedHTMLElement**: Cross-identifier balancing for sibling conditionals in the same grandparent — handles patterns where one condition opens a tag and a different condition closes it (e.g. `forloop.first` opens a wrapper div and `forloop.last` closes it). Extracts duplicated stack-matching logic into a `balanceOpenCloseTags` helper.

## 🎩 Tophatting

Test each fix locally by creating a theme with the Liquid examples below and running theme-check against it. Each example should produce **zero** offenses with this PR but would error on `main`.

### 1. MatchingTranslations — `customer_accounts.*` exclusion

Create a theme with `customer_accounts` translations in the default locale that are missing from a non-default locale:

**`locales/en.default.json`**
```json
{
  "customer_accounts": {
    "order": {
      "title": "Order",
      "subtitle": "Order details"
    }
  },
  "general": {
    "hello": "Hello"
  }
}
```

**`locales/fr.json`**
```json
{
  "customer_accounts": {
    "order": {
      "title": "Commande"
    }
  },
  "general": {
    "hello": "Bonjour"
  }
}
```

**Expected**: No `MatchingTranslations` offense for the missing `customer_accounts.order.subtitle` in `fr.json`. On `main`, this reports an error because `fr.json` is missing a key present in the default locale. A mismatch in `general.*` should still be reported.

### 2. ValidBlockTarget — local blocks in presets without root `blocks` array

Create a section (or theme block) that references a local block type (`_card`) in presets but does **not** declare a root-level `blocks` array:

**`blocks/_card.liquid`**
```liquid
<div>{{ block.settings.title }}</div>

{% schema %}
{
  "name": "Card",
  "settings": [
    { "type": "text", "id": "title", "label": "Title" }
  ]
}
{% endschema %}
```

**`sections/cards.liquid`**
```liquid
{% for block in section.blocks %}
  {% render block %}
{% endfor %}

{% schema %}
{
  "name": "Cards",
  "presets": [
    {
      "name": "Default",
      "blocks": {
        "card-1": { "type": "_card" }
      }
    }
  ]
}
{% endschema %}
```

**Expected**: No `ValidBlockTarget` offense — the section has no root `blocks` array, so `_card` is valid. If you add `"blocks": [{ "type": "@app" }]` to the section schema (root blocks with only `@app`), a `ValidBlockTarget` error **should** appear because there's now a declared blocks list that doesn't include `_card`.

### 3. UnclosedHTMLElement — cross-identifier conditional balancing

Create a template that opens a tag inside one condition and closes it inside a different condition within the same parent:

**`sections/grid.liquid`**
```liquid
<div class="grid">
  {% for item in section.blocks %}
    {% if forloop.first %}
      <div class="grid-row">
    {% endif %}

    <div class="grid-item">{{ item.settings.title }}</div>

    {% if forloop.last %}
      </div>
    {% endif %}
  {% endfor %}
</div>

{% schema %}
{
  "name": "Grid"
}
{% endschema %}
```

**Expected**: No `UnclosedHTMLElement` offense — the `<div class="grid-row">` opened under `forloop.first` is balanced by the `</div>` closed under `forloop.last` within the same grandparent. If you change the closing tag to `</section>` (mismatched name), an offense **should** appear.

### DIFF
--- a/packages/theme-check-common/src/checks/matching-translations/index.ts
+++ b/packages/theme-check-common/src/checks/matching-translations/index.ts
@@ -48,7 +48,8 @@ export const MatchingTranslations: JSONCheckDefinition = {
     const hasDefaultTranslations = () => defaultTranslations.size > 0;
     const isTerminalNode = ({ type }: JSONNode) => type === 'Literal';
     const isPluralizationNode = (node: PropertyNode) => PLURALIZATION_KEYS.has(node.key.value);
-    const isShopifyPath = (path: string) => path.startsWith('shopify.');
+    const isExternalPath = (path: string) =>
+      path.startsWith('shopify.') || path.startsWith('customer_accounts.');
 
     const hasDefaultTranslation = (translationPath: string) =>
       defaultTranslations.has(translationPath) ?? false;
@@ -129,7 +130,7 @@ export const MatchingTranslations: JSONCheckDefinition = {
         if (!hasDefaultTranslations()) return;
         if (isPluralizationNode(node)) return;
         if (!isTerminalNode(node.value)) return;
-        if (isShopifyPath(path)) return;
+        if (isExternalPath(path)) return;
 
         if (hasDefaultTranslation(path)) {
           // As `path` is present, we remove it from the
@@ -158,7 +159,7 @@ export const MatchingTranslations: JSONCheckDefinition = {
           const closest = closestTranslationKey(path);
 
           if (isPluralizationPath(path)) return;
-          if (isShopifyPath(path)) return;
+          if (isExternalPath(path)) return;
 
           context.report({
             message: `The translation for '${path}' is missing`,
--- a/packages/theme-check-common/src/checks/unclosed-html-element/index.ts
+++ b/packages/theme-check-common/src/checks/unclosed-html-element/index.ts
@@ -139,6 +139,13 @@ export const UnclosedHTMLElement: LiquidCheckDefinition = {
 
       async onCodePathEnd() {
         for (const [grandparent, stacks] of stacksByGrandparent) {
+          // First pass: match opens/closes within each condition identifier.
+          // Collect unmatched nodes with their identifier for cross-identifier balancing.
+          const unmatchedByIdentifier = new Map<
+            ConditionIdentifer,
+            (HtmlElement | HtmlDanglingMarkerClose)[]
+          >();
+
           for (const identifier of stacks.identifiers) {
             const openNodes = stacks.open.get(identifier) ?? [];
             const closeNodes = stacks.close.get(identifier) ?? [];
@@ -154,24 +161,30 @@ export const UnclosedHTMLElement: LiquidCheckDefinition = {
             //   Then we're going to push on open and pop when the close match.
             // If a close doesn't match,
             //   Then we'll push it onto the stack and everything after won't match.
-            const stack = [] as (HtmlElement | HtmlDanglingMarkerClose)[];
-            for (const node of nodes) {
-              if (node.type === NodeTypes.HtmlElement) {
-                stack.push(node);
-              } else if (
-                stack.length > 0 &&
-                getName(node) === getName(stack.at(-1)!) &&
-                stack.at(-1)!.type === NodeTypes.HtmlElement &&
-                node.type === NodeTypes.HtmlDanglingMarkerClose
-              ) {
-                stack.pop();
-              } else {
-                stack.push(node);
-              }
-            }
+            unmatchedByIdentifier.set(identifier, balanceOpenCloseTags(nodes));
+          }
 
-            // At the end, whatever is left in the stack is a reported offense.
-            for (const node of stack) {
+          // Second pass: try to balance unmatched opens from one condition
+          // identifier against unmatched closes from sibling condition identifiers
+          // within the same grandparent.
+          //
+          // This handles patterns where an open tag is inside one condition
+          // (e.g. `{% if forloop.first %}`) and the matching close is inside a
+          // sibling condition (e.g. `{% if forloop.last %}`).
+          const allUnmatched = ([] as (HtmlElement | HtmlDanglingMarkerClose)[])
+            .concat(...unmatchedByIdentifier.values())
+            .sort((a, b) => a.position.start - b.position.start);
+
+          const crossBalancedStack = balanceOpenCloseTags(allUnmatched);
+
+          // Build a set of nodes still unmatched after cross-identifier balancing.
+          // Nodes NOT in this set were resolved and should not be reported.
+          const crossBalancedRemaining = new Set(crossBalancedStack);
+
+          for (const [identifier, unmatched] of unmatchedByIdentifier) {
+            for (const node of unmatched) {
+              if (!crossBalancedRemaining.has(node)) continue;
+
               if (node.type === NodeTypes.HtmlDanglingMarkerClose) {
                 context.report({
                   message: `Closing tag does not have a matching opening tag for condition \`${identifier}\` in ${
@@ -282,6 +295,27 @@ function negateIdentifier(conditionIdentifier: ConditionIdentifer): ConditionIde
   return conditionIdentifier.startsWith('-')
     ? conditionIdentifier.slice(1)
     : `-${conditionIdentifier}`;
+}
+
+function balanceOpenCloseTags(
+  sortedNodes: (HtmlElement | HtmlDanglingMarkerClose)[],
+): (HtmlElement | HtmlDanglingMarkerClose)[] {
+  const stack: (HtmlElement | HtmlDanglingMarkerClose)[] = [];
+  for (const node of sortedNodes) {
+    if (node.type === NodeTypes.HtmlElement) {
+      stack.push(node);
+    } else if (
+      stack.length > 0 &&
+      getName(node) === getName(stack.at(-1)!) &&
+      stack.at(-1)!.type === NodeTypes.HtmlElement &&
+      node.type === NodeTypes.HtmlDanglingMarkerClose
+    ) {
+      stack.pop();
+    } else {
+      stack.push(node);
+    }
+  }
+  return stack;
 }
 
 function getName(node: LiquidHtmlNode) {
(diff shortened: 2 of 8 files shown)


packages/theme-check-common/src/checks/matching-translations/index.ts
```diff
@@ -48,7 +48,8 @@ export const MatchingTranslations: JSONCheckDefinition = {
     const hasDefaultTranslations = () => defaultTranslations.size > 0;
     const isTerminalNode = ({ type }: JSONNode) => type === 'Literal';
     const isPluralizationNode = (node: PropertyNode) => PLURALIZATION_KEYS.has(node.key.value);
-    const isShopifyPath = (path: string) => path.startsWith('shopify.');
+    const isExternalPath = (path: string) =>
```
graygilmore, line 51: 💡 **Improvement**: The `TranslationCompletionProvider` at `packages/theme-language-server-common/src/completions/providers/TranslationCompletionProvider.ts:46-47` has a similar filter that excludes `shopify`-prefixed translations from autocomplete suggestions but does not yet exclude `customer_accounts`. While this is a UX concern (noise in completions) rather than a false-positive linting error, the two locations now have divergent lists of external translation prefixes. As more external prefixes are added, maintaining independent hardcoded lists becomes a maintenance risk.

**Suggestion:** Extract a shared constant like:
```ts
export const EXTERNAL_TRANSLATION_PREFIXES = ['shopify.', 'customer_accounts.'] as const;
```
Then use it in both `isExternalPath` here and the filter in `TranslationCompletionProvider` (line 46-47) so they stay in sync.

packages/theme-check-common/src/checks/unclosed-html-element/index.ts
```diff
@@ -170,8 +177,44 @@ export const UnclosedHTMLElement: LiquidCheckDefinition = {
...
+          // within the same grandparent.
+          //
+          // This handles patterns where an open tag is inside one condition
+          // (e.g. `{% if forloop.first %}`) and the matching close is inside a
+          // sibling condition (e.g. `{% if forloop.last %}`).
+          const allUnmatched = ([] as (HtmlElement | HtmlDanglingMarkerClose)[])
+            .concat(...unmatchedByIdentifier.values())
+            .sort((a, b) => a.position.start - b.position.start);
+
+          const crossBalancedStack = [] as (HtmlElement | HtmlDanglingMarkerClose)[];
```
graygilmore: 📐 **Design**: The stack-based open/close matching algorithm in the second pass (lines 194-207) is structurally identical to the first pass (lines 164-177 in existing code). Both iterate over sorted nodes, push opens, pop when a close matches the top-of-stack open by name, and push unmatched closes. Extracting this into a named helper would reduce duplication, make the two-pass nature self-documenting, and ensure the matching logic can't diverge between passes in future edits.

**Suggestion:** Consider extracting a helper:
```ts
function balanceOpenCloseTags(
  sortedNodes: (HtmlElement | HtmlDanglingMarkerClose)[],
): (HtmlElement | HtmlDanglingMarkerClose)[] {
  const stack: (HtmlElement | HtmlDanglingMarkerClose)[] = [];
  for (const node of sortedNodes) {
    if (node.type === NodeTypes.HtmlElement) {
      stack.push(node);
    } else if (
      stack.length > 0 &&
      getName(node) === getName(stack.at(-1)!) &&
      stack.at(-1)!.type === NodeTypes.HtmlElement &&
      node.type === NodeTypes.HtmlDanglingMarkerClose
    ) {
      stack.pop();
    } else {
      stack.push(node);
    }
  }
  return stack;
}
```
Then both passes become a single call each.

packages/theme-check-common/src/checks/unclosed-html-element/index.ts
```diff
@@ -170,8 +177,44 @@ export const UnclosedHTMLElement: LiquidCheckDefinition = {
...
+              crossBalancedStack.at(-1)!.type === NodeTypes.HtmlElement &&
+              node.type === NodeTypes.HtmlDanglingMarkerClose
+            ) {
+              crossBalancedStack.pop();
+            } else {
+              crossBalancedStack.push(node);
+            }
+          }
+
+          // Build a set of nodes that were resolved by cross-identifier balancing
```
graygilmore: 🎨 **Code Style**: The comment says "Build a set of nodes that were resolved by cross-identifier balancing so we can skip reporting them" but `crossBalancedRemaining` is built from `crossBalancedStack` — nodes that survived the balancing pass, i.e., still **unmatched**. The code logic is correct: `if (!crossBalancedRemaining.has(node)) continue` skips resolved nodes and reports unresolved ones. But the comment says the inverse, which could easily mislead a future maintainer into introducing a regression by flipping the condition.

**Suggestion:** 
```suggestion
          // Build a set of nodes still unmatched after cross-identifier balancing.
          // Nodes NOT in this set were resolved and should not be reported.
```

packages/theme-check-common/src/utils/block.ts
```diff
@@ -143,6 +143,15 @@ export function isInvalidPresetBlock(
   }
 
   const isPrivateBlockType = blockNode.type.startsWith('_');
+  const hasNoRootBlocks = rootLevelThemeBlocks.length === 0;
```
graygilmore: 🐛 **Bug**: `rootLevelThemeBlocks.length === 0` is used as a proxy for "the schema has no root-level `blocks` declaration," but `rootLevelThemeBlocks` is already filtered to exclude `@app` entries. A schema with `blocks: [{"type": "@app"}]` explicitly declares a root-level blocks array and opts into block constraints, but the filtered list is empty — causing `hasNoRootBlocks` to be `true` and the early return to fire. This suppresses the requirement that private preset blocks be explicitly allowed at root level. The false negative extends to any call site that feeds through this helper, including nested validation paths.

**Suggestion:** Base this decision on the actual root schema shape rather than the filtered list. Thread a `hasRootLevelBlocksDeclaration` boolean (e.g. `Array.isArray(rootLevelBlocks)`) out of `getBlocks()`, pass it into `isInvalidPresetBlock`, and use it instead of `rootLevelThemeBlocks.length === 0`. Add a regression test for a schema whose root `blocks` array contains only `@app` while a preset references a `_private` block — that case should still require an explicit root allowlist.

packages/theme-check-common/src/utils/block.ts
```diff
@@ -143,6 +143,15 @@ export function isInvalidPresetBlock(
...
+
+  // Local blocks referenced in presets are valid when no root-level blocks
+  // are defined. They reference block files (blocks/_name.liquid) directly
+  // and their existence is validated separately.
+  if (isPrivateBlockType && hasNoRootBlocks) {
```
graygilmore: 💡 **Improvement**: The sibling function `isInvalidDefaultBlock` (lines 164-176) has analogous logic but does not include the new early return. When `rootLevelThemeBlocks` is empty and the block type is private (`_`-prefixed), `isInvalidDefaultBlock` computes `needsExplicitRootBlock = true` and returns `true` (invalid) — the same false-positive pattern this PR fixes for presets. If themes can use local blocks in `default` schema properties without root-level blocks, the same issue surfaces there. If this is intentionally scoped to presets only, a brief comment explaining why `default` is different would help future maintainers.

**Suggestion:** If the same pattern can appear in `default` schema configurations, apply the same early return to `isInvalidDefaultBlock`. If intentionally out of scope, add a comment to `isInvalidDefaultBlock` explaining why defaults have different semantics for local blocks.

packages/theme-check-common/src/checks/unclosed-html-element/index.ts
aswamy, line 1: I guess at runtime we still aren't sure if the this renders an unclosed html 🤔 But i guess we can't do anything about that.

```
{% if true %}
  <div class="grid-row">
{% endif %}

<div class="grid-item">{{ item }}</div>

{% if false %}
  </div>
{% endif %}
```

packages/theme-check-common/src/checks/valid-block-target/index.spec.ts
```diff
@@ -1310,6 +1310,75 @@ describe('Module: ValidBlockTarget', () => {
         expect(offenses[0].uri).to.equal(`file:///${path}/slideshow.liquid`);
       });
 
+      it(`should not report errors for local blocks in presets when no root-level blocks array exists (${path} bucket)`, async () => {
```
aswamy, line 1313: Is this valid? Like does it render if we actually run `shopify theme dev` on it?

aswamy (approved):
Tophatted everything! Couple of questions though 🤔 

---

Code completion

Image

Matching translation change for customer_accounts

Image


Unclosed HTML element

Image

Valid Block Target

Image

EvilGenius13 (approved):

packages/theme-check-common/src/checks/unclosed-html-element/index.ts
graygilmore: I know 😢, this is the downside of this

packages/theme-check-common/src/checks/valid-block-target/index.spec.ts
graygilmore: Oh like if the block files are empty?

graygilmore: I think it would complain that the files are empty but it shouldn't have any impact on the test for this. I'll double check though

graygilmore: I couldn't quite figure out what you were getting at here so feel free to poke me and I'll do a follow up

aswamy: I thought if you didn't have a root-level blocks array, things just break during rendering time. I didn't know there was an exception for private blocks 🤷

PR #1171, opened 2026-04-09
fix: use createRequire for node_modules resolution in VSCode extension

### LINKED ISSUE #1170: VSCode extension doesn't locate node_modules for custom checks package
**Describe the bug**

The VSCode extension doesn't look in node_modules when searching for extended packages. 

Works fine when running `shopify theme check` in the terminal. It also works when using an absolute path.

**Source**

```yml
extends:
  - 'theme-check:recommended'
  - '@grafikr/theme-check-extension/recommended.yml'
```

**Expected behaviour**
I expect the Theme Check extension to work in VSCode.

**Actual behaviour**
It sends this error:

```
[Error - 15.03.21] Request textDocument/documentLink failed.
  Message: Request textDocument/documentLink failed with message: Cannot find module '@grafikr/theme-check-extension/recommended.yml'
  Code: -32603 
```

**Debugging information**
- MacOS 15.7.4
- Shopify Liquid v3.11.1 (VSCode extension)

andershagbard:
## Summary

- Fixes the VSCode extension failing to resolve custom theme check packages (e.g. `@grafikr/theme-check-extension/recommended.yml`) from `node_modules`, while the CLI works fine.
- The root cause: the extension's language server is bundled by webpack, which replaces `require` with its own module system. webpack's `require.resolve` ignores the `paths` option and returns a numeric module ID instead of a real file path.
- The fix uses `createRequire` from `node:module` to create a proper Node.js require function that bypasses webpack's substitution. Node.js's native resolution already traverses ancestor `node_modules` directories, so the `getAncestorNodeModules` helper is no longer needed.

## Test plan

- [x] Confirm existing `read-yaml.spec.ts` tests pass (the node_module resolution test at line 105 exercises this exact code path)
- [x] Manually test by opening a Shopify theme project in VS Code that uses a custom `extends` package in `.theme-check.yml`, verify no "Cannot find module" errors

Closes #1170

### DIFF
--- a/packages/theme-check-node/src/config/resolve/read-yaml.ts
+++ b/packages/theme-check-node/src/config/resolve/read-yaml.ts
@@ -179,7 +179,7 @@ function resolvePath(
   // returns a numeric module ID instead of a file path. createRequire creates
   // a real Node.js require function that resolves from the given directory,
   // traversing ancestor node_modules automatically.
-  const req = createRequire(path.join(root, '__placeholder.js'));
+  const req = createRequire(path.join(root, '__placeholder'));
   return realpathSync(req.resolve(pathLike));
 }
 
--- a/packages/theme-check-node/src/config/resolve/read-yaml.ts
+++ b/packages/theme-check-node/src/config/resolve/read-yaml.ts
@@ -1,6 +1,7 @@
 import { CheckSettings, Modes, Severity } from '@shopify/theme-check-common';
 import { realpathSync } from 'node:fs';
 import fs from 'node:fs/promises';
+import { createRequire } from 'node:module';
 import path from 'node:path';
 import { parse } from 'yaml';
 import { AbsolutePath } from '../../temp';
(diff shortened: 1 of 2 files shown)


andershagbard:
I don't know how to test the VSCode extension. The build process seems to run in a loop for me

aswamy (approved):
This is great stuff. I was able to use your theme-check in VSCode extension, and using the cli binary in theme-tools

Image

Image

packages/theme-check-node/src/config/resolve/read-yaml.ts
```diff
@@ -172,7 +173,14 @@ function resolvePath(
...
   }
 
-  return realpathSync(require.resolve(pathLike, { paths: getAncestorNodeModules(root)! }));
+  // Use createRequire to get the real Node.js require rather than webpack's
+  // substitute. When the VSCode extension server is bundled, `require.resolve`
+  // becomes webpack's version which doesn't support the `paths` option and
+  // returns a numeric module ID instead of a file path. createRequire creates
+  // a real Node.js require function that resolves from the given directory,
+  // traversing ancestor node_modules automatically.
+  const req = createRequire(path.join(root, '__placeholder.js'));
```
aswamy: Does this need to be a `js` file? i think just `__placeholder` is fine. Not sure what the convention is though.

andershagbard:
```suggestion
  const req = createRequire(path.join(root, '__placeholder'));
```

andershagbard: I've updated it. Not sure what convention Shopify uses, if any, neither.

aswamy:
Thanks a lot @andershagbard for the fix. Merging this PR now 🙏. It should be part of the next patch for VSCode Extension. I think we _just_ [had one](https://github.com/Shopify/theme-tools/pull/1161) recently so look out for the next PR that looks like it when you want to see it released.

andershagbard:
Thanks for quick review and merge 😄

PR #1172, opened 2026-04-09
Add ValidScopedCSSClass theme check

aswamy:
## 🗒️ Summary

Adds a new theme check `ValidScopedCSSClass` that warns when a CSS class used in an HTML `class` attribute may be defined outside the scope of the current file.

This helps theme developers catch cases where a CSS class is used in a file but is only defined in a stylesheet tag that isn't accessible from that file's scope — indicating a potential dependency issue or misplaced style.

## ❓ What is considered "in scope"

A CSS class is considered in scope for a file if it is defined in any of the following:

- **Local `{% stylesheet %}` tag** — the file's own stylesheet
- **Ancestor `{% stylesheet %}` tags** — any file that directly renders this file (via `{% render %}`, `{% section %}`, etc.), traversed upward through the full ancestor chain. Only `direct` type references count (not `indirect` or `preset`)
- **Rendered snippet stylesheets** — any snippet the file renders via `{% render %}`, including that snippet's own snippet descendants (recursive)
- **Ancestor-rendered snippet stylesheets** — any snippet rendered by an ancestor file, including descendants
- **CSS asset files** — all `.css` files in the theme's `assets/` directory are globally in scope

## 👍 What is NOT reported

- **Classes not defined anywhere in the theme** — classes from external sources (CDNs, utility frameworks like Tailwind, inline JS injection, etc.) are silently ignored. The check only reports a class if it exists somewhere in the theme but isn't reachable from the current file's scope
- **Dynamic class attributes** — Liquid output tags (`{{ variable }}`) in class attributes are skipped since they can't be statically analyzed

#### Via VSCode Extension
- Checkout this branch
- Run `yarn install && yarn build`
- Press F5 to run the VScode extension in another window (might be `fn + F5` for some)
- Open a theme (or create one using Shopify CLI by running `npx shopify theme init`)
- Add a few CSS classes in stylesheet tags or `assets/*.css` files

#### Run cli-like environment
- Run the following command against the theme from this repo's root `node packages/theme-check-node/dist/cli.js [path-to-theme] > result` (replace [path-to-theme] with an actual path to theme)
  - See the theme check results from above appear in the `result` file it generated

### DIFF
--- a/packages/theme-check-common/src/checks/valid-scoped-css-class/index.ts
+++ b/packages/theme-check-common/src/checks/valid-scoped-css-class/index.ts
@@ -1,0 +1,189 @@
+import {
+  LiquidRawTag,
+  HtmlElement,
+  HtmlRawNode,
+  HtmlVoidElement,
+  HtmlSelfClosingElement,
+} from '@shopify/liquid-html-parser';
+import { LiquidCheckDefinition, Severity, SourceCodeType } from '../../types';
+import { AbstractFileSystem } from '../../AbstractFileSystem';
+import { recursiveReadDirectory } from '../../context-utils';
+import {
+  extractCSSClassNames,
+  collectUsedClasses,
+  collectUsedClassesFromSvg,
+} from '../../utils/styles';
+import {
+  getAncestorUris,
+  getAllSnippetDescendantUris,
+  getRenderedSnippetUris,
+} from '../../utils/traversal';
+
+/** Collect CSS classes from all .css files in the assets directory via getCSSClassesForURI. */
+async function getAssetCSSClasses(
+  fs: AbstractFileSystem,
+  toUri: (relativePath: string) => string,
+  getCSSClassesForURI: (uri: string) => Promise<Set<string>>,
+): Promise<Set<string>> {
+  const classes = new Set<string>();
+  try {
+    const assetsUri = toUri('assets');
+    const files = await fs.readDirectory(assetsUri);
+    const cssFileUris = files.filter(([uri]) => uri.endsWith('.css')).map(([uri]) => uri);
+    const results = await Promise.all(cssFileUris.map((uri) => getCSSClassesForURI(uri)));
+    for (const fileClasses of results) {
+      for (const cls of fileClasses) classes.add(cls);
+    }
+  } catch {
+    // assets directory might not exist
+  }
+  return classes;
+}
+
+/** Collect ALL CSS classes defined anywhere in the theme via getCSSClassesForURI. */
+async function getAllThemeCSSClasses(
+  fs: AbstractFileSystem,
+  toUri: (relativePath: string) => string,
+  getCSSClassesForURI: (uri: string) => Promise<Set<string>>,
+): Promise<Set<string>> {
+  const allClasses = new Set<string>();
+
+  // Collect from CSS asset files
+  const assetClasses = await getAssetCSSClasses(fs, toUri, getCSSClassesForURI);
+  for (const cls of assetClasses) allClasses.add(cls);
+
+  // Collect from all liquid files' stylesheet tags
+  try {
+    const rootUri = toUri('');
+    const liquidFiles = await recursiveReadDirectory(fs, rootUri, ([uri]) =>
+      uri.endsWith('.liquid'),
+    );
+    const results = await Promise.all(liquidFiles.map((uri) => getCSSClassesForURI(uri)));
+    for (const classes of results) {
+      for (const cls of classes) allClasses.add(cls);
+    }
+  } catch {
+    // root directory read failure
+  }
+
+  return allClasses;
+}
+
+export const ValidScopedCSSClass: LiquidCheckDefinition = {
+  meta: {
+    code: 'ValidScopedCSSClass',
+    name: 'Validates CSS classes used in class attributes are defined in an in-scope stylesheet',
+    docs: {
+      description:
+        'Reports CSS classes used in HTML class attributes that are not defined in any in-scope stylesheet tag or CSS asset file.',
+      recommended: true,
+      url: 'https://shopify.dev/docs/storefronts/themes/tools/theme-check/checks/valid-scoped-css-class',
+    },
+    type: SourceCodeType.LiquidHtml,
+    severity: Severity.WARNING,
+    schema: {},
+    targets: [],
+  },
+
+  create(context) {
+    const localCSSClasses = new Set<string>();
+    const usedClasses: { className: string; startIndex: number; endIndex: number }[] = [];
+
+    return {
+      async LiquidRawTag(node: LiquidRawTag) {
+        if (node.name === 'stylesheet') {
+          for (const cls of extractCSSClassNames(node.body.value)) {
+            localCSSClasses.add(cls);
+          }
+        }
+      },
+
+      async HtmlElement(node: HtmlElement) {
+        collectUsedClasses(node.attributes, usedClasses);
+      },
+
+      async HtmlVoidElement(node: HtmlVoidElement) {
+        collectUsedClasses(node.attributes, usedClasses);
+      },
+
+      async HtmlSelfClosingElement(node: HtmlSelfClosingElement) {
+        collectUsedClasses(node.attributes, usedClasses);
+      },
+
+      async HtmlRawNode(node: HtmlRawNode) {
+        if (node.name === 'svg') {
+          collectUsedClassesFromSvg(
+            node.attributes,
+            node.body.value,
+            node.body.position.start,
+            usedClasses,
+          );
+        }
+      },
+
+      async onCodePathEnd() {
+        if (usedClasses.length === 0) return;
+
+        const { getReferences, getDependencies, getCSSClassesForURI, fs, toUri } = context;
+        if (!getReferences || !getDependencies || !getCSSClassesForURI) return;
+
+        // Start with local CSS classes
+        const inScopeClasses = new Set(localCSSClasses);
+
+        // 1. Extract CSS classes from all .css files in the assets folder
+        const assetClasses = await getAssetCSSClasses(fs, toUri, getCSSClassesForURI);
+        for (const cls of assetClasses) {
+          inScopeClasses.add(cls);
+        }
+
+        // 2. Get all ancestors (following direct references upward)
+        const ancestors = await getAncestorUris(context.file.uri, getReferences);
+
+        // 3. Extract CSS classes from ancestor stylesheet tags
+        const ancestorClassResults = await Promise.all(
+          ancestors.map((uri) => getCSSClassesForURI(uri)),
+        );
+        for (const classes of ancestorClassResults) {
+          for (const cls of classes) inScopeClasses.add(cls);
+        }
+
+        // 4. Collect rendered snippets from this file and all ancestors
+        const filesToCheck = [context.file.uri, ...ancestors];
+        const allRenderedSnippetUris: string[] = [];
+        for (const fileUri of filesToCheck) {
+          const snippetUris = await getRenderedSnippetUris(fileUri, getDependencies);
+          allRenderedSnippetUris.push(...snippetUris);
+        }
+
+        // 5. BFS through snippet descendants to find all reachable snippets
+        const allSnippetUris = await getAllSnippetDescendantUris(
+          allRenderedSnippetUris,
+          getDependencies,
+        );
+
+        // 6. Extract CSS classes from all reachable snippet stylesheet tags
+        const snippetClassResults = await Promise.all(
+          allSnippetUris.map((uri) => getCSSClassesForURI(uri)),
+        );
+        for (const classes of snippetClassResults) {
+          for (const cls of classes) inScopeClasses.add(cls);
+        }
+
+        // 7. Collect ALL CSS classes defined anywhere in the theme
+        const allThemeClasses = await getAllThemeCSSClasses(fs, toUri, getCSSClassesForURI);
+
+        // 8. Only report classes that ARE defined somewhere in the theme but not in scope.
+        //    Classes not defined anywhere (e.g. from CDNs, utility frameworks) are silently ignored.
+        for (const { className, startIndex, endIndex } of usedClasses) {
+          if (allThemeClasses.has(className) && !inScopeClasses.has(className)) {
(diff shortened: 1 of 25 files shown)


packages/theme-graph/package.json
```diff
@@ -31,9 +31,11 @@
   "dependencies": {
     "@shopify/liquid-html-parser": "^2.9.2",
     "@shopify/theme-check-common": "^3.24.0",
-    "@shopify/theme-check-node": "^3.24.0",
     "acorn": "^8.14.1",
     "acorn-walk": "^8.3.4",
     "vscode-uri": "^3.0.7"
+  },
+  "devDependencies": {
+    "@shopify/theme-check-node": "^3.24.0"
```
aswamy, line 39: I had to move it to a dev-dependency so we can let theme-check-node be dependent on the theme-graph.

Why? Because CLI needs the theme-graph to be able to run the theme-check to detect cross-file CSS usage.

packages/theme-check-node/src/index.ts
```diff
@@ -135,6 +137,22 @@ export async function themeCheckRun(
...
+  const getSectionSchemaFn = async (name: string) => sectionSchemas.get(name)?.();
+  const getBlockSchemaFn = async (name: string) => blockSchemas.get(name)?.();
+
+  let themeGraph: ThemeGraph | undefined;
+  try {
+    themeGraph = await buildThemeGraph(config.rootUri, {
+      fs: NodeFileSystem,
+      getSectionSchema: getSectionSchemaFn as any,
+      getBlockSchema: getBlockSchemaFn as any,
+      getWebComponentDefinitionReference: () => undefined,
```
aswamy, line 173: I don't want to recreate the whole graph because we don't need it for this theme-check, but might silently error out if we add a theme-check that needs this in the future 🤔 

(but then again, we would silently fail if `getReferences` or `getDependencies` wasn't defined either so maybe not a HUGE deal 🤷)

packages/theme-language-server-common/src/server/startServer.ts
```diff
@@ -459,6 +459,9 @@ export function startServer(
         {
           globPattern: '**/*.liquid',
         },
+        {
+          globPattern: '**/assets/*.css',
```
aswamy, line 463: Need this or CSS files in assets are going to remain cached from initial load when running themecheck

packages/theme-check-common/src/checks/valid-scoped-css-class/index.ts
graygilmore, line 1: I tested a hilarious amount of edge cases and I think I found a few!

**Case sensitivity**

As far as I know HTML attributes aren't case sensitive so these should warn but currently don't:

```html
<div CLASS="out-of-scope-class"></div>
<div Class="out-of-scope-class"></div>
```

**SVGs**

```html
<svg class="out-of-scope-class">
  <rect class="out-of-scope-class" />
  <g class="out-of-scope-class"></g>
  <circle class="out-of-scope-class" />
  <path class="out-of-scope-class" />
</svg>
```

**Conditional classes**

I don't think we can actually do anything about these since we can't know the value of a dynamic class but I thought I'd show them just in case:

```liquid
<div class="scoped {% if true %}out-of-scope-class{% endif %}"></div>
```

`scoped` is checked correctly but inside the conditional isn't.

**Dynamic classes**

Similar to the above I know we can't validate dynamic classes but I wonder if this should still warn:

```liquid
{% assign butts = "butts" %}
<div class="out-of-scope-class{{ butts }} class1 class2"></div>
```

I think Claude found a suggestion for this one.

**Orphaned snippets**

I don't know if we actually want to cover this case but if you have two orphaned snippets that create their own little tree you'll get a false positive warning when running theme check (but you won't get the warning when viewing the file in the editor it seems?):

```liquid
<!-- snippets/orphan.liquid -->
{% render 'orphan-styles' %}
<div class="orphan-button">Click me</div>
```

```liquid
<!-- snippets/orphan-styles.liquid -->
{% stylesheet %}
  .orphan-button {
    color: red;
  }
{% endstylesheet %}
```

packages/theme-check-node/src/index.ts
```diff
@@ -134,6 +138,43 @@ export async function themeCheckRun(
...
+  // Build theme graph for cross-file checks (e.g. ValidScopedCSSClass)
+  const getSectionSchemaFn = async (name: string) => sectionSchemas.get(name)?.();
+  const getBlockSchemaFn = async (name: string) => blockSchemas.get(name)?.();
+
+  let themeGraph: ThemeGraph | undefined;
+  try {
+    themeGraph = await buildThemeGraph(config.rootUri, {
+      fs: NodeFileSystem,
+      getSectionSchema: getSectionSchemaFn as any,
+      getBlockSchema: getBlockSchemaFn as any,
```
graygilmore: Are we able to avoid the `as any`?

packages/theme-graph/package.json
graygilmore: I think this will blow up the theme-graph binary: https://github.com/Shopify/theme-tools/blob/main/packages/theme-graph/bin/theme-graph#L10

packages/theme-check-common/src/checks/valid-scoped-css-class/index.ts
```diff
@@ -0,0 +1,173 @@
...
+    schema: {},
+    targets: [],
+  },
+
+  create(context) {
+    const localCSSClasses = new Set<string>();
+    const usedClasses: { className: string; startIndex: number; endIndex: number }[] = [];
+
+    return {
+      async LiquidRawTag(node: LiquidRawTag) {
```
graygilmore, line 93: 🎨 **Code Style**: The `LiquidRawTag`, `HtmlElement`, `HtmlVoidElement`, and `HtmlSelfClosingElement` visitors are declared `async` but perform only synchronous operations — Set/Array mutations with no `await`. Other checks in the codebase use non-async visitors for synchronous work. While functionally harmless, this adds unnecessary microtask overhead per node visit.

**Suggestion:** Remove `async` from the synchronous visitors. The `onCodePathEnd` visitor correctly remains `async` since it uses `await`.

```suggestion
      LiquidRawTag(node: LiquidRawTag) {
```

packages/theme-check-common/src/checks/valid-scoped-css-class/index.ts
```diff
@@ -0,0 +1,173 @@
...
+        // 6. Extract CSS classes from all reachable snippet stylesheet tags
+        const snippetClassResults = await Promise.all(
+          allSnippetUris.map((uri) => getCSSClassesForURI(uri)),
+        );
+        for (const classes of snippetClassResults) {
+          for (const cls of classes) inScopeClasses.add(cls);
+        }
+
+        // 7. Collect ALL CSS classes defined anywhere in the theme
+        const allThemeClasses = await getAllThemeCSSClasses(fs, toUri, getCSSClassesForURI);
```
graygilmore, line 173: ⚡ **Performance**: `getAllThemeCSSClasses` is called inside `onCodePathEnd`, which runs once per checked file. Each invocation does a full recursive directory traversal and reads+parses every `.liquid` and `.css` file in the theme. For a theme with N files using class attributes and M total theme files, this means N full-theme scans. Even in the CLI path where `getCSSClassesForURI` is memoized per-URI, the directory listing itself repeats on every invocation. The same applies to `getAssetCSSClasses`.

The set of all theme CSS classes is a theme-wide constant. The established pattern in this codebase (e.g., `sectionSchemas`, `blockSchemas`, `docDefinitions`) is to precompute expensive run-scoped data once and share it through injected dependencies. Since this check is enabled by default, the cost scales with every theme that uses class attributes.

**Suggestion:** Cache the theme-wide class index once per lint run and share it through the injected dependencies, following the pattern used by other expensive run-scoped lookups in the codebase. For example, add `getAllThemeCSSClasses?: () => Promise<Set<string>>` to the `Dependencies` interface and have the host compute and cache it once. Alternatively, memoize the result within the check's `create()` closure.

packages/theme-check-common/src/utils/styles.ts
```diff
@@ -0,0 +1,155 @@
...
+
+/** Extract used class names from an HTML element's class attribute. */
+export function collectUsedClasses(
+  attributes: AttributeNode[],
+  usedClasses: { className: string; startIndex: number; endIndex: number }[],
+): void {
+  for (const attr of attributes) {
+    if (!isValuedHtmlAttribute(attr) || !isAttr(attr, 'class')) continue;
+
+    for (const valueNode of attr.value) {
```
graygilmore: 🐛 **Bug**: The class extraction helper tokenizes each `TextNode` independently, but the parser represents a class attribute as alternating text and Liquid nodes. A value like `class="btn-{{ size }}"` contributes `btn-` as if it were a complete static class. This is inconsistent with the intended behavior of skipping dynamic class usage — the check correctly skips fully dynamic classes but still emits partial fragments around Liquid interpolation, which can produce false positives for common dynamic-class patterns.

**Suggestion:** Consider accumulating class tokens across the full attribute value and discarding the current token whenever a non-text Liquid node interrupts it. That keeps `class="static {{ dynamic }}"` valid for `static` while avoiding false positives for partial tokens like `btn-` from `btn-{{ size }}`.

packages/theme-check-common/src/checks/valid-scoped-css-class/index.ts
aswamy: We can't do this because it breaks the type of the object being returned - each method needs to return `Promise<void>` - we could just do that for each method, but that's the same overhead of adding `async` to methods 🤷

stephanie-shopify (commented):
Ran a top hat, found one possible case to look at
Grandparent chain:
 Layout → section → snippet: When a layout file defines a CSS class and renders a section, which in turn renders a snippet that uses that class, the check warns that the class is out of scope.                                                                 
                                                                                                                                                                                                                                                                             
 The setup:                                                                                                                                                                                                                                                                  
 - layout/theme.liquid has `{% stylesheet %} .layout-class { ... } {% endstylesheet %}` and `{{ content_for_layout }}`                                                                                                                                                           
 - sections/section.liquid renders ` {% render 'deep-snippet' %} `                                                                                                                                                                                                              
 - snippets/deep-snippet.liquid uses `<div class="layout-class"> `

aswamy:
> Ran a top hat, found one possible case to look at Grandparent chain: Layout → section → snippet: When a layout file defines a CSS class and renders a section, which in turn renders a snippet that uses that class, the check warns that the class is out of scope.
> 
> The setup:
> 
> * layout/theme.liquid has `{% stylesheet %} .layout-class { ... } {% endstylesheet %}` and `{{ content_for_layout }}`
> * sections/section.liquid renders `{% render 'deep-snippet' %}`
> * snippets/deep-snippet.liquid uses `<div class="layout-class"> `

@stephanie-shopify Layout's are not considered *direct* parents of sections. Their relation is sort of just implied because of `content_for_layout`.

You'll see in the side bar, that the parent of `sections/hello-world.liquid` isn't a layout file, but the template json. I assume that's true for your situation as well. There are no _true_ parents for those json files.

image

packages/theme-check-common/src/checks/valid-scoped-css-class/index.ts
aswamy: The issue isn't really a problem on VSCode Extension since we cache the entire fs. But on CLI, there is partial concern to what you're saying because even though `getCSSClassesForURI` is memoed, we are iterating over the files individually using `recursiveReadDirectory`. We could have just made `getAllThemeCSSClasses` like you suggested which would do it in bulk, but i feel like this breaks the current pattern we are going for [here](https://github.com/Shopify/theme-tools/blob/f5b8d0712d7cdf9d7e96acfb1be3e2c1d8ee8b3f/packages/theme-check-common/src/types.ts#L347-L395). We always pass a URI for a specific file and it returns its memoized content.

Ill try it out in a subsequent PR though because i still think we take around 18sec of wall-clock time and any gains is nice.

packages/theme-graph/package.json
aswamy: Im going to use `theme-check-common` instead of `theme-check-node` in the binary and move any functions over that doesn't exist. I dont think anyone uses this raw binary anyway.

stephanie-shopify (approved):

aswamy:
@graygilmore Thanks for the thorough tophat and review Gray!

I've completed the following fixes:
✅ Case sensitivity
✅ SVGs - a little tougher than i thought because we consider the SVG tag a "raw" node so i had to add a helper to parse it
✅ Dynamic Classes - we do make sure we distinguish between `btn-{{ size }}` and `btn {{ size }}`. The first one the entire class is considered dynamic and discarded. The second one, we only check `btn` to see if it appears in other stylesheet tag


I think "Conditional classes" and "Orphaned snippets" might be a little bit too niche. We can always update it once we get feedback on this.

> if you have two orphaned snippets that create their own little tree you'll get a false positive warning when running theme check (but you won't get the warning when viewing the file in the editor it seems?):

That's a weird bug. I noticed that too. Happens in CLI but not VSCode extension? I wonder if it's a theme-graph issue for orphaned files because the orphan-styles file doesn't have any references in the side panel on VSCode. It should at least have `orphan`. Either way, orphaned files isn't too much of a concern for us.

packages/theme-check-common/src/utils/styles.ts
```diff
@@ -0,0 +1,238 @@
...
+
+      if (tokens.length === 0) continue;
+
+      // Drop first token if it touches the preceding Liquid node (no leading whitespace)
+      if (prevIsLiquid && !value.startsWith(' ') && !value.startsWith('\n')) {
+        tokens.shift();
+      }
+
+      // Drop last token if it touches the following Liquid node (no trailing whitespace)
+      if (nextIsLiquid && !value.endsWith(' ') && !value.endsWith('\n')) {
```
graygilmore: Last edge case! These two checks only check for new lines and spaces but HTML spec considers more than those as whitespace (tabs for example). Might want to use a regex to match all ASCII whitespace (I haven't validated the below):

```js
// startswith
if (prevIsLiquid && !/^\s/.test(value)) {}

// endswith
if (nextIsLiquid && !/\s$/.test(value)) {}
```

graygilmore (approved):
One more small edge case

PR #1176, opened 2026-04-16
liquid-html-parser: fix 'occured' -> 'occurred' comment typos in errors.ts

SAY-5:
Two inline comments in `packages/liquid-html-parser/src/errors.ts` (lines 22 and 61) explaining where the babel parser prints the error read `the parsing error occured`. Doc-only TS change.


graygilmore:
Thanks for the contribution, but we don't accept automated mass-submitted typo fixes.

Issue #1179, opened 2026-04-17
Getting code actions from 'Shopify Liquid'" causes severe save lag in v3.11.2


yankeyhotel:
### Description

After updating to v3.11.2, saving any file causes VS Code to hang on a "Getting code actions from 'Shopify Liquid'" dialog, making the editor extremely slow. Rolling back to the previous version resolves the issue immediately.

### Steps to reproduce

1. Install or update Shopify Liquid extension to v3.11.2
2. Open any `.liquid` file
3. Make a small edit and save (Ctrl+S / Cmd+S)
4. Observe the "Getting code actions from 'Shopify Liquid'" dialog hanging for an unusually long time

### Expected behavior

File saves quickly with no noticeable delay.

aswamy:
Hey @yankeyhotel. Can you see if disabling the `ValidScopedCSSClass` theme-check makes it fast again?

You can do that by adding the following chunk inside of your .theme-check.yml file (if you don't have one, just create it using `shopify theme check --init`

```
ValidScopedCSSClass:
  enabled: false
  severity: 1
```

aswamy:
I've temporarily disabled the theme-check for everyone on VSCode using the latest extension. Will re-enable it once the performance issue is resolved. Thanks for your for report @yankeyhotel!

PR #1181, opened 2026-04-17
perf: reduce ValidScopedCSSClass save latency on large themes

aswamy:
## Summary

Fixes the save-lag reported in #1179 and re-enables `ValidScopedCSSClass` in the recommended config (reversing the temporary opt-out from #1180).

## What was happening

The check needed to know every CSS class declared anywhere in the theme to avoid false positives on CDN/utility classes. The language server was rebuilding that information from scratch on every save — re-reading and re-parsing every Liquid file with a stylesheet tag, even though the language server already holds the parsed syntax tree for every file in memory.

## What this PR changes

Two architectural changes, no change to the check's output or public API:

1. **CSS class extraction is now a memoized derivation on the language server's document model** — alongside schema and doc-definition extraction, following the same pattern. When a document changes, the model replaces the cached extraction automatically, so invalidation is implicit. The expensive parse work is shared across every consumer of a file's CSS classes and across every save within a session.

2. **Early exit for files with no stylesheet tag** — cheap source-level scan avoids invoking the syntax-tree parser for files that cannot declare CSS classes.

## Impact (measured on a 240-file theme)

- First save after opening an editor: ~4.5s
- Subsequent saves: ~4.5s → <100ms

## Test plan
- [ ] Install branch build in VS Code on a large theme (horizon should work)
- [ ] First save latency lower than `main`; subsequent saves sub-second (see logs)
- [ ] Editing a stylesheet tag, saving, and opening an ancestor shows the new class immediately

### DIFF
--- a/packages/theme-check-common/src/utils/styles.ts
+++ b/packages/theme-check-common/src/utils/styles.ts
@@ -2,6 +2,7 @@ import {
   NodeTypes,
   TextNode,
   LiquidRawTag,
+  LiquidHtmlNode,
   toLiquidHtmlAST,
   AttributeNode,
 } from '@shopify/liquid-html-parser';
@@ -9,7 +10,6 @@ import safeParse from 'postcss-safe-parser';
 import selectorParser from 'postcss-selector-parser';
 import { SourceCodeType } from '../types';
 import { AbstractFileSystem } from '../AbstractFileSystem';
-import { recursiveReadDirectory } from '../context-utils';
 import { isValuedHtmlAttribute, ValuedHtmlAttribute } from '../checks/utils';
 import { visit } from '../visitor';
 
@@ -33,6 +33,24 @@ export function extractCSSClassNames(css: string): Set<string> {
   return classNames;
 }
 
+/** Extract CSS class names from a Liquid AST's {% stylesheet %} tags. */
+export function extractCSSClassesFromLiquidAST(ast: LiquidHtmlNode): Set<string> {
+  const classes = new Set<string>();
+  const cssStrings = visit<SourceCodeType.LiquidHtml, string>(ast, {
+    LiquidRawTag(node: LiquidRawTag) {
+      if (node.name === 'stylesheet') {
+        return node.body.value;
+      }
+    },
+  });
+  for (const css of cssStrings) {
+    for (const cls of extractCSSClassNames(css)) {
+      classes.add(cls);
+    }
+  }
+  return classes;
+}
+
 /** Read a Liquid file and extract CSS class names from its {% stylesheet %} tags. */
 export async function extractCSSClassesFromLiquidUri(
   uri: string,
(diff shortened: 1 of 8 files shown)


packages/theme-language-server-common/src/diagnostics/runChecks.ts
```diff
@@ -46,23 +46,39 @@ export function makeRunChecks(
     //  theme1/snippets/b.liquid
     //
     // then we recheck theme1
+    const runStart = Date.now();
+    console.error(
+      `[theme-check] runChecks start — ${triggerURIs.length} trigger URI(s): ${triggerURIs
+        .map((u) => u.split('/').pop())
+        .join(', ')}`,
+    );
```
aswamy: I was originally going to remove it after profiling it, but found it useful for debugging. I am okay with keeping it if others are cool about it

packages/theme-check-common/src/utils/styles.ts
```diff
@@ -41,19 +60,15 @@ export async function extractCSSClassesFromLiquidUri(
   const classes = new Set<string>();
   try {
     const source = await fs.readFile(uri);
+    // Most liquid files have no {% stylesheet %} tag — skip the AST parse
+    // entirely when the tag isn't present. Saves ~20ms/file on large themes.
+    if (!/\{%-?\s+stylesheet/.test(source)) {
+      return classes;
```
aswamy: Bit clunky, but avoid having to `toLiquidHtmlAST` (expensive function) whenever there is no stylesheet tag. But since we have the Document Manager do it most of the time on VSCode, we can probably remove this 🤔 . Let me remove and see what's up. Ill resolve the comment if i don't end up changing anything.

aswamy: So in VSCode, since we `preload` the theme, this is rarely hit so it's not needed. BUT on CLI, this does give _some_ benefit since we don't have Document Manager to preload and parse the AST.

packages/theme-check-common/src/utils/styles.ts
```diff
@@ -74,60 +88,6 @@ export async function extractCSSClassesFromAssetUri(
...
-    for (const classes of results) {
-      for (const cls of classes) allClasses.add(cls);
-    }
-  } catch {
-    // root directory read failure
-  }
-
-  return allClasses;
-}
-
```
tgeselle, line 130: I think there's a mismatch between the deletion of these functions and the changeset. 
Apparently removing a public API in a patch release is not recommended. 

I think you should either restore these as thin/deprecated wrappers for compatibility, or explicitly treat this as a breaking release and document the migration path.

packages/theme-check-common/src/utils/styles.ts
```diff
@@ -41,19 +59,15 @@ export async function extractCSSClassesFromLiquidUri(
   const classes = new Set<string>();
   try {
     const source = await fs.readFile(uri);
+    // Most liquid files have no {% stylesheet %} tag — skip the AST parse
+    // entirely when the tag isn't present. Saves ~20ms/file on large themes.
+    if (!/\{%-?\s+stylesheet/.test(source)) {
```
tgeselle: This requires at least one whitespace character after `{%/{%-`, but Liquid allows zero or more spaces for raw tags. That means `{%stylesheet%}` and `{%-stylesheet-%}` will be treated as "no stylesheet tag", so class extraction silently returns an empty set

I think you can do this instead:
```javascript
if (!/\{%-?\s*stylesheet\b/.test(source))
```

packages/theme-language-server-common/src/diagnostics/runChecks.ts
tgeselle: I'm not too familiar with the repo but usually leaving console stuff in the code is not recommended. I'll leave it to other to know if it's ok to keep them or not.

stephanie-shopify (approved):

PR #1184, opened 2026-04-18
fix(theme-language-server-common): preserve element type on array-returning filters

### LINKED ISSUE #1086: Improve doc typing when converting to array
Hello,

Liquid currently lacks a lot in the array manipulation area, but there are a lot of tricks to workaround. One common trick is to create an array from a variable using the sort filter:

```liquid
for media in product.media
  assign media_as_array = media | sort
endfor
```

Which then allows to concat it to another array. The issue is that currently the type is lost, so while media is properly typed as "media", after the sort operation it becomes untyped[]

Until we can have better array functions, could the typing be improved so that the type is preserved?

Thanks!

Jp-fix:

Fixes #1086.

When a typed value is piped through an array-returning filter like `sort`, the language server was showing `untyped[]` on hover instead of preserving the element type. Example from the issue:

```liquid
{% for media in product.media %}
  {% assign media_as_array = media | sort %}
  {# hover on `media_as_array` used to show `untyped[]` — now shows `media[]` #}
{% endfor %}
```

The filter docs declare `array_value: "untyped"` for these filters, so `docsetEntryReturnType` was always handing back `untyped[]`. Instead of changing the docs (they're pulled from `Shopify/theme-liquid-docs`), I added a small pass-through list in `TypeSystem.ts` covering `sort`, `sort_natural`, `reverse`, `uniq`, `compact` and `where`, and switched the filter chain from "only inspect the last filter" to a left-to-right reduce so chains like `x | sort | reverse` thread the element type correctly.

Scalar inputs (`media | sort` in the issue) are wrapped into an array of the same type, matching Liquid's runtime coercion.

- \`map\` is out of scope here — it would need to resolve a property lookup on the element type. Worth a separate PR.
- \`concat\` is also out of scope: the result type would be a union of both arrays' element types, which the type system doesn't represent today.

## Before you deploy

- [x] I included a patch bump \`changeset\`

### DIFF
--- a/.changeset/preserve-element-type-on-array-filters.md
+++ b/.changeset/preserve-element-type-on-array-filters.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-language-server-common': patch
+---
+
+Preserve the element type when piping a typed value through `sort`, `sort_natural`, `reverse`, `uniq`, `compact` or `where`. Previously these filters collapsed the type to `untyped[]` because their `array_value` is declared as `untyped` in the filter docs; they are now treated as pass-through in the type system. Fixes #1086.
--- a/packages/theme-language-server-common/src/TypeSystem.ts
+++ b/packages/theme-language-server-common/src/TypeSystem.ts
@@ -604,6 +604,29 @@ function resolveTypeRangeType(
   }
 }
 
+/**
+ * Filters whose return type is an array of the same element type as their
+ * input. The filter docs in `filters.json` declare `array_value: "untyped"`
+ * for these, which would make every pipeline collapse to `untyped[]`. We
+ * special-case them here to preserve the element type of the input.
+ *
+ * Intentionally excluded:
+ *  - `map` — projects a property, needs lookup resolution on the element type.
+ *  - `concat` — merges two arrays, element type would be a union.
+ */
+const ARRAY_PASSTHROUGH_FILTERS: ReadonlySet<string> = new Set([
+  'sort',
+  'sort_natural',
+  'reverse',
+  'uniq',
+  'compact',
+  'where',
+]);
+
+function elementTypeOf(t: PseudoType | ArrayType): PseudoType {
+  return isArrayType(t) ? t.valueType : t;
+}
+
 function inferType(
   thing: Identifier | ComplexLiquidExpression | LiquidVariable | AssignMarkup,
   symbolsTable: SymbolsTable,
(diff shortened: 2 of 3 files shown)


Jp-fix:
I have signed the CLA.

.changeset/preserve-element-type-on-array-filters.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/theme-language-server-common': patch
+---
+
+Preserve the element type when piping a typed value through `sort`, `sort_natural`, `reverse`, `uniq`, `compact` or `where`. Previously these filters collapsed the type to `untyped[]` because their `array_value` is declared as `untyped` in the filter docs; they are now treated as pass-through in the type system. Fixes #1086.
```
aswamy, line 5: Keep the changeset short and sweet. No implementation details or ticket numbers.

packages/theme-language-server-common/src/TypeSystem.ts
```diff
@@ -604,6 +604,29 @@ function resolveTypeRangeType(
...
+ *  - `concat` — merges two arrays, element type would be a union.
+ */
+const ARRAY_PASSTHROUGH_FILTERS: ReadonlySet<string> = new Set([
+  'sort',
+  'sort_natural',
+  'reverse',
+  'uniq',
+  'compact',
+  'where',
+]);
```
aswamy, line 624: Instead of having a static allow-list, could we invert this and exclude all array filters that _isnt_ map or concat.

packages/theme-language-server-common/src/TypeSystem.ts
```diff
@@ -604,6 +604,29 @@ function resolveTypeRangeType(
...
+
+function elementTypeOf(t: PseudoType | ArrayType): PseudoType {
```
aswamy, line 626: Have a comment that mentions that non-array inputs are treated as a single-element array of that type
```suggestion
/* Returns the element type of an array, or the input itself if it is a scalar. */
function elementTypeOf(t: PseudoType | ArrayType): PseudoType {
```

aswamy (commented):
Thanks for the contributions Jp!

I've left a few suggestions in the PR. Also, please follow the PR template for your PR descriptions

https://github.com/Shopify/theme-tools/blob/main/.github/PULL_REQUEST_TEMPLATE.md

PR #1189, opened 2026-04-21
fix: allow theme-check-disable-next-line to suppress checks inside doc tags

### LINKED ISSUE #945: Theme check disable next line does not work for `doc` tag
**Describe the bug**
- You can add `{% # theme-check-disable-next-line %}` to disable the theme check for the next liquid statement in the code
- If you do this inside of `doc` tag, it is treated as part of the LiquidDoc description
- Disabling theme-check at the doc-tag level does not disable checks inside the doc tag

Image

**Source**
```liquid
{% # theme-check-disable-next-line UnusedAssign %}
{% doc %}
  @param {product[]} prods - products
{% enddoc %}
```

**Expected behaviour**
- If we can't disable the theme-check from inside the `doc` tag, we should at least be able to disable the theme check inside the doc tag

**Actual behaviour**
- can't disable theme checks for next line when it's inside of `doc` tag

SinhSinhAn:

Fixes the case where `{% # theme-check-disable-next-line %}` placed before a `{% doc %}` tag does not actually suppress the checks that run against the doc tag's contents.

Today, when a snippet has a duplicate `@param` you intentionally want to keep, the only workaround is wrapping the whole tag in a disable/enable pair:

```liquid
{% # theme-check-disable-next-line UniqueDocParamNames %}
{% doc %}
  @param prods - products
  @param prods - intentional duplicate
{% enddoc %}
```

This is verbose. The single-line `disable-next-line` should just work.

**Root cause:** `findNextLinePosition` in `disabled-checks/index.ts` has a guard for block tags. When the next node has a `blockStartPosition` (like `{% if %}`, `{% for %}`, `{% doc %}`), it returns only `blockStartPosition` instead of the full `position`. That guard is there so `disable-next-line` does not accidentally suppress checks on nested children of control flow tags. But doc tag checks (`UniqueDocParamNames`, `ValidDocParamTypes`, `ReservedDocParamNames`) report offenses at positions *inside* the doc body, so they fall outside `blockStartPosition` and are never suppressed.

**Fix:** when the next node is a `LiquidRawTag` with `name === 'doc'`, return the full `position` span instead of `blockStartPosition`. Safe because doc tag bodies are parsed by a separate grammar (`LiquidDocGrammar`) that only produces doc-specific nodes (`@param`, `@description`, `@example`), so there is no risk of accidentally silencing unrelated checks on nested control flow.

The change is a 4-line conditional in `findNextLinePosition`.

Three new test cases were added to `disabled-checks/index.spec.ts`:
- `disable-next-line UniqueDocParamNames` before a doc tag with a duplicate param suppresses the offense
- `disable-next-line` (no specific check) suppresses all checks inside the doc tag
- Without the disable comment, the offense still fires (regression guard)

All 18 tests pass (15 existing + 3 new).

Closes #945

None planned. The narrow scope here (only `{% doc %}`, only when matched as the "next node") avoids the larger question of how to support `disable-next-line` *inside* a doc tag body, which is a separate problem because `LiquidDocGrammar` does not parse Liquid tags at all, so the comment would never appear as a `LiquidTag` node in the first place. That's out of scope for this PR.

## What did you learn?

The `blockStartPosition` guard in `findNextLinePosition` is correct for control flow tags (`{% if %}`, `{% for %}`) and the wrong default for "leaf-ish" block tags like `{% doc %}` whose children are themselves the thing being checked. Worth keeping in mind if other block tags in the future also report offenses on their inner nodes rather than on their wrappers.

## Tophatting

In a theme with a snippet that intentionally duplicates a `@param` for documentation purposes:

```liquid
{% # theme-check-disable-next-line UniqueDocParamNames %}
{% doc %}
  @param prods - products
  @param prods - intentional duplicate
{% enddoc %}
```

- Before this PR: `UniqueDocParamNames` still reports an offense.
- After this PR: the offense is suppressed.

Also confirm that `disable-next-line` before a regular block (e.g. `{% if %}`) still only disables the opening line, not the children inside.

## Before you deploy

- [x] I included a patch bump \`changeset\` (\`@shopify/theme-check-common: patch\`)

### DIFF
--- a/.changeset/fix-disable-doc-tag.md
+++ b/.changeset/fix-disable-doc-tag.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-check-common': patch
+---
+
+Fix theme-check-disable-next-line not working for checks inside doc tags
--- a/packages/theme-check-common/src/disabled-checks/index.spec.ts
+++ b/packages/theme-check-common/src/disabled-checks/index.spec.ts
@@ -3,6 +3,7 @@ import { Offense } from '..';
 import { describe, it, expect } from 'vitest';
 import { LiquidFilter, RenderMarkup } from './test-checks';
 import { UndefinedObject } from '../checks/undefined-object';
+import { UniqueDocParamNames } from '../checks/unique-doc-param-names';
 
 const commentTypes = [
   (text: string) => `{% # ${text} %}`,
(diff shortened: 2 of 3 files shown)


aswamy (commented):
Hey @SinhSinhAn Thanks for another PR. The code looks fine.

Could you ensure your PR descriptions for all your PRs follow the PR template recommended by this repo?

https://github.com/Shopify/theme-tools/blob/main/.github/PULL_REQUEST_TEMPLATE.md

SinhSinhAn:
Of course, I can make edits after this

aswamy (approved):

PR #1193, opened 2026-04-23
fix: update section type in config/settings_data.json on section rename

### LINKED ISSUE #713: Section rename handling in config/settings_data.json
**Describe the bug**
Looks like there's an undocumented behaviour of config/settings_data.json we'd need to think about:
https://github.com/Shopify/dawn/blob/main/config/settings_data.json#L173-L185

- Section type should be renamed
- Does that thing accept blocks? Do we rename blocks in there too?

reza6ix:
## Problem
Fixes #713. When a section file is renamed in a Shopify theme, the `type` attribute
inside `config/settings_data.json` presets was not updated, leaving the file pointing
at the old section name
Template files, section groups, and `{% section %}` tags were already handled by
`SectionRenameHandler`; `settings_data.json` was the only missing location.

## Changes
- `src/utils/uri.ts` — adds `isSettingsData` URI predicate
- `src/renamed/handlers/utils.ts` — adds `SettingsData` interfaces and `isValidSettingsData` type guard
- `src/renamed/handlers/SectionRenameHandler.ts` — adds `getSettingsDataChanges` method
- `src/renamed/handlers/SectionRenameHandler.spec.ts` — 4 new tests

Only the `type` attribute value is updated; section keys and all other fields are untouched.

## Test plan
- [ ] `yarn test` in `packages/theme-check-common` — 946 tests pass
- [ ] `yarn test` in `packages/theme-language-server-common` — 437 tests pass

### DIFF
--- a/packages/theme-check-common/vitest.config.ts
+++ b/packages/theme-check-common/vitest.config.ts
@@ -1,0 +1,16 @@
+import { defineConfig } from 'vitest/config';
+import { configDefaults } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    exclude: [...configDefaults.exclude],
+    pool: 'forks',
+    poolOptions: {
+      forks: {
+        singleFork: true,
+        isolate: true,
+      },
+    },
+    setupFiles: ['./src/test/test-setup.ts'],
+  },
+});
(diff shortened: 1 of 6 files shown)


reza6ix:
I signed the CLA

PR #1208, opened 2026-05-21
Migrate theme-tools to pnpm

graygilmore:
## Summary

Migrates theme-tools from Yarn to pnpm so `dev up` can run again with minimum dependency age enforcement enabled.

This updates the dev environment, package manager metadata, workspace config, lockfile, package scripts, CI/release workflows, VS Code tasks, and the public setup docs.

## Why

`dev up` now blocks Yarn projects because the supply-chain protection we use for JavaScript dependencies does not support Yarn. pnpm is supported, so this moves the repo onto pnpm instead of opting out of that protection.

## A few non-obvious changes

Most of this is mechanical `yarn` -> `pnpm`, but a few things are worth calling out:

- pnpm does not let packages silently borrow dependencies from another package's `node_modules`. A few dependencies are now declared where they were already imported directly.
- `pnpm.onlyBuiltDependencies` is pnpm's allowlist for dependency install scripts. This keeps the list small and explicit instead of making every developer or CI run `pnpm approve-builds` interactively.
- GitHub Actions now installs pnpm with `pnpm/action-setup` before `actions/setup-node`. `setup-node` runs pnpm while restoring the pnpm cache, so `corepack enable` in a later step was too late and caused CI to fail before install started.
- Workspace builds run with `--workspace-concurrency=1` because a few packages expect generated files from earlier packages. This keeps the old build order behavior rather than racing those generated files.
- The VS Code webpack config and prettier-plugin test setup now resolve Prettier through Node resolution instead of assuming hoisted `node_modules/prettier2` or `node_modules/prettier3` paths. pnpm stores package aliases differently, so the old paths were too specific to Yarn.
- The TS config now targets `es2022` because the codebase already uses `Array.prototype.at`. The language server common package also includes browser globals because it uses things like `URL`, timers, and `console`.
- The `syntaxes` submodule install uses `--ignore-workspace` so pnpm treats it as its own package instead of trying to fold it into the root workspace.

I intentionally left `scripts/dev-admin-build.sh` and `scripts/dev-admin-init.sh` on Yarn. Those need a separate pass because the linking behavior is a little more involved than a straight command swap.

## Testing

- GitHub Actions CI is green on this branch
- `dev up`
- `pnpm install --frozen-lockfile --config.confirmModulesPurge=false`
- `pnpm build`
- `pnpm test --run`
- Parsed the updated workflow YAML files with the repo's `yaml` dependency
- `git diff --check`

I did not run branch-scoped `dev test -i` / `dev style -i` / `dev check -i` because this repo does not define those project commands in `dev.yml`; the pnpm scripts above are the relevant checks here.

## 🎩 Tophatting

1. Pull this branch.
2. Run `dev up` from the repo root.
3. Confirm it no longer fails with the deprecated package manager warning.
4. Run `pnpm install --frozen-lockfile`.
5. Run `pnpm build` and `pnpm test --run`.


EvilGenius13 (approved):
🎩 went well. `dev up` worked as well as running the development server and trying out some changes.

Couple notes:
There is a `CONTRIBUTING.md` file at the root that also needs to be changed.
There are some inner package README's and a few `.yml/.sh` files that could also be updated.

packages/lang-jsonc/src/highlight.ts
```diff
@@ -1,6 +1,7 @@
+import type { NodePropSource } from '@lezer/common';
 import { styleTags, tags as t } from '@lezer/highlight';
 
-export const jsonHighlighting = styleTags({
+export const jsonHighlighting: NodePropSource = styleTags({
```
aswamy, line 4: probably don't need this change but doesn't hurt 🤷

packages/theme-language-server-common/src/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "outDir": "../dist",
     "rootDir": ".",
     "tsBuildInfoFile": "../dist/tsconfig.tsbuildinfo",
+    "lib": ["es2022", "dom"],
```
aswamy, line 10: I noticed we still have an error on this file because of the root-level tsconfig

This can be a post-PR fix:
Image

packages/theme-language-server-common/package.json
```diff
@@ -19,8 +19,8 @@
     "@shopify:registry": "https://registry.npmjs.org"
   },
   "scripts": {
-    "build": "yarn build:ts",
-    "build:ci": "yarn build",
+    "build": "pnpm run build:ts",
```
aswamy: tiny nit: we can just do `pnpm` instead of `pnpm run` everywhere

aswamy (approved):
Looks goood 👍 

Few follow-ups:

- We should also update the `gitignore` files in this repo to remove `yarn-error.log` and replace with whatever is the `pnpm` equivalent
- CONTRIBUTION.md on the root-level  (and in docs and vscode-extension package) needs to update the yarn reference
 - Should we also update the workflow files inside of `prettier-plugin-liquid`?

packages/prettier-plugin-liquid/.github/workflows/ci.yml
graygilmore, line 1: These workflows were dead code. GitHub doesn't run anything that isn't in the root.

scripts/dev-admin-build.sh
graygilmore, line 1: These admin scripts probably don't work at all. I'll look at either fixing them or removing them entirely in a later PR.

packages/vscode-extension/scripts/check-yarn-links.sh
graygilmore, line 1: This was dead code, wasn't called anywhere.

graygilmore:
@EvilGenius13 @aswamy thanks for the reviews! A couple of updates:

1. Removed `packages/vscode-extension/scripts/check-yarn-links.sh` (dead code)
2. Renamed all `pnpm run` to `pnpm` where I could
3. Addressed `contributing.md` and `.gitignore` suggestions
4. Deleted the stale GitHub workflows in the `prettier-plugin-liquid` package (these were never run)

EvilGenius13 (approved):

PR #1211, opened 2026-05-22
Use app blocks as theme graph entry points

graygilmore:
## Summary

Fixes a false positive where `OrphanedSnippet` reported snippets in theme app extensions as unused even when they were rendered by app blocks.

Theme app extension blocks are rendered by Shopify directly, without an in-repo template or section parent. The theme graph only used templates and sections as entrypoints, so the graph never traversed app blocks and never recorded their `{% render %}` references.

## What changed

- passes the theme-check config context into theme graph construction from the Node runner and language server
- treats `blocks/*.liquid` as graph entrypoints only in `context: app`
- leaves normal theme blocks on the existing reachability path so genuinely unused theme block trees can still be reported
- adds regression coverage for a TAE app block rendering a snippet
- adds a patch changeset for the affected packages

## Context

This became visible in `shopify app dev` after the CLI picked up theme-check-node 3.25.0. `OrphanedSnippet` existed before then, but the Node runner did not provide `getReferences`, so the check no-op'd in CLI mode. Once `getReferences` was wired through for other cross-file checks, the existing graph entrypoint gap started surfacing as orphaned snippet warnings.

## 🎩 Tophatting

This is unfortunately tough to test locally because theme-check doesn't use the local version of `theme-graph` (tbh this seems like something we should fix. I feel like all packages should use local package references during development).

I confirmed through a custom script that I only get `OrphanedSnippet` hitting on snippets that are actually orphaned.

### DIFF
--- a/.changeset/little-donkeys-mix.md
+++ b/.changeset/little-donkeys-mix.md
@@ -1,0 +1,7 @@
+---
+"@shopify/theme-language-server-common": patch
+"@shopify/theme-check-node": patch
+"@shopify/theme-graph": patch
+---
+
+Treat theme app extension blocks as theme graph entry points so snippets rendered by app blocks are not reported as orphaned.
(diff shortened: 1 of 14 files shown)


karreiro (approved):
Thank you, @graygilmore!

charlespwd (approved):

EvilGenius13 (approved):

PR #1213, opened 2026-05-27
Add unused translation key check

jamesmengo:

Adds `UnusedTranslationKey`, a new Theme Check rule that reports default locale keys that are not statically referenced by the theme.

This checks:

- `locales/*.default.json`
- `locales/*.default.schema.json`

I'm only adding this to `all` config for now.

In the language server, diagnostics are reported only for open files, but the check will check the entire theme.  
  
A principle I'm following here is to try to **eliminate false positives** by defaulting to not reporting any errors when we can't resolve dynamic translation keys.   
  
This has to strike a balance with making this check completely useless if it's always defaulting to silencing errors. Please scrutinize that specific choice here. Might be Worth trading off additional coverage from this check with simplicity. 

| Example | What happens |
| --- | --- |
| `{{ 'cart.title' \| t }}` | Protects `cart.title` as a literal storefront reference. |
| `{{ 'cart.title' \| translate }}` | Protects `cart.title`; `translate` is handled the same as `t`. |
| `"label": "t:sections.header.settings.title.label"` | Protects `sections.header.settings.title.label` for schema locale files. |
| `{{ 'cart.' \| append: 'title' \| t }}` | Resolves the literal append chain and protects `cart.title`. |
| `{{ 'title' \| prepend: 'cart.' \| t }}` | Resolves the literal prepend chain and protects `cart.title`. |
| `{{ 'products.product.' \| append: state \| t }}` | Treats `products.product.` as a safe prefix and protects `products.product.*`. |
| `{{ 'products.product.' \| replace: 'product.', '' \| append: state \| t }}` | Treats the storefront reference as dynamic because unsupported filters can rewrite the known prefix. |
| `{{ 'cart.items' \| t: count: cart.item_count }}` | Protects pluralization leaves such as `cart.items.one` and `cart.items.other`. |
| `{{ translation_key \| t }}` | Treats the storefront reference as unbounded dynamic and disables storefront unused-key reporting. |
| `{% assign key = 'actions.' \| append: action %}` + `{{ key \| t }}` | Also treated as unbounded dynamic; this PR does not trace Liquid assignment/data flow. |
| `shopify.sentence.words_connector` in a locale file | Ignored by default; not reported as unused. |
| `customer_accounts.order.title` in a locale file | Ignored by default; not reported as unused. Reviewer callout: this default is worth checking. |

Unsupported filters before `t` / `translate` are treated as dynamic rather than trying to preserve a stale prefix. When a storefront reference is dynamic and cannot be resolved to either an exact key or a safe prefix, storefront unused-key reporting is disabled for that theme.

Global `ignore` config is respected when collecting translation references, so ignored Liquid files do not protect keys or trigger dynamic-key suppression.

## What did you learn?

Resolving dynamic keys can be a much larger problem than I initially thought 😆. Oh, silly me.

## Tophatting

- [x] I added screenshots of the changes (before and after the changes if applicable)

image

## Before you deploy

    - [x] I ran the Theme Check build/config generator and confirmed there are no additional generated config changes
    - [x] `theme-app-extension.yml` is not applicable here; generated config did not change
    - [ ] [Shopifolk] I've made a PR to update the [shopify.dev theme check docs](https://github.com/Shopify/shopify-dev/tree/main/content/storefronts/themes/tools/theme-check/checks) if applicable (link PR here).

### DIFF
--- a/packages/theme-check-common/src/checks/unused-translation-key/index.spec.ts
+++ b/packages/theme-check-common/src/checks/unused-translation-key/index.spec.ts
@@ -1,0 +1,373 @@
+import { expect, describe, it } from 'vitest';
+import { check } from '../../test';
+import { UnusedTranslationKey } from '.';
+
+describe('Module: UnusedTranslationKey', () => {
+  it('reports unused default locale keys', async () => {
+    const offenses = await check(
+      {
+        'locales/en.default.json': JSON.stringify({
+          actions: {
+            add: 'Add',
+            remove: 'Remove',
+          },
+        }),
+        'snippets/cart.liquid': `{{ 'actions.add' | t }}`,
+      },
+      [UnusedTranslationKey],
+    );
+
+    expect(offenses).to.have.length(1);
+    expect(offenses).to.containOffense({
+      check: UnusedTranslationKey.meta.code,
+      message: "Translation key 'actions.remove' is not statically referenced",
+      uri: 'file:///locales/en.default.json',
+    });
+    expect(offenses[0]!).to.suggest(
+      JSON.stringify({
+        actions: {
+          add: 'Add',
+          remove: 'Remove',
+        },
+      }),
+      'Delete unused translation key',
+      {
+        startIndex: 0,
+        endIndex: JSON.stringify({
+          actions: {
+            add: 'Add',
+            remove: 'Remove',
+          },
+        }).length,
+        insert: JSON.stringify({ actions: { add: 'Add' } }, null, 2),
+      },
+    );
+  });
+
+  it('does not report non-default locale keys', async () => {
+    const offenses = await check(
+      {
+        'locales/en.default.json': JSON.stringify({
+          actions: {
+            add: 'Add',
+          },
+        }),
+        'locales/fr.json': JSON.stringify({
+          actions: {
+            add: 'Ajouter',
+            remove: 'Supprimer',
+          },
+        }),
+      },
+      [UnusedTranslationKey],
+    );
+
+    expect(offenses).to.have.length(1);
+    expect(offenses[0].uri).to.equal('file:///locales/en.default.json');
+  });
+
+  it('does not report keys referenced with the translate filter', async () => {
+    const offenses = await check(
+      {
+        'locales/en.default.json': JSON.stringify({
+          actions: {
+            add: 'Add',
+          },
+        }),
+        'snippets/cart.liquid': `{{ 'actions.add' | translate }}`,
+      },
+      [UnusedTranslationKey],
+    );
+
+    expect(offenses).to.have.length(0);
+  });
+
+  it('does not report keys referenced by static append chains', async () => {
+    const offenses = await check(
+      {
+        'locales/en.default.json': JSON.stringify({
+          products: {
+            product: {
+              add_to_cart: 'Add to cart',
+            },
+          },
+        }),
+        'snippets/product-form.liquid': `
+          {{ 'products.' | append: 'product.' | append: 'add_to_cart' | t }}
+        `,
+      },
+      [UnusedTranslationKey],
+    );
+
+    expect(offenses).to.have.length(0);
+  });
+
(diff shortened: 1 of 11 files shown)


packages/theme-check-common/src/checks/unused-translation-key/index.spec.ts
```diff
@@ -0,0 +1,337 @@
...
+          enabled: true,
+          ignoreKeys: ['dynamic.*'],
+        },
+      },
+    );
+
+    expect(offenses).to.have.length(0);
+  });
+
+  it('ignores Shopify-provided translation key namespaces by default', async () => {
```
jamesmengo, line 354: @karreiro please sanity check this

packages/theme-check-common/src/utils/translation-references.ts
```diff
@@ -0,0 +1,237 @@
...
+  /**
+   * Whether the theme contains translation filter usage that cannot be resolved
+   * to either a key or a static prefix.
+   */
+  hasDynamicReferences: boolean;
+};
+
+type ResolvedTranslationKey =
+  | { type: 'key'; value: string }
+  | { type: 'prefix'; value: string }
```
jamesmengo, line 48: no suffix because if you prepend a dynamic variable we can't determine what 'tree' / namespace to mark as 'used'

packages/theme-check-common/src/utils/translation-references.ts
```diff
@@ -0,0 +1,259 @@
...
+  type LiquidHtmlNode,
+  type LiteralNode,
+  type PropertyNode,
+  SourceCodeType,
+  type Theme,
+} from '../types';
+import { visit } from '../visitor';
+
+const TRANSLATION_FILTERS = new Set(['t', 'translate']);
+const STATIC_STRING_FILTERS = new Set(['append', 'prepend']);
```
jamesmengo, line 20: we only statically recover append / prepend. Any other filter before `t` / `translate` is treated as dynamic, because filters like `replace`, `remove`, `downcase`, etc. can rewrite the known prefix

packages/theme-check-common/src/checks/unused-translation-key/index.spec.ts
```diff
@@ -0,0 +1,373 @@
...
+      {
+        'locales/en.default.json': JSON.stringify({
+          products: {
+            product: {
+              add_to_cart: 'Add to cart',
+            },
+          },
+        }),
+        'snippets/product-form.liquid': `
+          {{ 'products.' | append: 'product.' | append: 'add_to_cart' | t }}
```
jamesmengo, line 96: @karreiro This is pretty nice, but I'm starting to see how this could scale poorly and get pretty complex. I'm wondering if we should scope this down and just have this valid static keys and to be more aggressive with what we can We consider as dynamic translations which turn off this check

packages/theme-check-common/src/checks/unused-translation-key/index.spec.ts
```diff
@@ -0,0 +1,373 @@
...
+        'snippets/cart.liquid': `
+          {% assign translation_key = 'actions.' | append: action %}
+          {{ translation_key | t }}
+        `,
+      },
+      [UnusedTranslationKey],
+    );
+
+    expect(offenses).to.have.length(0);
+  });
```
charlespwd, line 160: Is it a "as soon as you have a dynamic, we can't report errors" kind of situation?

jamesmengo: Unfortunately, yes.  
I did a best effort approach for suporting `append` and `prepend` tags (if they're provided static values), but I'm questioning whether it's worth doing so

charlespwd: No I think it's fine as is. Honestly this is a tough nut. I'm glad there's this there because otherwise we'd get false positives and those are worse than no red lines at all IMO.

PR #1214, opened 2026-05-28
Improve LiquidHTMLSyntaxError parity with Ruby Liquid syntax errors

karreiro:

This updates `LiquidHTMLSyntaxError` so Theme Check reports Liquid syntax errors with behavior that tracks Ruby Liquid more closely across the copied parity scenarios. The change keeps the existing check entrypoint, but adds a dedicated `checks/liquid-syntax-error` compatibility layer for Liquid tag and variable syntax detection where the current parser either accepts markup that Ruby Liquid rejects or reports it through a different path.

The implementation is split by Liquid construct, with shared tokenizer and markup helpers in `utils.ts` and focused detectors for tags such as `assign`, `case`, `comment`, `doc`, `for`, `if`, `elsif`, `unless`, `include`, `render`, `raw`, `schema`, and variable output. The parity harness loads the copied Ruby Liquid scenario and snapshot YAML files, registers every scenario as a strict test case, and treats both missed Ruby Liquid errors and Theme Check-only reports as failures.

## Tophatting

Run the Liquid syntax parity and check-level suites from the `theme-tools` repo:

```bash
pnpm exec vitest run packages/theme-check-common/src/checks/liquid-html-syntax-error/parity.spec.ts
PARITY_REPORT=1 pnpm exec vitest run packages/theme-check-common/src/checks/liquid-html-syntax-error/parity.spec.ts --reporter=basic
```


packages/theme-check-common/src/checks/liquid-html-syntax-error/index.ts
```diff
@@ -10,12 +17,127 @@ import { detectConditionalNodeUnsupportedParenthesis } from './checks/InvalidCon
...
+} from './checks/liquid-syntax-error/raw';
+import { checkRenderTag } from './checks/liquid-syntax-error/render';
+import { checkSchemaSourceStructure, checkSchemaTag } from './checks/liquid-syntax-error/schema';
+import { checkSectionTag } from './checks/liquid-syntax-error/section';
+import { checkSectionsTag } from './checks/liquid-syntax-error/sections';
+import { checkStyleTag } from './checks/liquid-syntax-error/style';
+import { checkStylesheetTag } from './checks/liquid-syntax-error/stylesheet';
+import { checkTablerowTag } from './checks/liquid-syntax-error/tablerow';
+import { checkUnlessTag } from './checks/liquid-syntax-error/unless';
+import { checkVariableOutput } from './checks/liquid-syntax-error/variable';
```
charlespwd: Oh my goodness.

karreiro: I've reorganized the strategy imports. Even though we're still delegating the check for each tag, it's a bit clearer now.

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/liquid-syntax-error/if.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  if (
+    hasInvalidBooleanExpressionComparisonMarkup(tagMarkup) ||
+    hasInvalidBooleanExpressionLexerMarkup(tagMarkup) ||
+    hasInvalidBooleanExpressionLookupMarkup(tagMarkup) ||
+    hasInvalidBooleanExpressionTokenMarkup(tagMarkup) ||
+    hasInvalidBooleanComparisonRhsMarkup(tagMarkup) ||
+    hasInvalidBooleanComparisonRhsLookupMarkup(tagMarkup) ||
+    hasInvalidComparisonRhsMarkup(tagMarkup)
+  ) {
```
aswamy: This path appears to retokenize the same `tagMarkup` many times. In a stress test with a single 4 MB quoted literal inside an `{% if %}` tag, this branch went from ~5.0s / ~2.9 GB memory on main to ~12.0s / ~3.9 GB memory on this PR.

If we have something like `analyzeMarkup` that would memoize the tokens, we'd be good:

E.g.
```
const tagMarkupAnalysis = analyzeMarkup(tagMarkup);

if (
  hasInvalidBooleanExpressionComparisonMarkup(tagMarkupAnalysis) ||
  hasInvalidBooleanExpressionLexerMarkup(tagMarkupAnalysis) ||
  hasInvalidBooleanExpressionLookupMarkup(tagMarkupAnalysis) ||
  ...
) {
  ...
}
```

`analyzeMarkup(tagMarkup)` would return something like:
```
{
  source: markup,
  tokens: meaningfulTokens(markup),       // computed once
  hasSkippedCharacters: boolean,          // computed once
  uncoveredCharacters?: UncoveredCharacter[]
}
```

aswamy (commented):
Didn't check every piece of code since we're just porting over logic from our ruby side.

I did run a perf check to see if we significantly increase in CPU or memory against Horizon, but looks like we're in the clear.

> PR branch: avg ~16.6s real, ~23.2s user, max RSS avg ~859 MB.
> origin/main: avg ~17.2s real, ~23.8s user, max RSS avg ~876 MB.

I let the AI do some stress testing and found some issues when reviewing large markup. Not a blocker, but the improvement could mitigate the issue. See comment below.

packages/theme-check-common/src/checks/liquid-html-syntax-error/checks/liquid-syntax-error/if.ts
karreiro: Thank you for the suggestion, @aswamy. I've applied that enhancement in the latest commit.

I benchmarked the 4 MiB quoted `{% if %}` case by running `runLiquidCheck(LiquidHTMLSyntaxError, source)` with one warm-up run and three measured runs per commit, and it's ~30% faster in wall time.

aswamy (approved):
I couldn't tophat the issue because of the pnpm issue we spoke about, but the rest looks fine. Thanks for this 👍

karreiro:
Removed in favor of https://github.com/Shopify/theme-tools/pull/1252

PR #1221, opened 2026-06-01
Migrate to Vitest 4

graygilmore:
## Summary

Bumps Vitest from `2.1.9` to `4.1.0` and applies the migration cleanup needed for CI:

- aligns package-local `@vitest/expect` versions with the root Vitest runner
- removes the old `ChaiUtils.eql` type augmentation now that Chai owns that type
- adds explicit Node test types for packages that use Node globals in test helpers
- moves the fork worker config to the Vitest 4 top-level options
- tightens a couple of language server test mock types that Vitest 4 now exposes more strictly

## Context

This supersedes https://github.com/Shopify/theme-tools/pull/1219. The Dependabot bump alone fails during the build because the older test setup no longer type-checks against the Vitest 4/Chai types.

This keeps the dependency upgrade but makes it an explicit migration PR instead of carrying source changes on the Dependabot branch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "tsconfig-paths-webpack-plugin": "^4.1.0",
     "typescript": "^5.9.0",
     "vite": "^6.4.2",
-    "vitest": "^2.1.9",
+    "vitest": "^4.1.0",
     "vscode-languageserver-protocol": "^3.17.3",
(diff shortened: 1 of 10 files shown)


karreiro (approved):
Thank you, @graygilmore!

PR #1222, opened 2026-06-02
fix(RemoteAsset): ignore base64 data URI images

andershagbard:
## Summary

- The `RemoteAsset` check was incorrectly flagging `src` attributes containing base64-encoded data URIs (e.g. `src="data:image/gif;base64,..."`) as needing to be served via the Shopify CDN.
- Data URIs are fully self-contained — they embed the image bytes directly in the attribute value and create **no external HTTP connections**, so the check's performance concern does not apply.
- Added an `isDataUri` helper (mirrors the existing `isHashUrl` helper) and an early-return guard in `checkHtmlNode` so data URIs are silently skipped.

## Example

```liquid
{{! Before: this was incorrectly reported as an offense }}
<img
  src="data:image/gif;base64,R0lGODlhAQABAAAAACH5BAEKAAEALAAAAAABAAEAAAICTAEAOw=="
/>
```

## Test plan

- [x] `should not report an offense for base64 encoded data URIs` — single GIF data URI passes
- [x] `should not report an offense for various base64 data URI image formats` — PNG, JPEG, SVG, WebP data URIs all pass
- [x] `should still report an offense for remote images alongside base64 images` — mixed case: base64 is ignored, remote URL is still caught
- [x] All 22 tests in `remote-asset/index.spec.ts` pass

### DIFF
--- a/packages/theme-check-common/src/checks/remote-asset/index.ts
+++ b/packages/theme-check-common/src/checks/remote-asset/index.ts
@@ -47,6 +47,10 @@ function isHashUrl(url: string): boolean {
   return url.startsWith('#');
 }
 
+function isDataUri(url: string): boolean {
+  return url.startsWith('data:');
+}
+
 /**
  * Checks if the attribute value starts with a variable lookup.
  * When a value starts with a VariableLookup (e.g., {{ source.url }}, {{ image }}),
@@ -184,6 +188,7 @@ export const RemoteAsset: LiquidCheckDefinition<typeof schema> = {
         (node): node is TextNode => node.type === NodeTypes.TextNode,
       );
       if (firstTextNode && isHashUrl(firstTextNode.value)) return;
+      if (firstTextNode && isDataUri(firstTextNode.value)) return;
 
       if (startsWithVariableLookup(urlAttribute)) return;
 
--- a/packages/theme-check-common/src/checks/remote-asset/index.ts
+++ b/packages/theme-check-common/src/checks/remote-asset/index.ts
@@ -240,6 +240,9 @@ export const RemoteAsset: LiquidCheckDefinition<typeof schema> = {
       if (hasAsset) return;
 
       const urlNode = parentNode.expression;
+
+      if (urlNode.type === NodeTypes.String && isDataUri(urlNode.value)) return;
+
       if (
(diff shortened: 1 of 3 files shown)


packages/theme-check-common/src/checks/remote-asset/index.ts
```diff
@@ -47,6 +47,10 @@ function isHashUrl(url: string): boolean {
   return url.startsWith('#');
 }
 
+function isDataUri(url: string): boolean {
+  return url.startsWith('data:');
```
graygilmore: I believe URL schemes like this are case-insensitive. I'm pretty sure I've seen folks use `DATA` for example. Let's tweak this a bit. Could be just a small tweak to something like:

```suggestion
  return /^data:/i.test(url);
```

packages/theme-check-common/src/checks/remote-asset/index.ts
```diff
@@ -184,6 +188,7 @@ export const RemoteAsset: LiquidCheckDefinition<typeof schema> = {
         (node): node is TextNode => node.type === NodeTypes.TextNode,
       );
       if (firstTextNode && isHashUrl(firstTextNode.value)) return;
+      if (firstTextNode && isDataUri(firstTextNode.value)) return;
```
graygilmore, line 191: Could we add this check to `checkLiquidFilter`, too? That way we get coverage on `{{ 'data:image/gif;base64,...' | img_tag }}`. It would look something like:

```diff
  const urlNode = parentNode.expression;
+ 
+ if (urlNode.type === NodeTypes.String && isDataUri(urlNode.value)) return;

  if (
    urlNode.type === NodeTypes.String &&
    !isUrlHostedbyShopify(urlNode.value, allowedDomains)
  ) {
    context.report({
      message: 'Asset should be served by the Shopify CDN for better performance.',
      startIndex: urlNode.position.start,
      endIndex: urlNode.position.end,
    });
  }
```

graygilmore (commented):
Love this! Two small tweaks suggested.

graygilmore (approved):
Thanks @andershagbard! Just needs a changeset which you can create with `pnpm changeset add`. A `patch` is fine for this.

graygilmore (approved):

PR #1224, opened 2026-06-04
Support placeholder on color settings

NathanPJF:

This PR wires up support for `placeholder` on `color` settings. 

Extending `Color` as `placeholder` is not available to all `ColorBase` extensions

Related PR in theme-liquid-docs: https://github.com/Shopify/theme-liquid-docs/pull/1709

## Before you deploy

### DIFF
--- a/.changeset/color-placeholder-i18n.md
+++ b/.changeset/color-placeholder-i18n.md
@@ -1,6 +1,0 @@
----
-'@shopify/theme-check-common': patch
-'@shopify/theme-language-server-common': patch
----
-
-Allow `placeholder` on `color` settings in theme schemas.
--- a/.changeset/color-placeholder-i18n.md
+++ b/.changeset/color-placeholder-i18n.md
@@ -1,0 +1,6 @@
+---
+'@shopify/theme-check-common': patch
+'@shopify/theme-language-server-common': patch
+---
+
+Allow `placeholder` on `color` settings in theme schemas.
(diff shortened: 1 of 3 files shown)


packages/theme-check-common/src/types/schemas/setting.ts
```diff
@@ -196,7 +196,10 @@ export declare namespace Setting {
     default?: string;
   }
 
-  export interface Color extends ColorBase<Type.Color> {}
+  export interface Color extends ColorBase<Type.Color> {
+    placeholder?: string;
```
aswamy, line 200: Could you also add the following prop please :pray:

```
alpha?: boolean;
```

Seems like this file has drifted

.changeset/color-placeholder-i18n.md
```diff
@@ -0,0 +1,6 @@
+---
+'@shopify/theme-check-common': patch
+'@shopify/theme-language-server-common': patch
```
aswamy: I dont think you need this package in this changeset

aswamy (approved):
Just make sure you run `pnpm format` to make CI pass before merging after the changes

.changeset/color-placeholder-i18n.md
NathanPJF: @aswamy you're right. I had an additional update in here, but I pulled that file out but didn't correct this line. Thanks!

PR #1225, opened 2026-06-05
Fix pnpm workspace package debugging

aswamy:
## Summary
- Use workspace protocol for internal package dependencies so pnpm links local packages during development
- Restore VS Code source maps to local workspace sources for breakpoints
- Include changeset

## Test plan
- pnpm install --frozen-lockfile --offline
- pnpm --dir packages/vscode-extension run type-check
- pnpm --dir packages/vscode-extension run build:extension:dev


graygilmore (approved):
Thank you!

PR #1230, opened 2026-06-09
fix(JSONMissingBlock): allow app block references

### LINKED ISSUE #1144: JSONMissingBlock false positive for app block when section supports both @app and @theme
https://github.com/Shopify/theme-tools/blob/5612587be7af61fdc9c94b1ee621f1b89cf6f9ba/packages/theme-check-common/src/checks/json-missing-block/missing-block-utils.ts#L31

**Describe the bug**
When a section schema allows both @app and @theme blocks, Theme Check reports JSONMissingBlock for valid app block types in [product.json](app://-/index.html#).
It tries to resolve the app block URI as a local theme block file ([... .liquid](app://-/index.html#)), which is incorrect.

**Source**
When having blocks defined in section defined like below
```liquid
blocks: [{ type: '@app' }, { type: '@theme' }],
```
allows us to add two different types of blocks. This can result in structure like below in product.json
```liquid
"main": {
      "type": "main-product",
      "blocks": {
        "some_app_block": {
          "type": "shopify://apps/some_app_block/....",
        }
      },
```
which will generate a false error
```liquid
[error]: JSONMissingBlock
Theme block 'blocks/shopify://apps/some_app_block/.....liquid' does not exist.
```

**Expected behaviour**
App block URIs (shopify://apps/...) should be treated as valid when section schema includes @app, even if same section also includes @theme.

**Actual behaviour**
Theme Check raises JSONMissingBlock and treats the app block URI as a missing local theme block file.

chippleh1392:

Fixes a false positive where `JSONMissingBlock` reports app block references as missing local block files when the containing schema allows `@app`.

App block references use `shopify://apps/...` types and do not map to local `blocks/*.liquid` files. This change skips local file validation for those references while still requiring the containing section or block schema to include `@app`.

Regression coverage includes schemas allowing `@app` with `@theme`, schemas allowing only `@app`, and schemas that do not allow app blocks.

Fixes #1144

None planned.

## Tophatting

Run Theme Check against a JSON template containing a `shopify://apps/...` block reference:

- With `@app` in the containing schema, `JSONMissingBlock` does not report an offense.
- Without `@app`, `JSONMissingBlock` reports that the block type is not allowed.
- Local block validation continues to report missing or disallowed local blocks.

Validation completed:

- `pnpm build`
- `CI=true pnpm exec vitest run`
- `pnpm type-check`
- `pnpm format:check`

## Before you deploy

### DIFF
--- a/.changeset/calm-penguins-check.md
+++ b/.changeset/calm-penguins-check.md
@@ -1,0 +1,5 @@
+---
+'@shopify/theme-check-common': patch
+---
+
+Avoid reporting missing block offenses for app block references when the containing schema allows app blocks.
--- a/packages/theme-check-common/src/checks/json-missing-block/index.spec.ts
+++ b/packages/theme-check-common/src/checks/json-missing-block/index.spec.ts
@@ -80,6 +80,119 @@ describe('Module: JsonMissingBlock', () => {
       expect(offenses).to.be.empty;
     });
 
+    it('should not report an offense for an app block when the section allows @app and @theme', async () => {
+      const theme: MockTheme = {
+        'templates/product.app-block.json': `{
+          "sections": {
+            "main": {
+              "type": "main-product",
+              "blocks": {
+                "some_app_block": {
+                  "type": "shopify://apps/some-app/blocks/example/1234"
+                }
+              },
+              "block_order": ["some_app_block"]
+            }
+          },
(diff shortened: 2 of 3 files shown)


chippleh1392:
I have signed the CLA!

PR #1234, opened 2026-06-14
fix(liquid-html-parser): support filter expressions in render named arguments

andershagbard:
**Redacted to bury in search engines**


andershagbard:
My mistake. No issue here. PR description seems like AI slop

PR #1240, opened 2026-06-19
Add Liquid docs bundle command

charlespwd:

Related issue: Shopify/developer-tools-team#783

Adds a stable Liquid docs bundle generator to `@shopify/theme-check-docs-updater`.

New package API:

```ts
import { buildThemeLiquidDocsBundle } from '@shopify/theme-check-docs-updater';

const bundle = await buildThemeLiquidDocsBundle();
```

New CLI command:

```sh
pnpm --dir packages/theme-check-docs-updater theme-docs bundle
```

The command writes a self-contained JSON bundle to stdout:

```json
{
  "schemaVersion": 1,
  "revision": "40f966fc7793462726d9f8573c8d522f8ab44c54",
  "tags": [],
  "filters": [],
  "objects": [],
  "systemTranslations": {},
  "schemas": []
}
```

Logs from docs loading go to stderr so stdout can be redirected directly into a JSON artifact.

Why: browser-based hosts need a way to publish Liquid docs updates independently from theme-tools and extension releases. This PR only adds the generator; it does not change browser extension runtime behavior.

Follow-up PR: #1016 teaches the browser extension to fetch a published docs bundle at runtime and fall back to bundled docs if unavailable.

A separate host/CDN publisher can use this command after this package version is released, for example:

```sh
pnpm dlx @shopify/theme-check-docs-updater@<version> theme-docs bundle > liquid-docs-bundle.json
```

## Tophatting

- [x] `pnpm --dir packages/theme-check-docs-updater type-check`
- [x] `CI=1 pnpm vitest run packages/theme-check-docs-updater/src/themeLiquidDocsBundle.spec.ts packages/theme-check-docs-updater/src/themeLiquidDocsManager.spec.ts`
- [x] `pnpm --dir packages/theme-check-docs-updater build`
- [x] `pnpm --silent --dir packages/theme-check-docs-updater theme-docs bundle | jq '.revision, .schemaVersion'`
- [x] `pnpm format:check`
- [x] `git diff --check`
- [x] Screenshots are not applicable; this is CLI/package API behavior.

## Before you deploy

### DIFF
--- a/packages/theme-check-docs-updater/scripts/cli.js
+++ b/packages/theme-check-docs-updater/scripts/cli.js
@@ -2,47 +2,77 @@
 
 const path = require('path');
 const fs = require('fs');
-const { downloadThemeLiquidDocs, root } = require(path.resolve(__dirname, '../dist'));
+const {
+  ThemeLiquidDocsManager,
+  buildThemeLiquidDocsBundle,
+  downloadThemeLiquidDocs,
+  root,
+} = require(path.resolve(__dirname, '../dist'));
 
 // Get the command line arguments
 const args = process.argv.slice(2);
 
-// Check if a command was provided
-if (args.length === 0) {
+function printUsage() {
   console.log(`
 Please provide a command.
 
 Usage:
  download <dir> \t\tDownloads all docsets and JSON Schemas to the specified directory.
+ bundle \t\t\tPrints a self-contained Liquid docs JSON bundle.
  root \tPrints the default docsets root directory.
  clear-cache \tClears the default docsets root directory.
 `);
-  process.exit(1);
 }
 
-// Handle the command
-switch (args[0]) {
-  case 'download':
-    if (args.length > 2) {
-      console.log('Please provide a directory to download docs into.');
-      process.exit(1);
-    }
-    console.log('Downloading docs...');
-
-    downloadThemeLiquidDocs(args[1], console.error.bind(console));
-
-    break;
-
-  case 'root':
-    console.log(root);
-    break;
-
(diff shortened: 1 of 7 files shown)


packages/theme-check-docs-updater/scripts/cli.js
```diff
@@ -2,47 +2,77 @@
...
 
-// Check if a command was provided
-if (args.length === 0) {
+function printUsage() {
   console.log(`
 Please provide a command.
 
 Usage:
  download <dir> \t\tDownloads all docsets and JSON Schemas to the specified directory.
+ bundle \t\t\tPrints a self-contained Liquid docs JSON bundle.
```
FKauwe, line 21: this is the first time I've seen the `\t\t\t `and asked Claude about it. Apparently it's a way to align the descriptions into columns. C mentioned that there are some libraries that exist that that handle alignment automatically. Just curious what the tradeoffs are between this syntax and a library-- maybe a whole library is just overkill and the current syntax is more lightweight for the purposes you need?

FKauwe (approved):
I tophatted and the commands compile, build, run on my machine and the unit tests test both happy and sad path. This seems like a helpful addition to have "fresh" docs available in between extension releases

packages/theme-check-docs-updater/scripts/cli.js
charlespwd: `\t` is a tab. While we could use a library, I think it's overkill for this CLI since this is mostly for internal use. It's not worth the trouble nor the dependency risk.

PR #1248, opened 2026-06-26
Add UnsupportedFilterArguments check

NathanPJF:
### What & why

Filters aren't supported on values passed as arguments to render/include tags — but today theme-check says nothing when you write one. The value is silently passed through unchanged, so this:

 ```liquid
   {% render 'paragraph', text: 'Hello,' | append: ' World!' %}
 ```

renders `Hello,` (not `Hello, World!`) with zero feedback making it hard to debug or notice unexpected markup in production.

Examples of what happens in a production Shopify store

<details><summary>Code used</summary>

```html
<h1>Test page</h1>

{% render 'paragraph', text: 'Sample text' %}

{% render 'paragraph', text: 'Hello,' | append: ' World!' %}

{% assign id = 'unique-id' %}

{% render 'paragraph', text: 'ID:' | append: id %}

{% render 'paragraph', text: 4 | plus: 2 %}
```

</details>

Screenshot 2026-06-26 at 11 16 49 AM

### Real world use-case: 

We were creating unique IDs to pass into snippets, but making them unique inline at the render tag. The result was all the `id` values got rendered as the same value. 😬

**Root cause**: theme-check's parser is tolerant — when a tag doesn't match the strict grammar, it doesn't error, it just stores the tag as a raw text blob and moves on. The grammar intentionally doesn't allow filters on render arguments, so `{% render 'foo', x: 'a' | append: 'b' %}` fails the strict parse and falls back to a generic LiquidTag holding the raw string, instead of becoming the structured RenderMarkup node a valid render produces. No error is raised, which is the silent failure. It also means existing render checks (which run on RenderMarkup nodes) never see it — so this check runs on the generic LiquidTag fallback and detects the leftover filter pipe directly.

### Approach

A LiquidTag-visitor check that keys off exactly that fallback:

1. Only render / include tags.
2. Only when typeof node.markup === 'string' (i.e. strict parse failed).
3. Find the first `|` not inside a quoted string literal (so `sep: 'a|b'` doesn't false-positive).
4. Report an error pointing at the filter through the end of the markup.

This throws a clean, hard error. The shared grammar is untouched.

**Decision** Severity is `ERROR` and the check is recommended (on by default), since the failure is silent and produces incorrect output. We could change this to `WARNING`, but I think that `ERROR` is appropriate since we are protecting against unexpected outputs that can affect the site's functionality.

**Limitation**: the guard is a heuristic on raw markup ("render/include that failed strict parse + contains an unquoted pipe"), because the AST node genuinely doesn't exist for malformed renders. This means no per-argument
precision (e.g. "filter on param2"). Acceptable given an unquoted `|` inside a render tag is essentially always a filter attempt.

### Changes

- New check: packages/theme-check-common/src/checks/no-filters-in-render-arguments/{index,index.spec}.ts
- Registered in packages/theme-check-common/src/checks/index.ts
- Generated configs (all.yml, recommended.yml) regenerated via pnpm build
- Changeset (minor on @shopify/theme-check-common)

### Testing

My test cases:

```
{% render 'paragraph', text: 'Sample text' %}

{% render 'paragraph', text: 'Hello,' | append: ' World!' %}

{% assign id = 'unique-id' %}

{% render 'paragraph', text: 'ID:' | append: id %}

{% render 'paragraph', text: 4 | plus: 2 %}
```

6 new unit specs (render filter, include filter, no-filter, 'a|b' no false positive, plain render, nested) — all green; full theme-check-common suite (959 tests) passes.

I ran the local build against my test file. 3 real offenses, 0 false positives.

```bash
  pnpm build                                                                                                         
  pnpm theme-check /path/to/theme 2>/dev/null \                                                                      
    | sed -n '/^\[/,$p' \                                                                                            
    | jq -rs '.[0][] | select(.severity == 0) | "\(.uri):\(.start.line + 1)  [\(.check)] \(.message)"'               
```

Output:

```
templates/page.blank.liquid:5 [NoFiltersInRenderArguments] Filters cannot be used on arguments passed to the 'render' tag. Apply the filter beforehand (e.g. with {% assign %}) and pass the result instead.
templates/page.blank.liquid:9 [NoFiltersInRenderArguments] Filters cannot be used on arguments passed to the 'render' tag. Apply the filter beforehand (e.g. with {% assign %}) and pass the result instead.
templates/page.blank.liquid:11 [NoFiltersInRenderArguments] Filters cannot be used on arguments passed to the 'render' tag. Apply the filter beforehand (e.g. with {% assign %}) and pass the result instead.
```

 Co-Authored by AI

## Before you deploy

Check changes 
  - [x] I ran `pnpm build` and committed the updated configuration files
  - [x] [Shopifolk] I've made a PR to update the [shopify.dev theme check docs](https://github.com/Shopify/shopify-dev/tree/main/content/storefronts/themes/tools/theme-check/checks) if applicable 
    - [PR for shopify-dev addition](https://github.com/shop/world/pull/880656) 

Public API changes, new features

Closes shop/issues#58903

### DIFF
--- a/packages/theme-check-common/src/checks/no-filters-in-render-arguments/index.ts
+++ b/packages/theme-check-common/src/checks/no-filters-in-render-arguments/index.ts
@@ -1,0 +1,86 @@
+import { LiquidCheckDefinition, Severity, SourceCodeType } from '../../types';
+
+/**
+ * Finds the index of the first filter pipe (`|`) in a raw render/include markup
+ * string that is *not* inside a quoted string literal. Returns -1 if none is
+ * found.
+ *
+ * This is a heuristic on the raw markup because, when a `render`/`include` tag
+ * contains a filter, the strict grammar refuses it and the parser falls back to
+ * a base-case `LiquidTag` whose `markup` is a raw string (there is no structured
+ * `RenderMarkup` node to inspect).
+ */
+function indexOfFilterPipe(markup: string): number {
+  let quote: "'" | '"' | null = null;
+
+  for (let i = 0; i < markup.length; i++) {
+    const char = markup[i];
+
+    if (quote) {
+      if (char === quote) {
+        quote = null;
+      }
+      continue;
+    }
+
+    if (char === "'" || char === '"') {
+      quote = char;
+      continue;
+    }
+
+    if (char === '|') {
+      return i;
+    }
+  }
+
+  return -1;
+}
+
+export const NoFiltersInRenderArguments: LiquidCheckDefinition = {
+  meta: {
+    code: 'NoFiltersInRenderArguments',
+    name: 'No Filters in Render Arguments',
+    docs: {
+      description:
+        "This check warns against using filters on values passed as arguments to a 'render' or 'include' tag. " +
+        'Filters are not applied in that position and the value is passed through unchanged, which silently ' +
+        'produces incorrect output.',
+      recommended: true,
+      url: 'https://shopify.dev/docs/storefronts/themes/tools/theme-check/checks/no-filters-in-render-arguments',
+    },
+    type: SourceCodeType.LiquidHtml,
+    severity: Severity.ERROR,
+    schema: {},
+    targets: [],
+  },
+
+  create(context) {
+    return {
+      async LiquidTag(node) {
+        if (node.name !== 'render' && node.name !== 'include') return;
+
+        // When the tag parses successfully, `markup` is a structured
+        // `RenderMarkup` object and filters are impossible. A raw string means
+        // the strict parse failed (e.g. because of a filter pipe).
+        if (typeof node.markup !== 'string') return;
+
+        const relativePipeIndex = indexOfFilterPipe(node.markup);
+        if (relativePipeIndex === -1) return;
+
+        const markupStart = node.source.indexOf(node.markup, node.position.start);
+        if (markupStart === -1) return;
+
+        const startIndex = markupStart + relativePipeIndex;
+        const endIndex = markupStart + node.markup.length;
+
+        context.report({
+          message:
+            `Filters cannot be used on arguments passed to the '${node.name}' tag. ` +
+            `Apply the filter beforehand (e.g. with {% assign %}) and pass the result instead.`,
+          startIndex,
+          endIndex,
+        });
+      },
+    };
+  },
+};
--- a/packages/theme-check-common/src/checks/no-filters-in-render-arguments/index.ts
+++ b/packages/theme-check-common/src/checks/no-filters-in-render-arguments/index.ts
@@ -1,12 +1,15 @@
+import { NamedTags } from '@shopify/liquid-html-parser';
 import { LiquidCheckDefinition, Severity, SourceCodeType } from '../../types';
 
+const CHECKED_TAGS: string[] = [NamedTags.render, NamedTags.include, NamedTags.content_for];
+
 /**
- * Finds the index of the first filter pipe (`|`) in a raw render/include markup
- * string that is *not* inside a quoted string literal. Returns -1 if none is
+ * Finds the index of the first filter pipe (`|`) in a raw render/include/content_for
+ * markup string that is *not* inside a quoted string literal. Returns -1 if none is
  * found.
  *
- * This is a heuristic on the raw markup because, when a `render`/`include` tag
- * contains a filter, the strict grammar refuses it and the parser falls back to
+ * This is a heuristic on the raw markup because, when a `render`/`include`/`content_for`
+ * tag contains a filter, the strict grammar refuses it and the parser falls back to
  * a base-case `LiquidTag` whose `markup` is a raw string (there is no structured
  * `RenderMarkup` node to inspect).
  */
@@ -42,7 +45,7 @@ export const NoFiltersInRenderArguments: LiquidCheckDefinition = {
     name: 'No Filters in Render Arguments',
     docs: {
       description:
-        "This check warns against using filters on values passed as arguments to a 'render' or 'include' tag. " +
+        "This check warns against using filters on values passed as arguments to a 'render', 'include', or 'content_for' tag. " +
         'Filters are not applied in that position and the value is passed through unchanged, which silently ' +
         'produces incorrect output.',
       recommended: true,
@@ -57,7 +60,7 @@ export const NoFiltersInRenderArguments: LiquidCheckDefinition = {
   create(context) {
     return {
       async LiquidTag(node) {
-        if (node.name !== 'render' && node.name !== 'include') return;
+        if (!CHECKED_TAGS.includes(node.name)) return;
 
         // When the tag parses successfully, `markup` is a structured
         // `RenderMarkup` object and filters are impossible. A raw string means
--- a/packages/theme-check-common/src/checks/unsupported-filter-arguments/index.spec.ts
+++ b/packages/theme-check-common/src/checks/unsupported-filter-arguments/index.spec.ts
@@ -1,11 +1,11 @@
 import { expect, describe, it } from 'vitest';
 import { highlightedOffenses, runLiquidCheck } from '../../test';
-import { NoFiltersInRenderArguments } from './index';
+import { UnsupportedFilterArguments } from './index';
 
-describe('Module: NoFiltersInRenderArguments', () => {
+describe('Module: UnsupportedFilterArguments', () => {
   it('reports an offense when a render argument uses a filter', async () => {
     const sourceCode = `{% render 'foo', param1: 'bar', param2: 'hello' | append: ' world' %}`;
-    const offenses = await runLiquidCheck(NoFiltersInRenderArguments, sourceCode);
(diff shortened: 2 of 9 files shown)


packages/theme-check-common/src/checks/unsupported-filter-arguments/index.spec.ts
mbarak, line 1: 💡 **Improvement**: The current specs cover the core offense/no-offense paths, but they do not fully pin the two most regression-prone pieces of the implementation: quote-aware pipe selection and mapping a raw-markup-relative offset back to the source. A broken scanner or source offset calculation could still pass if only the offense count is asserted.

**Suggestion:** Add focused regression specs for: a quoted pipe before a real filter, asserting the highlight starts at the real filter; two render/include tags on one line, asserting distinct highlights; a render/include statement inside a `{% liquid %}` tag; and quoted-pipe false positives in the raw fallback path, including double-quoted strings if practical.

packages/theme-check-common/src/checks/unsupported-filter-arguments/index.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  it('does not report when no filter is used', async () => {
+    const sourceCode = `{% render 'foo', param1: 'bar', param2: 'hello' %}`;
+    const offenses = await runLiquidCheck(NoFiltersInRenderArguments, sourceCode);
+
+    expect(offenses).toHaveLength(0);
+  });
+
+  it('does not report a false positive for a pipe inside a string literal', async () => {
+    const sourceCode = `{% render 'foo', sep: 'a|b' %}`;
```
mbarak, line 53: 💡 **Improvement**: The current quoted-pipe fixture is valid render markup, so the parser can return structured `RenderMarkup` and the check exits before calling the raw-markup scanner. That still proves valid quoted pipes do not report, but it would not catch a regression in `indexOfFilterPipe` itself.

**Suggestion:** Consider replacing this fixture, or adding a separate malformed fallback case, so the raw scanner runs while the only pipe is still inside a quoted string:

```suggestion
    const sourceCode = `{% render 'foo', sep: 'a|b', invalid_argument %}`;
```

packages/theme-check-common/src/checks/unsupported-filter-arguments/index.ts
```diff
@@ -0,0 +1,86 @@
...
+    docs: {
+      description:
+        "This check warns against using filters on values passed as arguments to a 'render' or 'include' tag. " +
+        'Filters are not applied in that position and the value is passed through unchanged, which silently ' +
+        'produces incorrect output.',
+      recommended: true,
+      url: 'https://shopify.dev/docs/storefronts/themes/tools/theme-check/checks/no-filters-in-render-arguments',
+    },
+    type: SourceCodeType.LiquidHtml,
+    severity: Severity.ERROR,
```
mbarak, line 55: 📐 **Design**: This check is both recommended and `Severity.ERROR`, so themes currently using filters in render/include arguments can start failing CI as soon as they upgrade. That may be the right tradeoff because the behavior silently produces incorrect output, but it is stricter than the surrounding render-argument check family, which otherwise reports warnings.

**Suggestion:** Confirm maintainers are comfortable with this shipping as a recommended error. If a softer rollout is preferred, consider starting as `Severity.WARNING` and promoting later; otherwise, keeping `Severity.ERROR` is reasonable if the breaking-on-upgrade behavior is intentional.

packages/theme-check-common/src/checks/unsupported-filter-arguments/index.ts
```diff
@@ -0,0 +1,86 @@
...
+    schema: {},
+    targets: [],
+  },
+
+  create(context) {
+    return {
+      async LiquidTag(node) {
```
mbarak, line 62: 🎨 **Code Style**: The `LiquidTag` visitor does not contain any `await` expressions or other asynchronous work. Since visitor methods can be synchronous, dropping `async` keeps the implementation slightly simpler and avoids unnecessary Promise wrapping.

**Suggestion:** 
```suggestion
      LiquidTag(node) {
```

packages/theme-check-common/src/checks/no-filters-in-render-arguments/index.ts
```diff
@@ -0,0 +1,86 @@
...
+    type: SourceCodeType.LiquidHtml,
+    severity: Severity.ERROR,
+    schema: {},
+    targets: [],
+  },
+
+  create(context) {
+    return {
+      async LiquidTag(node) {
+        if (node.name !== 'render' && node.name !== 'include') return;
```
mbarak: 💬 **Suggestion**: The guard currently limits the check to `render` and `include`, but `content_for` appears to have the same grammar shape: filters in named arguments are not part of the structured markup parse and can fall back to raw string markup. If so, the same silent-filter behavior this check catches for render/include may remain uncaught for `content_for`.

**Suggestion:** Consider extending the tag guard to include `content_for` as well, or documenting in the PR why `content_for` is intentionally out of scope for this check.

mbarak: 🎨 **Code Style**: Several LiquidTag-oriented checks use the parser's `NamedTags` enum rather than string literals for known tag names. Using the enum here would make this guard match the surrounding style and reduce the chance of future spelling drift.

**Suggestion:** If this remains a two-tag guard, consider importing `NamedTags` from `@shopify/liquid-html-parser` and using:

```suggestion
        if (node.name !== NamedTags.render && node.name !== NamedTags.include) return;
```

mbarak (approved):

packages/theme-check-common/src/checks/unsupported-filter-arguments/index.ts
NathanPJF: Yep, this is a reasonable question. Up in the "Decision" section of the PR I say this is a conscious choice to immediately flag to folks that their output is likely not what they think it is :)

mbarak: yes but do we want to "break ci" for those folks? i'll follow your lead here.

packages/theme-check-common/src/checks/unsupported-filter-arguments/index.spec.ts
NathanPJF: Good call. Added focused regression specs: a quoted pipe preceding a real filter (asserts the highlight starts at the real filter), two render tags on one line (asserts two distinct highlights), a `render` inside a `{% liquid %}` tag, and quoted-pipe false-positive cases in the malformed-fallback path for both single and double quotes. (dc94471)

packages/theme-check-common/src/checks/unsupported-filter-arguments/index.spec.ts
NathanPJF: Agreed — kept the original valid-markup case (proves structured parse short-circuits) and added a separate malformed-fallback case using `invalid_argument` so the raw scanner actually runs with the only pipe inside a quoted string. Also added a double-quoted variant. (dc94471)

packages/theme-check-common/src/checks/no-filters-in-render-arguments/index.ts
NathanPJF: Done — now importing `NamedTags` from `@shopify/liquid-html-parser` and matching against it. (dc94471)

NathanPJF: Confirmed `content_for` has the exact same fallback shape — `{% content_for 'block', id: 'bar' | append: 'x' %}` fails the strict parse and lands on a base-case LiquidTag with raw string markup, so the silent-filter bug applies there too. Extended the guard (and the message/docs/changeset) to cover `content_for` and added a spec for it. (dc94471)

packages/theme-check-common/src/checks/unsupported-filter-arguments/index.ts
NathanPJF: The visitor type `CheckNodeMethod` is typed to return `Promise<void>`, so a plain sync `LiquidTag(node)` fails to type-check (`Type 'void' is not assignable to type 'Promise<void>'`), and every other check in the package uses `async` for the same reason. I kept `async` here to stay consistent and satisfy the types. Happy to revisit if we relax the visitor signature project-wide.

packages/theme-check-common/src/checks/unsupported-filter-arguments/index.ts
NathanPJF: Thanks — keeping it as a recommended `Severity.ERROR`. The output is silently wrong (the filter is dropped and the raw value renders), so I'd rather fail loudly on upgrade than let incorrect markup ship unnoticed. If we get signal that it's too disruptive we can demote to `WARNING` later.

NathanPJF (commented):
Thanks for the review @mbarak . I decided to rename the check now to `UnsupportedFilterArguments` since it's covering more than just `render` now. That makes it easier to expand in the future if need be.

I also noticed that none of the other checks started with "No", so I chose something that feels closer to the existing patterns. 

Going to update the Dev Docs PR now.

PR #1252, opened 2026-07-14
Adopt hand-written liquid-html-parser for performance

karreiro:
Introduces the hand-written liquid-html-parser, replacing the Ohm grammar-based parser. Benchmarks on the theme corpus show it is roughly 75x faster than the Ohm parser.


lukeh-shopify (approved):
LGTM

stephanie-shopify (approved):

PR #1255, opened 2026-07-14
Add resilient liquid-html parsing

karreiro:
Adds resilient, error-recovering parsing so malformed templates still produce a usable AST instead of failing outright.

Part 4 of 7 in the stacked series, based on PR 3.


packages/liquid-html-parser/src/errors.ts
```diff
@@ -25,20 +25,32 @@ export class LiquidHTMLASTParsingError extends SyntaxError {
...
+    const end = lc.fromIndex(Math.min(Math.max(endIndex, 0), lastIndex));
 
     // Plugging ourselves into @babel/code-frame since this is how
     // the babel parser can print where the parsing error occured.
     // https://github.com/prettier/prettier/blob/cd4a57b113177c105a7ceb94e71f3a5a53535b81/src/main/parser.js
     this.loc = {
       start: {
-        line: start!.line,
-        column: start!.col,
+        line: start?.line ?? 1,
```
lukeh-shopify, line 60: why do we have to do this now?

lukeh-shopify (approved):

packages/liquid-html-parser/src/errors.ts
karreiro: We need this because `fromIndex(i)` returns `null` indexes for out of range, and an unclosed `%}` tags may report `-1` (so the old `start!.line` assertion would throw a `TypeError`)

PR #1267, opened 2026-07-31
Allow section and block objects in theme-mode snippets

NathanPJF:
### What

`UndefinedObject` no longer reports `section` and `block` as undefined in theme-mode **snippets**. It adds them (alongside `app`) to the contextual objects returned for `snippets/` in theme mode.

Issue of false positive warning

Screenshot 2026-07-31 at 3 34 01 PM

### Why

Snippets are commonly rendered from a section or block, which pass `section` and `block` down into scope. Referencing them in a snippet is valid, but the check flagged them:

```
[UndefinedObject] Unknown object 'section' used.
```

This was a false positive. App-mode snippets already had access to the full block-level object set; this brings `section` and `block` into theme-mode snippets too. `recommendations` stays section/block-only and is still flagged in theme-mode snippets.

### How

- `getContextualObjects` returns `['app', 'section', 'block']` for theme-mode `snippets/`.
- Updated tests: `section`/`block` are now allowed in theme-mode snippets; `recommendations` is still flagged.
- Added a changeset (patch).

Co-Authored by AI

### DIFF
--- a/packages/theme-check-common/src/checks/undefined-object/index.ts
+++ b/packages/theme-check-common/src/checks/undefined-object/index.ts
@@ -220,7 +220,10 @@ function getContextualObjects(relativePath: string, mode: Mode = 'theme'): strin
     // In a theme app extension, snippets can only be rendered from blocks,
     // so they have access to the same contextual objects as blocks.
     if (mode === 'app') return BLOCK_CONTEXTUAL_OBJECTS;
-    return ['app'];
+    // In a theme, a snippet is commonly rendered from a section or block, which
+    // pass `section` and `block` down into scope, so they should not be reported
+    // as undefined here.
+    return ['app', 'section', 'block'];
   }
 
   return [];
--- a/.changeset/undefined-object-section-block-snippets.md
+++ b/.changeset/undefined-object-section-block-snippets.md
@@ -1,0 +1,11 @@
+---
+'@shopify/theme-check-common': patch
+---
+
+`UndefinedObject` no longer reports `section` and `block` as undefined in theme-mode snippets.
+
+Snippets are commonly rendered from a section or block, which pass `section` and `block` down into scope. Treating them as contextual objects in theme-mode snippets (alongside `app`) avoids false positives such as:
+
+```
+[UndefinedObject] Unknown object 'section' used.
+```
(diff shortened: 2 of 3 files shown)


packages/theme-check-common/src/checks/undefined-object/index.ts
```diff
@@ -220,7 +220,10 @@ function getContextualObjects(relativePath: string, mode: Mode = 'theme'): strin
     // In a theme app extension, snippets can only be rendered from blocks,
     // so they have access to the same contextual objects as blocks.
     if (mode === 'app') return BLOCK_CONTEXTUAL_OBJECTS;
-    return ['app'];
+    // In a theme, a snippet is commonly rendered from a section or block, which
+    // pass `section` and `block` down into scope, so they should not be reported
+    // as undefined here.
+    return ['app', 'section', 'block'];
```
graygilmore, line 226: I'm a little torn because a snippet is not guaranteed to be rendered from a section or block. This is probably a reasonable trade off, though.

graygilmore (approved):

packages/theme-check-common/src/checks/undefined-object/index.ts
NathanPJF: This is fair to flag. I went back and forth on it, but here's what made me think this was the correct trade-off:

This follows the same reasoning `UndefinedObject` already uses for objects like `collection` and `product`. The `globalObjects` const checks `access.global === true || access.template.length > 0` 

They are teated as available everywhere to avoid false positives when the rendering context is unknown. So these changes extends that leniency to `section` and `block`. 

For sure there's a risk that a snippet will call `collection.*` outside of a collection template, and then things will break. Or call `block.*` outside the context of a block file.

But the alternatives are either (1) at the theme-dev level to set `collection`/`block` as a `@param` to signal context... which isn't fool proof because we don't support `object` type params, and especially not resource-type params.  Or (2) the check snakes through the actual use each time in-place, like the `getReferences` method does - which could be good, but changes the behaviour of the check today.

Option 2 could be useful (might need a MAJOR change), but for the time being I'm hoping to ease the check for these 2 objects :)

NathanPJF: Option 1 is what we do in Horizon's codebase; and it bugs me every time

```
  @param {object} [block] - The block object
  @param {object} [section] - The section object
```

Set them as optional because I don't _actually_ want anyone to pass an object most of the time; but I also don't want to disable the check on any lines 🙃

PR #1268, opened 2026-07-31
Add Ruby Liquid parity test suite for LiquidSyntaxError

lukeh-shopify:

Migrate the Liquid parity test suite for LiquidSyntaxError to the public theme-check-common

Copies like-for-like, and ensures the test run in their new location


karreiro (approved):
Thank you for this PR, @lukeh-shopify!

PR #1281, opened 2026-08-18
Add check for standard_event_data

lukeh-shopify:

This new filter has parameter values that are only validated on the server side, and it would be helpful to have a dedicated check for them

As the parameter values don't often change, replicate the list of allowed values in a dedicated check

### DIFF
--- a/packages/theme-check-common/src/checks/valid-standard-event-data/index.ts
+++ b/packages/theme-check-common/src/checks/valid-standard-event-data/index.ts
@@ -1,4 +1,4 @@
-import { LiquidNamedArgument, NodeTypes } from '@shopify/liquid-html-parser';
+import { LiquidExpression, LiquidNamedArgument, NodeTypes } from '@shopify/liquid-html-parser';
 import { LiquidCheckDefinition, Severity, SourceCodeType } from '../../types';
 
 const FILTER_NAME = 'standard_event_data';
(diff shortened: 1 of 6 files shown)


packages/theme-check-common/src/checks/valid-standard-event-data/index.ts
```diff
@@ -0,0 +1,67 @@
...
+        );
+        const contextValue = contextArgument?.value;
+
+        if (
+          contextValue?.type === NodeTypes.String &&
+          !CONTEXTS_SUPPORTED_BY_ANY_INPUT_TYPE.includes(contextValue.value)
+        ) {
+          context.report({
+            message: `Unsupported context '${
+              contextValue.value
```
frandiox: Maybe this could say that omitting `context` is also fine?

packages/theme-check-common/src/checks/valid-standard-event-data/index.ts
```diff
@@ -0,0 +1,67 @@
...
+  },
+
+  create(context) {
+    return {
+      async LiquidFilter(node) {
+        if (node.name !== FILTER_NAME) return;
+
+        const eventType = node.args.find((arg) => arg.type !== NodeTypes.NamedArgument);
+
+        if (eventType?.type === NodeTypes.String && eventType.value !== SUPPORTED_EVENT_TYPE) {
```
frandiox: I think this checks skips reporting `type: 123` because it only reports on strings?

frandiox (commented):
Thanks! Is it possible to detect what's the drop passed to the filter? If not then this is good I think. If it's possible, then perhaps better to branch the validation. E.g. `recommended` context is not valid for carts but it is for products.

charlespwd (approved):
Code LGTM, can't vouch for business logic :)

EvilGenius13 (approved):

packages/theme-check-common/src/checks/valid-standard-event-data/index.ts
lukeh-shopify: fixed

packages/theme-check-common/src/checks/valid-standard-event-data/index.ts
lukeh-shopify: fixed

lukeh-shopify:
> Thanks! Is it possible to detect what's the drop passed to the filter? If not then this is good I think. If it's possible, then perhaps better to branch the validation. E.g. `recommended` context is not valid for carts but it is for products.

Added some changes to detect a direct global like `{{ product | ... }}` or `{{ cart | ... }}`

PR #1282, opened 2026-08-20
Accept canonical app block paths in block tags

richpoirier:

Theme Check reports a `LiquidHTMLSyntaxError` for canonical inline app block paths. This change accepts `shopify://apps/<app>/blocks/<handle>/<uuid>` values in `block` tags while preserving errors for malformed paths and existing invalid block types.

It also removes duplicate block-name validation from `LiquidSyntaxError`. The parser now owns block-name validation, so successfully parsed `BlockMarkup` values are not rejected by a second validator.

### DIFF
--- a/packages/liquid-html-parser/src/tags/block.test.ts
+++ b/packages/liquid-html-parser/src/tags/block.test.ts
@@ -218,6 +218,33 @@ describe('blockTag', () => {
     );
   });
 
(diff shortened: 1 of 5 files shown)


charlespwd (changes requested):
Can you please add a changeset. Rest of PR looks good to me.

PR #1286, opened 2026-08-26
Add `JSONMissingSection` check

lukeh-shopify:

Theme Check already validates block and section references (in liquid files), but nothing validates the section type references inside json files. A theme can pass theme check locally while referencing a section file that doesn't exist, and then fail server-side validation on upload/publish.

This adds a JSONMissingSection check so that error surfaces at lint time instead

- Modeled on JSONMissingBlock
- For every entry in the sections map of a json template or section group that has a string type, report an offense on the type node when `sections/<type>.liquid` doesn't exist.
- Skips apps and _blocks and entries without a type.

### 📸 

image

### DIFF
--- a/packages/theme-check-common/src/checks/json-missing-section/index.ts
+++ b/packages/theme-check-common/src/checks/json-missing-section/index.ts
@@ -1,0 +1,69 @@
+import { getLocEnd, getLocStart, nodeAtPath } from '../../json';
+import { getSchemaFromJSON } from '../../to-schema';
+import { JSONCheckDefinition, Severity, SourceCodeType } from '../../types';
+import { doesFileExist } from '../../utils/file-utils';
+
+const SECTION_TYPES_WITH_PLATFORM_DEFAULTS = ['apps', '_blocks'];
+
+export const JSONMissingSection: JSONCheckDefinition = {
+  meta: {
+    code: 'JSONMissingSection',
+    name: 'Check for missing section files in JSON templates and section groups',
+    docs: {
+      description:
+        'This check ensures that section types in JSON templates and section groups refer to existing section files.',
+      recommended: true,
+      url: 'https://shopify.dev/docs/storefronts/themes/tools/theme-check/checks/json-missing-section',
+    },
+    type: SourceCodeType.JSON,
+    severity: Severity.ERROR,
+    schema: {},
+    targets: [],
+  },
+
+  create(context) {
+    const relativePath = context.toRelativePath(context.file.uri);
+    const isJsonTemplate = relativePath.startsWith('templates/');
+    const isSectionGroup = relativePath.startsWith('sections/');
+    if (!isJsonTemplate && !isSectionGroup) return {};
+
+    return {
+      async onCodePathEnd() {
+        const schema = await getSchemaFromJSON(context);
+        const { ast } = schema ?? {};
+        if (!ast || ast instanceof Error) return;
+
+        const sectionsNode = nodeAtPath(ast, ['sections']);
+        if (!sectionsNode || sectionsNode.type !== 'Object') return;
+
+        await Promise.all(
+          sectionsNode.children.map(async (property) => {
+            const sectionNode = property.value;
+            if (sectionNode.type !== 'Object') return;
+
+            const typeNode = sectionNode.children.find(
+              (child) => child.key.value === 'type',
+            )?.value;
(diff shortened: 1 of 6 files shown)


packages/theme-check-common/src/checks/json-missing-section/index.ts
```diff
@@ -0,0 +1,69 @@
...
+            if (typeof sectionType !== 'string') return;
+            if (SECTION_TYPES_WITH_PLATFORM_DEFAULTS.includes(sectionType)) return;
+
+            const sectionFileExists = await doesFileExist(
+              context,
+              `sections/${sectionType}.liquid`,
+            );
+            if (sectionFileExists) return;
+
+            const typeNode = nodeAtPath(ast, ['sections', sectionKey, 'type']) as
```
graygilmore: (feel free to hit this up in a follow up if you want)

This is kind of a nit because we'd already hit a duplicate key error but I thought it was worth flagging:

Image

The solution to this would require reworking this whole function a bit but the net change wouldn't be too bad:

```diff
 import { getLocEnd, getLocStart, nodeAtPath } from '../../json';
 import { getSchemaFromJSON } from '../../to-schema';
-import { JSONCheckDefinition, JSONNode, Severity, SourceCodeType } from '../../types';
+import { JSONCheckDefinition, Severity, SourceCodeType } from '../../types';
 import { doesFileExist } from '../../utils/file-utils';
 
 const SECTION_TYPES_WITH_PLATFORM_DEFAULTS = ['apps', '_blocks'];
@@ create(context)
     return {
       async onCodePathEnd() {
         const schema = await getSchemaFromJSON(context);
-        const { ast, parsed } = schema ?? {};
+        const { ast } = schema ?? {};
         if (!ast || ast instanceof Error) return;
-        if (!parsed || parsed instanceof Error) return;
 
-        const sections = parsed.sections;
-        if (!sections || typeof sections !== 'object') return;
+        const sectionsNode = nodeAtPath(ast, ['sections']);
+        if (!sectionsNode || sectionsNode.type !== 'Object') return;
 
         await Promise.all(
-          Object.entries(sections).map(async ([sectionKey, section]) => {
-            if (!section || typeof section !== 'object') return;
+          sectionsNode.children.map(async (property) => {
+            const sectionNode = property.value;
+            if (sectionNode.type !== 'Object') return;
 
-            const sectionType = (section as { type?: unknown }).type;
+            const typeNode = sectionNode.children.find(
+              (child) => child.key.value === 'type',
+            )?.value;
+            if (!typeNode || typeNode.type !== 'Literal') return;
+
+            const sectionType = typeNode.value;
             if (typeof sectionType !== 'string') return;
             if (SECTION_TYPES_WITH_PLATFORM_DEFAULTS.includes(sectionType)) return;
 
             const sectionFileExists = await doesFileExist(
               context,
               `sections/${sectionType}.liquid`,
             );
             if (sectionFileExists) return;
 
-            const typeNode = nodeAtPath(ast, ['sections', sectionKey, 'type']) as
-              | JSONNode
-              | undefined;
-            if (!typeNode) return;
-
             context.report({
               message: `Section type '${sectionType}' does not refer to an existing section file`,
               startIndex: getLocStart(typeNode),
               endIndex: getLocEnd(typeNode),
             });
           }),
         );
       },
     };
```

Kind of nice that we'd get rid of the `JSONNode` casting, too.

### Bot comment

> 🐛 Bug: sectionType comes from parsed, where a duplicated section key resolves last-wins. typeNode comes from the AST, where nodeAtPath returns the first match. With a duplicate key like `"foo": {"type":"aaa"}, "foo": {"type":"bbb"}`, the message names 'bbb' while the underline sits on "aaa".
> 
> Suggestion: Read the type from the AST node itself (typeNode.value) instead of parsed. Then the message and the range always come from the same node, and duplicate keys each get checked correctly.
> 
> It points at the exact mechanism (two sources, two duplicate-key semantics) and gives one concrete repro input, which is usually enough for the author to reproduce and land on the AST-walk fix themselves.

graygilmore (approved):
👍🏻 🎩 

Image

packages/theme-check-common/src/checks/json-missing-section/index.ts
lukeh-shopify: nice catch! i reworked it in this PR. seems to work even with duplicate keys now

image

PR #1295, opened 2026-09-10
Parse empty filter argument lists

charlespwd:

Parse a filter colon followed by the end of markup or another filter as an empty argument list.

These forms now produce a structured `LiquidVariable` with an empty `args` array, and the filter range includes the colon:

```liquid
{{ 'key' | t: }}
{{ 'key' | t: | escape }}
```

Malformed arguments such as `{{ 'key' | t: @ }}` continue to use string markup. Tests cover strict Liquid and LiquidHTML parsing, tolerant parsing, filter ranges, and `echo` tags.

None.

## Tophatting

Run `CI=1 pnpm test` from the repository root.

- [x] Screenshots are not applicable to parser-only changes.

## Before you deploy

- [x] I included a patch bump `changeset`.

### DIFF
--- a/.changeset/empty-filter-arguments.md
+++ b/.changeset/empty-filter-arguments.md
@@ -1,0 +1,5 @@
+---
+'@shopify/liquid-html-parser': patch
+---
+
+Parse empty filter argument lists such as `{{ 'key' | t: }}`.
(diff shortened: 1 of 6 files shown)


karreiro (approved):
Thank you, @charlespwd!

PR #1301, opened 2026-09-15
Fix several Theme Check paper cuts

karreiro:
### What are you adding in this PR?

This PR fixes several small gaps across Theme Check:

- Fix responsive LCP preload validation by allowing explicit image links with `fetchpriority="high"`
- Fix LiquidDoc parameter validation by reserving Liquid literal keywords
- Fix schema expressions to resolve assigned Liquid variables and recognize their usage
- Fix Liquid variable-name validation to allow a leading underscore
- Fix orphaned-snippet detection by following layout and `{% block %}` references

Each fix includes regression coverage and its own patch changeset.

### Before you deploy

- [x] This PR includes patch changesets for each bug fix.


packages/theme-check-common/src/checks/valid-visible-if/index.ts
```diff
@@ -39,7 +44,15 @@ export const ValidVisibleIf: LiquidCheckDefinition = {
...
+    const assignedVariables: Vars = {};
+
     return {
+      async LiquidTag(node) {
+        if (isLiquidTagAssign(node)) {
+          assignedVariables[node.markup.name] = true;
+        }
+      },
+
       async LiquidRawTag(node) {
```
aswamy, line 56: Could we avoid making this source-order dependent? 

Since `{% schema %}` tag can appear even before source code, we need to defer this logic to `onCodePathEnd` so we don't get errors when we have something like this:

```
{% schema %}
{
  "name": "Example",
  "settings": [{
    "type": "checkbox",
    "id": "enabled",
    "label": "Enabled",
    "visible_if": "{{ effective_fill == 'image' }}"
  }]
}
{% endschema %}

{% assign effective_fill = 'image' %}
```

packages/theme-check-common/src/checks/asset-preload/index.ts
```diff
@@ -9,6 +9,21 @@ function isPreload(attr: ValuedHtmlAttribute): boolean {
   );
 }
 
+function isHighPriorityImagePreload(attributes: ValuedHtmlAttribute[]): boolean {
+  return (
+    attributes.some(
+      (attr) =>
+        isAttr(attr, 'as') &&
+        attr.value.some((node) => node.type === NodeTypes.TextNode && node.value === 'image'),
```
aswamy, line 17: Could we require the entire attribute to equal the expected value? Checking individual text nodes exempts values like `as="image{{ '-invalid' }}"`. Using `hasAttributeValueOf(attr, 'image')` and the same check for `"high"` would avoid this false negative.

e.g.
```
<link
  href="hero.png"
  rel="preload"
  as="image{{ '-invalid' }}"
  fetchpriority="high"
>
```

aswamy (approved):
Couple of tweaks recommended but overall good approach

packages/theme-check-common/src/checks/asset-preload/index.ts
karreiro: Good catch! Thank you! Fixed by using `hasAttributeValueOf` for both attributes 🙇

packages/theme-check-common/src/checks/valid-visible-if/index.ts
karreiro: Thank you! Good catch! Fixed :)

PR #1302, opened 2026-09-22
Add SchemaPresetsAndDefault check

### LINKED ISSUE #750: Theme check should warn you when you have `presets` and `default`
**Describe the bug**
If you have a `presets` and `default` in your section file's schema, you will see no theme check error, but when you run `shopify theme dev`, you will see an error saying the error.

**Source**

Image

Image

**Expected behaviour**
- Theme check should catch the error in VSCode
- Theme check in CLI should catch the error (i.e. running `shopify theme check`)

**Actual behaviour**
- No errors are shown until you run `shopify theme dev`

web-orbit12:

Fixes #750

A section schema that defines both `presets` and `default` passes Theme Check today, but `shopify theme dev` rejects it with:

```
Invalid schema: cannot define both 'default' and 'presets'
```

This adds a `SchemaPresetsAndDefault` check that reports the same error in the editor and in `shopify theme check`, highlighting the `default` value.

- Runs on section files only: theme block schemas have no `default` property.
- Only reports on otherwise valid schemas, like the other `SchemaPresets*` checks.
- The message matches the platform error word for word.

The check needs a page on shopify.dev (`checks/schema-presets-and-default`), which a Shopify team member would have to add.

I treated key presence as the trigger, so an empty `presets: []` with `default` is also flagged. Happy to switch to non-empty `presets` only if the platform accepts that combination.

## Tophatting

- Put the schema below in `sections/test.liquid`: the error appears on `default`. Remove either key and it goes away.

```liquid
{% schema %}
{
  "name": "Test",
  "presets": [{ "name": "Test" }],
  "default": { "settings": {} }
}
{% endschema %}
```

- Tested in the playground (`pnpm playground`) with the section above, built from `main` and from this branch:

**Before** (`main` @ 8bc5e038): no error

Before: main, no error

**After** (this branch @ d4dcdf93): `default` is flagged

After: this branch, error on default

- `pnpm vitest run packages/theme-check-common/src/checks/schema-presets-and-default` (5 tests), plus the full `theme-check-common` and `theme-check-node` suites pass.

- [x] I added screenshots of the changes (before and after the changes if applicable)

## Before you deploy

  - [x] I ran `pnpm build` and committed the updated configuration files
    - [ ] If applicable, I've updated the `theme-app-extension.yml` config (not applicable: in app mode, sections have no schema to check, so the check is a no-op)


web-orbit12:
I have signed the CLA!

